PR #38, opened 2015-01-05
Restructured code

ProgramFOX:
This commit restructures the code. Almost all code was stored in ws.py
before so it became kind of a mess. Now the code is split up in multiple
files.


ProgramFOX:
Because this change is very big, I made it a pull request instead of just committing it to master.

PR #39, opened 2015-01-07
It really should be "setup.sh".

Vincentyification:
So, here's a README fix.


ProgramFOX:
If you look at the ChatExchange repository, you'll see that the file is really called setp.sh. So this is not a typo, and changing it would make the info inaccurate.

hichris1234:
Also see #29.

PR #45, opened 2015-03-20
Add "mytechlabs" to the list of bad websites.

Vincentyification:
http://chat.meta.stackexchange.com/transcript/message/3185215#3185215


ProgramFOX:
Thanks!

PR #48, opened 2015-03-26
Added two new conversion spam links

Seth-Johnson:


ProgramFOX:
Thanks!

I merged the patch and added a new commit to:

 - Escape the dots in the sites (they're immediately used in the regex).
 - Remove the trailing slash. It might happen that the spammers post these links without this slash.

Seth-Johnson:
Thanks ProgramFOX. It was late and I wasn't entirely sure how the URLs were handled. Now I know for next time!

PR #52, opened 2015-04-06
Increase link text size

JC3:
Increase link text size because the chat action icons obscure the link
on my monitor
http://meta.stackexchange.com/questions/252888/chat-message-actions-obscure-links-at-end-of-line
...


ProgramFOX:
Looks good, thanks!

PR #56, opened 2015-05-29
Updates black listed websites with most common spam URLs from May 2015

AWegnerGitHub:
More details: http://chat.stackexchange.com/transcript/message/21888796#21888796


Undo1:
So flake8 doesn't like you either. Join the club :)

PR #57, opened 2015-06-28
skin ?care, removed redundant sites

normalhuman:
1) skin care is sometimes written as one word, so I've made space optional.
2) sites with garcinia in the name don't need to be blacklisted: garcinia is blacklisted without word boundaries

P.S. user currently known as 1999


hichris1234:
merged, thanks.

PR #59, opened 2015-06-30
neuroluma, eyeserum, health-regexes

normalhuman:
1. Neuroluma showed up 3 times recently: http://stackexchange.com/search?q=neuroluma
2. Moved "eye ?serum" to nwb list, so that it also takes care of domains and variations
3. Generalized health{numbers} regex because numbers vary:  health786.com  appeared recently
4. Moved "tophealth" to nwb list because domains vary. No false positives even on Health.SE (checked with http://data.stackexchange.com/health/query/330774/search-with-no-word-boundaries?pattern=tophealth 
5. Removed redundant garcinia site (garcinia is nwb blacklisted)
6.... not sure why the diff shows the first line as edited

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1,4 +1,4 @@
-﻿# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 import regex
 import phonenumbers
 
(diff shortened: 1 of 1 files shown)


Undo1:
bah, beat me to it @Manish :)

PR #70, opened 2015-07-13
make muscle, anti-aging nwb, etc

normalhuman:
- Moved "muscle" outside of word boundaries on non-health sites. It's not a part of any other word, but is usually a part of spam domain name.
- moved anti[- ]?aging to nwb list. They hardly ever use this term on biology/health/fitness; even there it's likely to be spam. 
- added "serum" to "bad part of domain" list


Undo1:
Someday I'm just going to stay up all night so that I can catch one of these before Manish does

PR #71, opened 2015-07-14
health+pro = spam

normalhuman:
More general patterns of "health"+something spam domains (can't blacklist health on its own). If they work as expected, some of specific domains can be trimmed later from the blacklist.
 
Also added a few specific domains.


Manishearth:
FOILED

Manishearth:
I step away from my laptop for _one minute_ and @ProgramFOX gets to it first, grrr.

ProgramFOX:
Yay, I beat Manish! :P

PR #72, opened 2015-07-14
Added easydatafeed.com to blacklist

durron597:
There have been many recent spam posts for easydatafeed.com such as this one: http://stackoverflow.com/a/31416687/1768232 (10k link obviously)


ProgramFOX:
It's already there, in a different format:
https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L63

PR #74, opened 2015-07-16
-redundant, +premium{stuff}

normalhuman:
- removed specific domains that are covered by general "cheap" pattern added last time
- added a pattern for "premium", avoiding legitimate sites like premiumbeat, premiumcoding, premiumpixels. 
- added a couple specific domains


Undo1:
I need to just move to a different timezone

ProgramFOX:
@Undo1 Heh, I believe you're closer to normalhuman's timezone than I am.

PR #76, opened 2015-07-18
+profactor, better fatburn

normalhuman:
*  Added Profactor T 
* Generalized fatburn pattern; it's reliably spam when used in site domain. Same for topcare.


Manishearth:
On airport wifi, *still* faster than @Undo1 _blows raspberry_

Undo1:
grumble grumble sleeping midnight grumble

PR #88, opened 2015-08-26
Added herpes to spam filter

durron597:
We had a spam that had herpes in it, I want to make sure it gets caught (it didn't). 

https://chat.stackexchange.com/transcript/message/23677836#23677836


normalhuman:
Added to a different rule, to exclude health and biology.

PR #90, opened 2015-09-03
Change %-formatting to {}-formatting

michaelb958:
So I'm told that %-formatting is officially a thing that should not be
used in new code(tm), and I figured this was a half-decent way to
acquaint myself with the codebase.

### DIFF
--- a/bayesian/bayes.py
+++ b/bayesian/bayes.py
@@ -17,7 +17,7 @@ if __name__ == '__main__':
 		register_mode(Status)
 
 		args = sys.argv
-		usage = 'Usage: %s %s <mode specific args>' % (args[0], '|'.join(modes.keys()))
+		usage = 'Usage: {} {} <mode specific args>'.format(args[0], '|'.join(modes.keys()))
 
 		if (len(args) < 2):
 			raise ValueError(usage)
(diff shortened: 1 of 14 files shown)


ProgramFOX:
That's interesting! Where is its deprecated-ness stated?

normalhuman:
@ProgramFOX  This seems to be the official source https://www.python.org/dev/peps/pep-3101/ 
It seems one has to be careful around Unicode vs non-Unicode strings, but all of ours are already Unicode: http://stackoverflow.com/a/12252460

kevin-brown:
> Where is its deprecated-ness stated?

It's not deprecated yet, it's just _preferred_ to use the new `string.format` over the old overloaded `%` operator.

You might find the following interesting: [Pythons many ways of string formatting — are the older ones (going to be) deprecated?](http://stackoverflow.com/q/13451989/359284)

ProgramFOX:
I think this can be merged. Does everyone else agree?

Undo1:
@ProgramFOX  Looks good to me.

normalhuman:
@ProgramFOX  No objections from me.

Undo1:
In that case, I'm doing it. Let's break stuff.

hichris1234:
Late :+1: from me. ;)

PR #91, opened 2015-09-04
Cleanup watcher and fix chat.meta/3918608

michaelb958:
Fixes http://chat.meta.stackexchange.com/transcript/message/3918608#3918608, and generally makes the command-watcher code cleaner.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -63,9 +63,6 @@ def watcher(ev, wrap2):
     shortcut_messages = []
     if message_parts[0].lower() == "sd":
         message_parts = preprocess_shortcut_command(content_source).split(" ")
-        if len(GlobalVars.latest_smokedetector_messages[ev_room]) == 0:
-            ev.message.reply("I don't have any messages posted after the latest reboot.")
-            return
         commands = message_parts[1:]
         length_without_skips = 0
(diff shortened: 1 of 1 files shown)


normalhuman:
I think it's best to wait until the errors caused by #90 are sorted out.

hichris1234:
I just tested this out in [Charcoal HQ](http://chat.stackexchange.com/rooms/11540/charcoal-hq)... and while I like some of the improvements, this can't work since markdown (and thus links) aren't supported in multi-line messages. 

http://chat.meta.stackexchange.com/transcript/message/3918608#3918608 is by-design since it's supposed to be just a quick way of identifying what messages had what return values. Although, if there's any ideas you have on how to improve it...

I like the simplifying aspect to it. If you could make that a separate PR (or if you just want me to do it), I'd be fine with merging that in.

hichris1234:
Implemented the simplification aspect in bbcdaeaf4c7522c0cf472ef743dc939f9a00feb4 and 3ee86d407db96dc8b93533a52b9c197455514aa4. At this point, we're not going to include the links as chat just doesn't like multi-line markdown. :(

Closing this as I believe I've addressed all parts of it. Feel free to ping me if there's something I missed.

PR #100, opened 2015-11-08
Adjusted keywords to reduce false positives.

BrockA:
Removed "black magic" and adjusted "weight loss" to reduce false positives.

Regarding "weight loss":

 - It was causing too many false positives due to [the term *deadweight loss*](https://en.wikipedia.org/wiki/Deadweight_loss), which frequently occurs in economics.
 - Reference [this chat on Charcoal HQ](http://chat.stackexchange.com/transcript/message/25247160#25247160)
 - Tweaked the regex to only match when word starts with "weight".

<br>
For "black magic" see:
 - Starting [here, on Tavern](http://chat.meta.stackexchange.com/transcript/message/4165391#4165391)  
  and
 - [Here, on Charcoal HQ](http://chat.stackexchange.com/transcript/message/25247400#25247400)

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -45,7 +45,7 @@ class FindSpam:
                     "vashi?k[ae]r[ae]n", "kolcak", "Zapyo", "we (offer|give out) (loans|funds|funding)",
                     "porn", "molvi", "judi bola", "ituBola.com", "lost lover'?s?",
                     "rejuvenated skin", "ProBrain", "restore[ -]?samsung[ -]?data",
-                    "LifeForce", "swtor2credits", "me2.do",  # Removed "black magic"; too many false positives
+                    "LifeForce", "swtor2credits", "me2.do", "^(?s).{0,200}black magic",
                     "bam2u", "Neuro(3X|flexyn|fuse|luma|plex)", "TesteroneXL", "Nitroxin",
                     "Bowtrol", "Slim ?Genix", "Cleanse EFX", "Alpha Rush",
                     "Blackline Elite", "TestCore Pro", "blank(ed)? ?ATM(card)?", "ATM Machine Vault",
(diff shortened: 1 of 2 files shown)


Manishearth:
I'd rather reduce the number of sites "black magic" applies on.

There is a recurring spammer who posts about black magic.

BrockA:
@Manishearth Okay, I'll investigate that.  I've seen nothing but false positives for it so far.

Can we kill it for now while I look into the per-site thing?

Manishearth:
@ProgramFOX thoughts?

BrockA:
[He already seemed to agreed to nuke it.](http://chat.meta.stackexchange.com/transcript/message/4166288#4166288)

Manishearth:
"from Islam"

BrockA:
You're right, but then: http://chat.stackexchange.com/transcript/message/25247400#25247400

Manishearth:
Yes, I wonder what he says in light of the "recurring spammer" thing

BrockA:
Can you point out some of that spam?  Might be able to fix it another way, maybe...

Manishearth:
Well, it's all gone now :smile: 

It's similar to the vashikaran spam. It was one of the reasons we _created_ smokey in the first place.

BrockA:
@Manishearth Okay, I put the check back in for now.   
Sure those other BM spams wouldn't have been caught by another factor? (Links, etc.) because this keyword is 0 for 3 since I started noticing it.

Manishearth:
Not sure. r=me once travis passes

BrockA:
Huh, wonder what the delay is...

BrockA:
Okay @Manishearth, all checks have passed.

BrockA:
Thanks!

PR #101, opened 2015-11-08
Removed "black magic" check only from Islam site.

BrockA:
**Warning: Untested!**  (How do we test filters without making bad posts, other than the `!!/test` command?)

"Black magic" was causing too many false positives.

Removed it from the Islam site only.

See:
 - Starting [here, on Tavern](http://chat.meta.stackexchange.com/transcript/message/4165391#4165391)  
  and
 - [Here, on Charcoal HQ](http://chat.stackexchange.com/transcript/message/25247400#25247400)

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -282,7 +282,7 @@ class FindSpam:
         {'regex': u".*<pre>.*", 'all': False, 'sites': ["puzzling.stackexchange.com"], 'reason': 'Code block', 'title': False, 'body': True, 'username': False, 'stripcodeblocks': False, 'report_everywhere': False, 'body_summary': False},
         {'regex': u"(?i)(erica|jeff|er1ca|spam|moderator)", 'all': False, 'sites': ["parenting.stackexchange.com"], 'reason': "Bad keyword in {}", 'title': False, 'body': True, 'username': False, 'stripcodeblocks': False, 'body_summary': True},
         {'regex': ur"^(?is).{0,200}black magic", 'all': False,
-         'sites': [islam.stackexchange.com], 'reason': "Black magic in {}", 'title': True, 'body': True, 'username': True, 'stripcodeblocks': False, 'body_summary': True}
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -279,7 +280,9 @@ class FindSpam:
...
         {'regex': u".*<pre>.*", 'all': False, 'sites': ["puzzling.stackexchange.com"], 'reason': 'Code block', 'title': False, 'body': True, 'username': False, 'stripcodeblocks': False, 'report_everywhere': False, 'body_summary': False},
-        {'regex': u"(?i)(erica|jeff|er1ca|spam|moderator)", 'all': False, 'sites': ["parenting.stackexchange.com"], 'reason': "Bad keyword in {}", 'title': False, 'body': True, 'username': False, 'stripcodeblocks': False, 'body_summary': True}
+        {'regex': u"(?i)(erica|jeff|er1ca|spam|moderator)", 'all': False, 'sites': ["parenting.stackexchange.com"], 'reason': "Bad keyword in {}", 'title': False, 'body': True, 'username': False, 'stripcodeblocks': False, 'body_summary': True},
+        {'regex': ur"^(?is).{0,200}black magic", 'all': False,
+         'sites': [islam.stackexchange.com], 'reason': "Black magic in {}", 'title': True, 'body': True, 'username': True, 'stripcodeblocks': False, 'body_summary': True}
```
Manishearth: Should be all: True. Then the sites key is the blacklist, otherwise it's the whitelist

Undo1:
> How do we test filters without making bad posts

That's what https://github.com/Charcoal-SE/SmokeDetector/blob/master/test/test_regexes.py is for :)

BrockA:
Thanks, @Undo1.  I don't see an existing check for "black magic" though.?

BrockA:
@Manishearth, fixed the logic issue, I think, and the CI checks pass.

BrockA:
Thanks @Manishearth!

PR #111, opened 2015-11-27
Adding Robert Longson to Charcoal HQ too .

BrockA:
He's already in for the Tavern, but HQ is much nicer for spam busting. It's less chitchat. :)


BrockA:
Thanks!

PR #117, opened 2015-12-20
Don't zero-out quota record on API glitches

BrockA:
Don't zero-out the quota record when the API glitches or when a filter that doesn't include quota_remaining is used.

See: 
- http://chat.stackexchange.com/transcript/11540?m=26314837#26314837
- http://chat.stackexchange.com/transcript/11540?m=26260124#26260124
- etc.

### DIFF
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -80,9 +80,6 @@ class BodyFetcher:
             if response["quota_remaining"] - GlobalVars.apiquota >= 1000 and GlobalVars.apiquota >= 0:
                 GlobalVars.charcoal_hq.send_message("API quota rolled over with {} requests remaining.".format(GlobalVars.apiquota))
             GlobalVars.apiquota = response["quota_remaining"]
-        else:
-            GlobalVars.apiquota = 0
-            return
 
         if site == "stackoverflow.com":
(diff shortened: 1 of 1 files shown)


bodyfetcher.py
```diff
@@ -80,9 +80,6 @@ def make_api_call_for_site(self, site):
             if response["quota_remaining"] - GlobalVars.apiquota >= 1000 and GlobalVars.apiquota >= 0:
                 GlobalVars.charcoal_hq.send_message("API quota rolled over with {} requests remaining.".format(GlobalVars.apiquota))
             GlobalVars.apiquota = response["quota_remaining"]
-        else:
-            GlobalVars.apiquota = 0
-            return
```
Manishearth: should we still be returning?

BrockA: I don't think so. That variable is not used elsewhere.

Undo1: If the API glitches, don't we *want* to return, so that we don't try to parse bad data? That's what the return is for IIRC.

normalhuman: The lack of return doesn't make a difference in the end; if there is an error then we don't get [items] from the server, and that's the end of the function.

PR #128, opened 2016-01-22
added praveen Kumar

rschrieken:
for SOCVR


Undo1:
I'm not terribly enthusiastic about this for a few reasons, but we'll see if it works out.

PR #139, opened 2016-02-13
Created !!/ami 

rselwyn:
I have noticed lately in the SOCVR chat room that people often misspell `!!/amiprivileged`, which is why I think people should be able to type in `!!/ami` instead to do the same task.

Let me know what you all think.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -517,7 +517,7 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
             result += "\n----------\n"
             result += why
         return result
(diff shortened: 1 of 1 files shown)


ProgramFOX:
I don't really think that this is a good idea, for two reasons:

1. It's not clear from the command name what it does.
2. It closes the door for future "am I ...?" commands.

BrockA:
Correctly inputting `!!/amiprivileged` is a not too difficult.  Do we want to encourage those who can't do this to become more involved with Smokey's operation? :)

Undo1:
Agree with FOX and Brock above. Good idea, but I'm pretty sure this isn't the best way to execute it.

PR #153, opened 2016-03-23
+educba.com

ferrybig:
https://metasmoke.erwaysoftware.com/search?body=educba.com


normalhuman:
Yes, I think a little redundancy is okay here. The Blacklisted website reason gives more confidence than Bad keyword, when it appears in reports.

PR #158, opened 2016-05-19
Fixed allspam and reportuser issue

The-Quill:
http://chat.stackexchange.com/transcript/message/29774087#29774087


normalhuman:
This isn't the issue: the "why" message was not meant to be printed in chat.

PR #159, opened 2016-05-21
General cleanup

The-Quill:
I did some general variable name cleanup, syntax cleanup etc

The code was tested with the test modules and by running in chat....


bodyfetcher.py
```diff
@@ -208,12 +239,39 @@ def make_api_call_for_site(self, site):
...
+                        why,
+                        owner_rep,
+                        post_score,
+                        up_vote_count,
+                        down_vote_count,
+                        None
+                    )
                 except:
-                    print "NOP"
+                    print "NOPE"
```
AWegnerGitHub: Does this need to be a `print`? If not, just make it a `pass`.

bodyfetcher.py
```diff
@@ -233,12 +291,39 @@ def make_api_call_for_site(self, site):
...
+                            why,
+                            owner_rep,
+                            post_score,
+                            up_vote_count,
+                            down_vote_count,
+                            q_id
+                        )
                         except:
-                            print "NOP"
+                            print "NOPE"
```
AWegnerGitHub: Again, do we need a `print` or can we make it `pass`?

chatcommunicate.py
```diff
@@ -119,18 +132,27 @@ def watcher(ev, wrap2):
...
+                    current_message.lower(),
+                    current_message.split(" "),
+                    ev_room, ev_room_name,
+                    ev_user_id,
+                    ev_user_name,
+                    wrap2,
+                    current_message,
+                    message_id
+                )
+                result = pure_result
```
AWegnerGitHub: What's the point of these two variables? Why not use `pure_result` in the following conditional checks instead of having both?

bodyfetcher.py
```diff
@@ -208,10 +241,37 @@ def make_api_call_for_site(self, site):
                 owner_rep = 0
             q_id = str(post["question_id"])
 
-            is_spam, reason, why = check_if_spam(title, body, owner_name, owner_link, site, q_id, False, False, owner_rep, post_score)
+            is_spam, reason, why = check_if_spam(
```
hichris1234: Making each argument get its own line is... really ugly. I believe the Google Style Guide agrees with me here: https://google.github.io/styleguide/pyguide.html#Line_length

chatcommunicate.py
```diff
@@ -532,12 +602,26 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-        return "*brews a cup of " + random.choice(['earl grey', 'green', 'chamomile', 'lemon', 'darjeeling', 'mint', 'jasmine']) + " tea for @" + ev_user_name.replace(" ", "") + "*"
+        return "*brews a cup of {choice} tea for @{user}*".format(
+            choice=random.choice([
+                'earl grey',
+                'green',
+                'chamomile',
+                'lemon',
+                'darjeeling',
+                'mint'
+                , 'jasmine'
```
AWegnerGitHub: Very minor - why's this line start with a comma, but the others in the list end with the comma?

chatcommunicate.py
```diff
@@ -59,7 +58,10 @@ def special_room_watcher(ev, wrap2):
         post_site_id = fetch_post_id_and_site_from_msg_content(content_source)
         post_url = fetch_post_url_from_msg_content(content_source)
         if post_site_id is not None and post_url is not None:
-            t_check_websocket = Thread(target=DeletionWatcher.check_if_report_was_deleted, args=(post_site_id, post_url, ev.message))
```
hichris1234: Also, do we really need to wrap everything to 80 characters? I believe that was something we specifically didn't want to do in the beginning -- 80 characters is a large restriction (https://github.com/Charcoal-SE/SmokeDetector/commit/127d2f743cbe35ec00768b00d1b5b2ffa20cb68b#diff-b91f3d5bd63fcd17221b267e851608e8R2)

config
```diff
@@ -1,4 +1,4 @@
 [Config]
 location=location_here
 
-metasmoke_host=http://127.0.0.1:3000
+metasmoke_host=dvsbsbsfb
```
Undo1: As [Andy noticed](http://chat.stackexchange.com/transcript/message/29807982#29807982), looks like this shouldn't have been committed.

AWegnerGitHub: This line change needs to be reverted. This is a bad placeholder (or reply [here](http://chat.stackexchange.com/transcript/message/29807982#29807982) so we understand)

findspam.py
```diff
@@ -150,273 +156,280 @@ def keyword_email(s, site):   # a keyword and an email in the same post
...
-                            "cuidados-saude", "klereumcol\\.com", "gupshupchatroom\\.com", "petsworld\\.in",
-                            "godiabetesrevenge", "reflectivevestsindia", "anyframe\\.net", "canadianprofits\\.tv",
-                            "dcweddingandevents\\.com", "slimdreneavis", "wefix365\\.us", "esofttools\\.com",
-                            "wondershare\\.com", "pulsionerotica\\.com", "worldtraveltime\\.net", "antivirus\\.comodo\\.com",
-                            "cardvdonline\\.com", "icasnetwork\\.org", "epicresearch\\.co", "\\.soup\\.io", "pccdkeys\\.com",
-                            "hotxt\\.co\\.uk", "rcframecontractors", "bsgolds\\.com", "okaygoods\\.com", "thedropnet\\.com",
-                            "science\\.misis\\.ru", "pdf\\.ac/", "slotobit\\.com", "thaykinhmanhinhcamung\\.com",
-                            "sfdcbot\\.herokuapp\\.com", "technosanatgroup\\.com", "csgostrong\\.com", "freenom\\.com"]
+    bad_keywords = [
+        "baba ji", "fifa.{0,20}coins?", "fifabay", "Long Path Tool",
```
hichris1234: According to the Google Style Guide, both styles are correct here: https://google.github.io/styleguide/pyguide.html#Line_length

Undo1: For what it's worth, I personally like Quill's version here better.

hichris1234:
> On a side note, I added anime.se to the list of non-latin acceptable sites because they get Japanese lyric or quote questions sometimes

Please separate that out into another PR - that way, it can be merged easily & there's a clear history of why (adding it as part of a cleanup commit is kinda strange).

bodyfetcher.py
```diff
@@ -233,12 +291,39 @@ def make_api_call_for_site(self, site):
...
+                        True,
+                        False,
+                        owner_rep,
+                        post_score
+                    )
                     if is_spam:
                         try:
-                            handle_spam(title, body, owner_name, site, link, owner_link, a_id, reason, True, why, owner_rep, post_score, up_vote_count, down_vote_count, q_id)
+                            handle_spam(
+                            title,
```
ArtOfCode-: You've got weird indentation here - arguments should be indented from the method name if you're using arguments on new lines; otherwise (see hichris' comment) they should just be on the same line as the method call.

chatcommunicate.py
ArtOfCode-: @hichris1234 I'd tend to agree; PEP8 is generally decent but the 80 char thing is restrictive. A more common limit is 100 or 120.

chatcommunicate.py
```diff
@@ -269,12 +305,17 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-                second_part_lower == "del") and is_privileged(ev_room, ev_user_id, wrap2):
+
+            return "Recorded answer as an NAA in metasmoke."
+
+        if (second_part_lower.startswith("delete") \
+            or second_part_lower.startswith("remove") \
+            or second_part_lower.startswith("gone") \
+            or second_part_lower.startswith("poof") \
+            or second_part_lower == "del") \
+        and is_privileged(ev_room, ev_user_id, wrap2):
```
ArtOfCode-: More weird indentation - I'm not sure what it *should* be, but having some conditionals on the same indent and one four spaces less feels weird.

chatcommunicate.py
```diff
@@ -532,12 +602,26 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
         GlobalVars.users_chatting[ev_room] = list(set(GlobalVars.users_chatting[ev_room]))  # Make unique
         user_to_blame = random.choice(GlobalVars.users_chatting[ev_room])
         return u"It's [{}]({})'s fault.".format(user_to_blame[0], user_to_blame[1])
-    if "smokedetector" in content_lower and "fault" in content_lower and ("xkcdbot" in ev_user_name.lower() or "bjb568" in ev_user_name.lower()):
+    if "smokedetector" in content_lower and "fault" in content_lower \
+    and ("xkcdbot" in ev_user_name.lower() \
+        or "bjb568" in ev_user_name.lower()
```
ArtOfCode-: Did this not fit on one line? Feels like it would make more sense.

continuousintegration.py
```diff
@@ -38,11 +38,17 @@ def watch_ci():
                   'Content-Length: 2\n' +
                   '\nOK\n')
         conn.close()
-        r = requests.get('https://api.github.com/repos/Charcoal-SE/SmokeDetector/git/refs/heads/master')
-        latest_sha = r.json()["object"]["sha"]
+        request = requests.get(
+            'https://api.github.com/repos/Charcoal-SE/SmokeDetector/git/refs/heads/master'
```
ArtOfCode-: This seems like a valid point where the line can go over 80 chars - it's one argument, so newlining it is weird, and it's a URL, so it doesn't make sense to split it.

ArtOfCode-:
I've added a number of line notes, most of which are purely stylistic things. This is a lot of work, so thanks for doing it :)

chatcommunicate.py
normalhuman: I think the idea is that handle_commands can return pure_result in various forms, so that variable is distinguished from "result", as "a thing we can send to chat": `reply += result + os.linesep`. Seems ok to me.

bodyfetcher.py
normalhuman: On the other hand, seeing these on separate lines suggests further improvement (for the future): naming these arguments. It's a long list with some "False, False" in the middle, and the only way to know what these do is to read the code in another module.

hichris1234: Quite true, but as of now, there's no improvement in reading ability -- if anything, it makes it slower to comprehend what's going on (as you now have to read 12 lines instead of 1).

chatcommunicate.py
The-Quill: I just renamed the variables there, in the following lines the properties of `result` get changed while `pure_result` stay pure from the source. I have no idea why the code is like that... I can take a look at refactoring it if you want @AWegnerGitHub

bodyfetcher.py
The-Quill: @hichris1234 I was considering the idea of creating temporary model files before they get passed into functions so we can fill the models when they come in and unpack them when they get out to avoid confusion (unknown property usage vs named properties is much more clear) and it would simplify the massive amount of parameters going into each function

config
The-Quill: yeah.... it was a shoddy commit, I thought I reverted it... on another note, perhaps we should be adding the config file to gitignore

findspam.py
The-Quill: It reduces the indentation, I suppose. I don't mind for smaller blocks, but for massive blocks it makes a bit of a difference

chatcommunicate.py
```diff
@@ -147,21 +170,34 @@ def watcher(ev, wrap2):
...
+                ev.message.reply(result[1], False)
+        if result[0] is False:
             add_to_listen_if_edited(wrap2.host, message_id)
 
 
 def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user_id, ev_user_name, wrap2, content, message_id):
-    message_url = "//chat." + wrap2.host + "/transcript/message/" + str(message_id)
+    message_url = "//chat.{host}/transcript/message/{id}#{id}".format(
+        host=wrap2.host,
+        id=str(message_id)
```
kevin-brown: `string.format` should be able to do automatic casting of integers to strings, is there any reason why we aren't using that here (and other places)?

bodyfetcher.py
The-Quill: @AWegnerGitHub I don't understand why this is a thing here.... `/me shrugs`

chatcommunicate.py
The-Quill: @AWegnerGitHub human failure, it seems. fixed in the latest

continuousintegration.py
The-Quill: Fair enough @ArtOfCode- I refactored the API url a little better in the latest

spamhandling.py
```diff
@@ -17,11 +17,8 @@ def should_whitelist_prevent_alert(user_url, reasons):
     if not is_whitelisted:
         return False
     reasons_copy = list(set(reasons))
-    reasons_copy_copy = list(reasons_copy)
-    for reason in reasons_copy:
-        if "username" in reason:
-            reasons_copy_copy.remove(reason)
-    return len(reasons_copy_copy) == 0
+    reasons_comparison = [r for r in list(reasons_copy) if "username" not in r]
+    return len(reasons_comparison) == 0
```
kevin-brown: Wouldn't it make sense to use [`any`](http://stackoverflow.com/q/19389490/359284#19389957) here?

```python
return any(r for r in set(reasons) if "username" not in r)
```

Also, you shouldn't need to call `list` if you are making it a `set` ahead of time.

The-Quill: That's code from before me <_<  I'll take a look though

bodyfetcher.py
hichris1234: The whole point of this was debugging in the beginning to note that nothing occurred here. (see https://en.wikipedia.org/wiki/NOP) Likely, we can just remove this, and make it a `pass` as @AWegnerGitHub suggests.

chatcommunicate.py
The-Quill: @hichris1234 in these specific cases IMO it looks nicer to have the named parameters on individuals lines, I haven't run a line length check over every line, I just refactored each case that was either reaaaaaaaally long (like 200~ long) or could look better

hichris1234: @The-Quill This line (and quite a few others) were only ~120, I don't see a reason to change these.

chatcommunicate.py
```diff
@@ -269,12 +306,19 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-                return "Recorded answer as an NAA in metasmoke."
-            else:
+            if quiet_action:
                 return None
-        if (second_part_lower.startswith("delete") or second_part_lower.startswith("remove") or second_part_lower.startswith("gone") or second_part_lower.startswith("poof") or
-                second_part_lower == "del") and is_privileged(ev_room, ev_user_id, wrap2):
+
+            return "Recorded answer as an NAA in metasmoke."
+
+        if (
```
hichris1234: Leave this all on one line -- it looks pretty ugly this way

chatcommunicate.py
The-Quill: I don't share the same opinion about the `Thread(target=` style indenting on one line, but if that's the way you want it

parsing.py
```diff
@@ -153,10 +153,12 @@ def url_to_shortlink(url):
         return url
     if id_and_site[2] == "question":
         return "http://{}/questions/{}".format(id_and_site[1], id_and_site[0])
-        # We're using "/questions" and not "/q" here because when the URL
-        # is made protocol-relative, /q would redirect to http even if the
-        # shortlink is https. Same for /a. But there we still use /a because
-        # there is no /answers or something like that.
+        """
```
hichris1234: Don't use a docstring here, that pretty much goes against PEP8

chatcommunicate.py
The-Quill: Yeah... flake8 was really digging into me there if I changed it from a oneliner <_<

spamhandling.py
```diff
@@ -103,8 +101,21 @@ def handle_spam(title, body, poster, site, post_url, poster_url, post_id, reason
             username = poster.strip()
             user_link = poster_url
 
-        t_metasmoke = Thread(target=Metasmoke.send_stats_on_post,
-                             args=(title, post_url, reason.split(", "), body, username, user_link, why, owner_rep, post_score, up_vote_count, down_vote_count))
+        t_metasmoke = Thread(
+            target=Metasmoke.send_stats_on_post,
+            args=(
```
The-Quill: @hichris1234 do you have a problem with the multiline formatting here?

Undo1: @The-Quill  Personally, I don't - but as hichris seems to care more about style, I'll leave it to him.

hichris1234: Yeah, seems much better in the original. Especially as the original doesn't use one line per argument...

globalvars.py
```diff
@@ -20,194 +20,210 @@ class GlobalVars:
     charcoal_room_id = "11540"
     meta_tavern_room_id = "89"
     socvr_room_id = "41570"
-    blockedTime = {"all": 0, charcoal_room_id: 0, meta_tavern_room_id: 0, socvr_room_id: 0}
+    blockedTime = {
```
hichris1234: Keep this as it was, the one liner looks better

globalvars.py
```diff
@@ -20,194 +20,210 @@ class GlobalVars:
...
+            "2025923",  # Tushar
+            "5292302",  # Petter Friberg
+            "792066",   # Braiam
+            "5666987",  # Ian
+            "3160466",  # ArtOfCode
+            "5735775",  # Ashish Ahuja
+            "3476191"   # Nobody Nada
+        ]
+    }
+    smokeDetector_user_id = {
```
hichris1234: Ditto for this and the ones below

spamhandling.py
```diff
@@ -103,8 +101,7 @@ def handle_spam(title, body, poster, site, post_url, poster_url, post_id, reason
             username = poster.strip()
             user_link = poster_url
 
-        t_metasmoke = Thread(target=Metasmoke.send_stats_on_post,
```
hichris1234: Looks like you went a little too far. :) We had two lines for these and the ones in ws.py

globalvars.py
The-Quill: So you want me to add the indentation back again?

spamhandling.py
The-Quill: I'm a little confused here? Do you want me to A) break consistency and return it to how it was before or B) leave it as is?

globalvars.py
hichris1234: Oh, no, I meant https://github.com/Charcoal-SE/SmokeDetector/pull/159#discussion_r64142822 (making `smokeDetector_user_id` and the ones below one-liners)

spamhandling.py
hichris1234: Leave it as it was - we broke it since the arguments are pretty long so it makes it easier to read.

metasmoke.py
```diff
@@ -81,12 +98,21 @@ def send_status_ping(self):
                 json_response = response.json()
                 commit_response = json_response["commit_status"]
                 if commit_response["status"] == "success":
+                    autopull_message = "Message contains 'autopull', pulling..."
```
hichris1234: This seems really confusing to read. `{0} succeeded!` is... not the easiest to parse. Perhaps leaving it as it was is best.

findspam.py
```diff
@@ -94,7 +97,11 @@ def has_phone_number(s, site):
             try:
                 z = phonenumbers.parse(phone_number, testf)
                 if phonenumbers.is_possible_number(z) and phonenumbers.is_valid_number(z):
-                    print "Possible {}, Valid {}, Explain: {}".format(phonenumbers.is_possible_number(z), phonenumbers.is_valid_number(z), z)
+                    print "Possible {}, Valid {}, Explain: {}".format(
```
hichris1234: Keep this as a one liner

findspam.py
```diff
@@ -39,11 +41,12 @@ def has_few_characters(s, site):
 
 def has_repeating_characters(s, site):
     s = regex.sub('http[^"]*', "", s)    # remove URLs for this check
-    if s is None or len(s) == 0 or len(s) >= 300 or regex.compile("<pre>|<code>").search(s):
+    if s is None or len(s) == 0 or \
```
hichris1234: Keep this as a one liner

findspam.py
```diff
@@ -6,7 +6,9 @@
 def all_caps_text(s, site):
     s = regex.sub("<[^>]*>", "", s)   # remove HTML tags
     s = regex.sub("&\w+;", "", s)     # remove HTML entities
-    if len(s) <= 150 and regex.compile(ur"SQL|\b(ERROR|PHP|QUERY|ANDROID|CASE|SELECT|HAVING|COUNT|GROUP|ORDER BY|INNER|OUTER)\b").search(s):
```
hichris1234: Keep this as a one liner

excepthook.py
```diff
@@ -14,15 +14,23 @@ def uncaught_exception(exctype, value, tb):
...
-    logged_msg = exception_only + '\n' + str(now) + " UTC" + '\n' + tr + '\n\n'
+    logged_msg = "{exception}\n{now} UTC\n{row}\n\n".format(
+        exception=exception_only,
+        now=now,
+        row=tr
+    )
     print(logged_msg)
     with open("errorLogs.txt", "a") as f:
         f.write(logged_msg)
+    exit_code = 1
```
hichris1234: Keep this as it was, it's much more explicit and readable in the original

excepthook.py
```diff
@@ -14,15 +14,23 @@ def uncaught_exception(exctype, value, tb):
     seconds = delta.total_seconds()
     tr = '\n'.join((traceback.format_tb(tb)))
     exception_only = ''.join(traceback.format_exception_only(exctype, value)).strip()
-    logged_msg = exception_only + '\n' + str(now) + " UTC" + '\n' + tr + '\n\n'
```
hichris1234: Neither of these look particularly clean... I have no clue what to do here. :P

datahandling.py
```diff
@@ -298,8 +297,12 @@ def refresh_sites():
     has_more = True
     page = 1
     while has_more:
-        resp = requests.get("https://api.stackexchange.com/2.2/sites?filter=!%29Qpa1bTB_jCkeaZsqiQ8pDwI&pagesize=500&page=" + str(page) + "&key=IAkbitmze4B8KpacUfLqkw((")
-        data = json.loads(resp.text)
```
hichris1234: Keep this as it was, don't worry about 80 chars

continuousintegration.py
```diff
@@ -24,7 +24,9 @@ def watch_ci():
     s.listen(10)
     print 'listening for CI changes'
 
-    while 1:
+    api_url = 'https://api.github.com/repos/Charcoal-SE/SmokeDetector/'
```
hichris1234: Don't do this, it looks really confusing to me. I really have a hard time parsing the `{0}` stuff.

chatcommunicate.py
```diff
@@ -532,12 +589,25 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
         GlobalVars.users_chatting[ev_room] = list(set(GlobalVars.users_chatting[ev_room]))  # Make unique
         user_to_blame = random.choice(GlobalVars.users_chatting[ev_room])
         return u"It's [{}]({})'s fault.".format(user_to_blame[0], user_to_blame[1])
-    if "smokedetector" in content_lower and "fault" in content_lower and ("xkcdbot" in ev_user_name.lower() or "bjb568" in ev_user_name.lower()):
+    if "smokedetector" in content_lower and "fault" in content_lower \
```
hichris1234: Keep this as a one liner (although we should probably just delete it...)

chatcommunicate.py
```diff
@@ -502,11 +550,17 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
             post_message_in_room(ev_room, logs_part, False)
     if content_lower.startswith("!!/pull"):
         if is_privileged(ev_room, ev_user_id, wrap2):
-            r = requests.get('https://api.github.com/repos/Charcoal-SE/SmokeDetector/git/refs/heads/master')
```
hichris1234: Keep this as it was, the lines even fit in the browser page so I don't think it's too long

chatcommunicate.py
```diff
@@ -430,22 +472,28 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
         if ev_room == GlobalVars.charcoal_room_id:
             return 'Of course'
         elif ev_room == GlobalVars.meta_tavern_room_id or ev_room == GlobalVars.socvr_room_id:
-            return random.choice(['Yup', 'You doubt me?', 'Of course', '... did I miss something?',
```
hichris1234: Keep this as it was (again, don't worry about 80 chars)

chatcommunicate.py
```diff
@@ -412,16 +442,28 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
             batch = ""
             if len(urls) > 1:
                 batch = " (batch report: post {} out of {})".format(index, len(urls))
-            handle_spam(post_data.title, post_data.body, post_data.owner_name, post_data.site, post_data.post_url,
-                        post_data.owner_url, post_data.post_id, ["Manually reported " + post_data.post_type + batch],
-                        post_data.post_type == "answer", why, post_data.owner_rep, post_data.score, post_data.up_vote_count,
-                        post_data.down_vote_count, post_data.question_id)
+            handle_spam(
```
hichris1234: 4 lines is much easier to read than however many this was... keep it as it was

chatcommunicate.py
```diff
@@ -147,21 +164,34 @@ def watcher(ev, wrap2):
...
+            message_with_reply = u":{} {}".format(message_id, result[1])
+            if len(message_with_reply) <= 500 or "\n" in result[1]:
+                ev.message.reply(result[1], False)
+        if result[0] is False:
             add_to_listen_if_edited(wrap2.host, message_id)
 
 
 def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user_id, ev_user_name, wrap2, content, message_id):
-    message_url = "//chat." + wrap2.host + "/transcript/message/" + str(message_id)
+    message_url = "//chat.{host}/transcript/message/{id}#{id}".format(
```
hichris1234: Make this a one liner but keep the format part, I like it

chatcommunicate.py
```diff
@@ -147,21 +164,34 @@ def watcher(ev, wrap2):
             if len(message_with_reply) <= 500 or "\n" in reply:
                 ev.message.reply(reply, False)
     else:
-        r = handle_commands(content_source.lower(), message_parts, ev_room, ev_room_name, ev_user_id, ev_user_name, wrap2, content_source, message_id)
-        if type(r) != tuple:
-            r = (True, r)
-        if r[1] is not None:
+        result = handle_commands(
```
hichris1234: Keep this as a one liner

chatcommunicate.py
```diff
@@ -93,7 +92,10 @@ def watcher(ev, wrap2):
     message_parts = content_source.split(" ")
 
     ev_user_name = ev.data["user_name"]
-    ev_user_link = "//chat." + wrap2.host + "/users/" + str(ev.user.id)
+    ev_user_link = "//chat.{host}/users/{user_id}".format(
```
hichris1234: Make this a one liner but keep using `format`

bodyfetcher.py
```diff
@@ -145,25 +162,38 @@ def make_api_call_for_site(self, site):
...
+                        .replace('.stackexchange', '') + ": {0}\n".format(str(quota_used))
                 api_quota_used_per_site = api_quota_used_per_site.strip()
                 GlobalVars.charcoal_hq.send_message(api_quota_used_per_site, False)
                 clear_api_data()
             if response["quota_remaining"] == 0:
                 GlobalVars.charcoal_hq.send_message("API reports no quota left!  May be a glitch.")
                 GlobalVars.charcoal_hq.send_message(str(response))  # No code format for now?
             if GlobalVars.apiquota == -1:
-                GlobalVars.charcoal_hq.send_message("Restart: API quota is {}.".format(response["quota_remaining"]))
+                GlobalVars.charcoal_hq.send_message(
```
hichris1234: Keep the one liner here

bodyfetcher.py
```diff
@@ -145,25 +162,38 @@ def make_api_call_for_site(self, site):
...
+                    "Current quota: {0}.".format(response["quota_remaining"])
+                )
+                sorted_calls_per_site = sorted(
+                    GlobalVars.api_calls_per_site.items(),
+                    key=itemgetter(1),
+                    reverse=True
+                )
                 api_quota_used_per_site = ""
                 for site_name, quota_used in sorted_calls_per_site:
-                    api_quota_used_per_site = api_quota_used_per_site + site_name.replace('.com', '').replace('.stackexchange', '') + ": " + str(quota_used) + "\n"
```
hichris1234: Keep the one liner here

bodyfetcher.py
```diff
@@ -145,25 +162,38 @@ def make_api_call_for_site(self, site):
         message_hq = ""
         if "quota_remaining" in response:
             if response["quota_remaining"] - GlobalVars.apiquota >= 5000 and GlobalVars.apiquota >= 0:
-                GlobalVars.charcoal_hq.send_message("API quota rolled over with {} requests remaining. Current quota: {}.".format(GlobalVars.apiquota, response["quota_remaining"]))
```
hichris1234: Keep using one liners on both of these

bodyfetcher.py
```diff
@@ -124,9 +134,16 @@ def make_api_call_for_site(self, site):
                 pagesize = "50"
             else:
                 pagesize = "25"
-            url = "http://api.stackexchange.com/2.2/questions?site=stackoverflow&filter=!)E0g*ODaEZ(SgULQhYvCYbu09*ss(bKFdnTrGmGUxnqPptuHP&key=IAkbitmze4B8KpacUfLqkw((&pagesize=" + pagesize + min_query
+            pagesize_modifier = "&pagesize={0}{1}".format(pagesize, min_query)
         else:
-            url = "http://api.stackexchange.com/2.2/questions/" + ";".join(str(x) for x in posts) + "?site=" + site + "&filter=!)E0g*ODaEZ(SgULQhYvCYbu09*ss(bKFdnTrGmGUxnqPptuHP&key=IAkbitmze4B8KpacUfLqkw(("
+            question_modifier = "/{0}".format(";".join(str(x) for x in posts))
```
hichris1234: I don't really think this looks any better...

bodyfetcher.py
```diff
@@ -124,9 +134,16 @@ def make_api_call_for_site(self, site):
                 pagesize = "50"
             else:
                 pagesize = "25"
-            url = "http://api.stackexchange.com/2.2/questions?site=stackoverflow&filter=!)E0g*ODaEZ(SgULQhYvCYbu09*ss(bKFdnTrGmGUxnqPptuHP&key=IAkbitmze4B8KpacUfLqkw((&pagesize=" + pagesize + min_query
+            pagesize_modifier = "&pagesize={0}{1}".format(pagesize, min_query)
```
hichris1234: Keep this as it was, I'm quite confused with `{0}{1}`

bodyfetcher.py
```diff
@@ -12,45 +12,50 @@
...
+        "patents.stackexchange.com": 1,
+        "pets.stackexchange.com": 1,
+        "startups.stackexchange.com": 1,
+        "travel.stackexchange.com": 1,
+        "webapps.stackexchange.com": 1,
+        "woodworking.stackexchange.com": 1,
+        "writers.stackexchange.com": 1
+    }
+
+    time_sensitive = [
```
hichris1234: Keep this as it was, don't worry about 80 chars

chatcommunicate.py
```diff
@@ -119,18 +127,27 @@ def watcher(ev, wrap2):
             current_message = shortcut_messages[i]
             if length > 1:
                 reply += str(i + 1) + ". "
-            reply += "[" + current_message.split(" ")[0] + "] "
+            reply += "[{0}] ".format(current_message.split(" ")[0])
             if current_message.split(" ")[1] != "-":
-                r = handle_commands(current_message.lower(), current_message.split(" "), ev_room, ev_room_name, ev_user_id, ev_user_name, wrap2, current_message, message_id)
-                result = r
+                result = handle_commands(
```
hichris1234: Keep the one liner here

chatcommunicate.py
```diff
@@ -532,12 +589,25 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-    if "smokedetector" in content_lower and "fault" in content_lower and ("xkcdbot" in ev_user_name.lower() or "bjb568" in ev_user_name.lower()):
+    if "smokedetector" in content_lower and "fault" in content_lower \
+            and ("xkcdbot" in ev_user_name.lower() or "bjb568" in ev_user_name.lower()):
+
         return "Liar"
     if content_lower.startswith("!!/coffee"):
         return "*brews coffee for @" + ev_user_name.replace(" ", "") + "*"
     if content_lower.startswith("!!/tea"):
-        return "*brews a cup of " + random.choice(['earl grey', 'green', 'chamomile', 'lemon', 'darjeeling', 'mint', 'jasmine']) + " tea for @" + ev_user_name.replace(" ", "") + "*"
+        return "*brews a cup of {choice} tea for @{user}*".format(
```
hichris1234: Keep this as a one liner but keep using `format()`

bodyfetcher.py
kevin-brown: I think this would have been better if

1. The arguments to `format` where named, so we weren't dealing with `{0}` and `{1}`
2. It was somehow made more clear what `min_query` is and why it is being put at the very end of this

As it currently stands, I think most people could figure out what `&pagesize={0}` does. The trailing `{1}` is strange though.

chatcommunicate.py
```diff
@@ -595,19 +665,24 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-            if not quiet_action:
-                return "You'll now get pings from me if I report a post on `%s`, in room `%s` on `chat.%s`" % (full_site, room_id, chat_site)
-            else:
+        response, full_site = add_to_notification_list(user_id, chat_site, room_id, se_site)
+        if response == 0:
+            if quiet_action:
                 return None
-        elif r == -1:
+
+            return "You'll now get pings from me if I report a post "\
```
hichris1234: This can probably be made into one line... I think...

chatcommunicate.py
```diff
@@ -617,18 +692,21 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-            if not quiet_action:
-                return "I will no longer ping you if I report a post on `%s`, in room `%s` on `chat.%s`" % (se_site, room_id, chat_site)
-            else:
+        response = remove_from_notification_list(user_id, chat_site, room_id, se_site)
+        if response:
+            if quiet_action:
                 return None
-        else:
-            return "That configuration doesn't exist."
+            return "I will no longer ping you if I report a post "\
```
hichris1234: Ditto here with making it one line

datahandling.py
```diff
@@ -328,50 +331,53 @@ def check_site_and_get_full_name(site):
...
 
 def get_all_notification_sites(user_id, chat_site, room_id):
-    user_id = int(user_id)
-    room_id = int(room_id)
     sites = []
-    for n in GlobalVars.notifications:
-        if n[0] == user_id and n[1] == chat_site and n[2] == room_id:
-            sites.append(n[3])
+    for notification in GlobalVars.notifications:
+        if notification[0] == int(user_id) \
```
hichris1234: Keep this as a one liner

datahandling.py
```diff
@@ -328,50 +331,53 @@ def check_site_and_get_full_name(site):
...
     return sites
 
 
 def get_user_ids_on_notification_list(chat_site, room_id, se_site):
     uids = []
-    for t in GlobalVars.notifications:
-        if t[1] == chat_site and t[2] == int(room_id) and t[3] == se_site:
-            uids.append(t[0])
+    for notification in GlobalVars.notifications:
+        if notification[1] == chat_site \
```
hichris1234: Ditto here with keeping it on one line

deletionwatcher.py
```diff
@@ -44,23 +44,23 @@ def check_websocket_for_deletion(self, post_site_id, post_url, timeout):
...
                 return False
             if a is not None and a != "":
                 try:
                     d = json.loads(json.loads(a)["data"])
                 except:
                     continue
-                if d["a"] == "post-deleted" and str(d["qId"]) == question_id and ((post_type == "answer" and "aId" in d and str(d["aId"]) == post_id) or post_type == "question"):
-                    t_metasmoke = Thread(target=Metasmoke.send_deletion_stats_for_post,
-                                         args=(post_url, True))
+                if d["a"] == "post-deleted" \
```
hichris1234: Probably good to split this one over two lines, but you can probably make line 56 part of line 55 and still keep it readable

parsing.py
```diff
@@ -153,10 +153,12 @@ def url_to_shortlink(url):
         return url
     if id_and_site[2] == "question":
         return "http://{}/questions/{}".format(id_and_site[1], id_and_site[0])
+
```
hichris1234: Looks like some extra whitespace crept in here

spamhandling.py
```diff
@@ -36,7 +32,11 @@ def check_if_spam(title, body, user_name, user_url, post_site, post_id, is_answe
             blacklisted_post_url = blacklisted_user_data[2]
             if blacklisted_post_url:
                 rel_url = blacklisted_post_url.replace("http:", "", 1)
-                why += u"\n" + u"Blacklisted user - blacklisted for {} (http://metasmoke.erwaysoftware.com/posts/by-url?url={}) by {}".format(blacklisted_post_url, rel_url, message_url)
+                why += u"\nBlacklisted user - blacklisted for {} (http://metasmoke.erwaysoftware.com/posts/by-url?url={}) by {}".format(
```
hichris1234: Keep this as all one line

datahandling.py
```diff
@@ -298,8 +297,12 @@ def refresh_sites():
     has_more = True
     page = 1
     while has_more:
-        resp = requests.get("https://api.stackexchange.com/2.2/sites?filter=!%29Qpa1bTB_jCkeaZsqiQ8pDwI&pagesize=500&page=" + str(page) + "&key=IAkbitmze4B8KpacUfLqkw((")
-        data = json.loads(resp.text)
+        response = requests.get(
+            "https://api.stackexchange.com/2.2/sites?" +
+            "filter=!%29Qpa1bTB_jCkeaZsqiQ8pDwI&pagesize=500&page={}".format(str(page)) +
```
kevin-brown: If, after this is reverted, you continue using `format`, you don't need `str` here.

excepthook.py
The-Quill: ...

chatcommunicate.py
The-Quill: Delete it?

bodyfetcher.py
The-Quill: There was also a formatting issue here somewhere that I think this fixed... although I'm not 100% that it fixed it <_<

chatcommunicate.py
hichris1234: Yeah, go ahead (this whole `if` block)

parsing.py
The-Quill: ./parsing.py:165:1: E302 expected 2 blank lines, found 1

hichris1234: Flake seemed to be okay with it before... so it doesn't pass if you remove the two newlines?

The-Quill: won't pass without the two lines, I'll do a sample commit for you with them missing

chatcommunicate.py
The-Quill: done, but not shown for some reason

bodyfetcher.py
The-Quill: Fixed, but not shown

bodyfetcher.py
The-Quill: Fixed but not shown

chatcommunicate.py
The-Quill: Fixed but not shown

chatcommunicate.py
The-Quill: Fixed but not shown

continuousintegration.py
The-Quill: I refactored it a little to use the named parameters, see if you like that, otherwise I'll return it

metasmoke.py
The-Quill: Changed this to use named parameters, see if you like it, otherwise I'll return it

parsing.py
The-Quill: https://github.com/Charcoal-SE/SmokeDetector/pull/159/commits/e84f77641e4582837a51b47844c5d0844ea011e8

datahandling.py
The-Quill: fixed but not shown

parsing.py
hichris1234: Oh, we're referring to two different places. I'm taking about above & below the block comment, that whitespace. The marker must have shifted with the changes (or I placed it wrong...)

The-Quill: The marker shifted a few lines ahead with the changes

hichris1234:
In general this seems to look much better, I'd like to take another look in the morning though. (Probably the others would like to weigh in too)

Also, note to self/whoever pulls this in: make sure to squash the commits.

bodyfetcher.py
hichris1234: Can you make `sorted_calls_per_site` a one liner again too?

chatcommunicate.py
```diff
@@ -106,10 +105,16 @@ def watcher(ev, wrap2):
...
             return
         if len(commands) > len(latest_smokedetector_messages):
-            ev.message.reply("I've only posted {} messages since the latest reboot; that's not enough to execute all commands. No commands were executed.".format(len(latest_smokedetector_messages)))
+            ev.message.reply(
+                "I've only posted {} messages".format(len(latest_smokedetector_messages)) +
+                "since the latest reboot; that's not enough to execute all commands. No commands were executed."
+            )
             return
         for i in xrange(0, len(commands)):
-            shortcut_messages.append(":" + str(latest_smokedetector_messages[-(i + 1)]) + " " + commands[i])
```
hichris1234: Maybe use named modifiers here, and keep it all on one line

chatcommunicate.py
```diff
@@ -232,25 +237,23 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
+                    return "Blacklisted user."
+                return "Recorded answer as true positive in metasmoke. If you want to blacklist the poster of the answer, use `trueu` or `tpu`."
+
         if second_part_lower.startswith("ignore") and is_privileged(ev_room, ev_user_id, wrap2):
             if post_site_id is None:
                 return "That message is not a report."
 
-            t_metasmoke = Thread(target=Metasmoke.send_feedback_for_post,
-                                 args=(post_url, second_part_lower, ev_user_name, ev_user_id, ))
+            t_metasmoke = Thread(target=Metasmoke.send_feedback_for_post, args=(post_url, second_part_lower, ev_user_name, ev_user_id,))
```
hichris1234: Keep this as a two liner, looks prettier to me. :)

chatcommunicate.py
```diff
@@ -382,10 +385,12 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
             and is_privileged(ev_room, ev_user_id, wrap2):
         crn, wait = can_report_now(ev_user_id, wrap2.host)
         if not crn:
-            return "You can execute the !!/report command again in {} seconds. " \
-                   "To avoid one user sending lots of reports in a few commands and slowing SmokeDetector down due to rate-limiting, " \
-                   "you have to wait 30 seconds after you've reported multiple posts using !!/report, even if your current command just has one URL. " \
-                   "(Note that this timeout won't be applied if you only used !!/report for one post)".format(wait)
+            return "You can execute the !!/report command again in {} seconds. ".format(wait) + \
```
hichris1234: Keep this on 4 lines (I bet this was limiting to 80 chars)

chatcommunicate.py
```diff
@@ -430,22 +434,22 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-            now = datetime.utcnow()
-            diff = now - UtcDate.startup_utc_date
-            minutes, remainder = divmod(diff.seconds, 60)
-            minutestr = "minutes" if minutes != 1 else "minute"
-            return 'Running since {} UTC ({} {})'.format(GlobalVars.startup_utc, minutes, minutestr)
+        now = datetime.utcnow()
+        diff = now - UtcDate.startup_utc_date
+        minutes, remainder = divmod(diff.seconds, 60)
+        minute_str = "minutes" if minutes != 1 else "minute"
+        return 'Running since {} UTC ({} {})'.format(
```
hichris1234: This can probably stay all on one line

continuousintegration.py
hichris1234: I think it looks better the original way, since it's only two links I don't think we need to factor out the API URL into a variable

metasmoke.py
hichris1234: Looks better, but I think something was transposed here (you have `status` and then `{success}`. Also, can you move L104-107 to L103 and L109-111 to L108?

chatcommunicate.py
The-Quill: done but not shown for some reason

parsing.py
hichris1234: Can you go ahead and remove the whitespace around the block comment?

excepthook.py
hichris1234: I think it'll look better if this is all on one line, but keep using format/named parameters

chatcommunicate.py
```diff
@@ -617,18 +619,20 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-            if not quiet_action:
-                return "I will no longer ping you if I report a post on `%s`, in room `%s` on `chat.%s`" % (se_site, room_id, chat_site)
-            else:
+        response = remove_from_notification_list(user_id, chat_site, room_id, se_site)
+        if response:
+            if quiet_action:
                 return None
-        else:
-            return "That configuration doesn't exist."
+            return "I will no longer ping you if I report a post on `{site_name}`, in room `{room_id}` on `chat.{chat_domain}`".format(
```
hichris1234: Can you put this all on one line?

chatcommunicate.py
hichris1234: It doesn't look like this made it into a commit?

chatcommunicate.py
```diff
@@ -430,22 +430,18 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
-                                  'Watching this endless list of new questions *never* gets boring',
-                                  'Kinda sorta'])
+            return random.choice(['Yup', 'You doubt me?', 'Of course', '... did I miss something?', 'plz send teh coffee', 'Watching this endless list of new questions *never* gets boring', 'Kinda sorta'])
     if content_lower.startswith("!!/rev") or content_lower.startswith("!!/ver"):
-            return '[' + \
-                GlobalVars.commit_with_author + \
-                '](https://github.com/Charcoal-SE/SmokeDetector/commit/' + \
-                GlobalVars.commit + \
-                ')'
+        return '[{0}](https://github.com/Charcoal-SE/SmokeDetector/commit/{1})'.format(
```
hichris1234: Move this to all on one line (but keep the format part)

bodyfetcher.py
```diff
@@ -115,18 +119,29 @@ def make_api_call_for_site(self, site):
...
             # fetch all recently modified SO questions to work around that.
-            min_query = ""
             if self.last_activity_date != 0:
-                min_query = "&min=" + str(self.last_activity_date)
                 pagesize = "50"
             else:
                 pagesize = "25"
-            url = "http://api.stackexchange.com/2.2/questions?site=stackoverflow&filter=!)E0g*ODaEZ(SgULQhYvCYbu09*ss(bKFdnTrGmGUxnqPptuHP&key=IAkbitmze4B8KpacUfLqkw((&pagesize=" + pagesize + min_query
+
+            pagesize_modifier = "&pagesize={pagesize}&min={time_length}".format(
```
hichris1234: This can be moved all to one line

excepthook.py
The-Quill: Fixed

PR #162, opened 2016-06-10
Adds named args to several function calls

AWegnerGitHub:
This updates calls to `handle_spam`, `check_if_spam`, `post_message_in_room` and `handle_commands` to use named parameters instead of positional parameters.


Undo1:
I really like this, is a lot more readable to me. Let's see what the others think.

ProgramFOX:
+1 from me

ArtOfCode-:
+11 from me

Undo1:
If there are no major objections by noon US time, I'm going to very strong consider (read: will) merging this.

normalhuman:
belated +1

PR #164, opened 2016-06-16
Added QPaysTaxes to the SOCVR privileged user list

K-Guan:


newbiedoodle:
Yay :D (for the record, this is QPaysTaxes)

Issue #166, opened 2016-06-16
Clean up handle_commands discussion


AWegnerGitHub:
# Summary

The [`handle_commands`](https://github.com/Charcoal-SE/SmokeDetector/blob/aaf332daa8588394af555357d2bd1da32966a234/chatcommunicate.py#L170) function is a 500 line monstrosity. I'd like to discuss cleaning it up to do the following:

- Standardize what is returned from `handle_commands`
- Have individual functions for each command. 
 
If we agree that this work is needed, I'll take on the work load over the next several weeks (translation: after my summer vacation) to do this work.
 
# Clean up details

This section describes my plan for the above changes. I'd like to do both changes at the same time, but if opposition to that exists one change at a time is acceptable to me.

## Standardize what is returned from `handle_commands`

Currently, the function has multiple return paths and multiple ways of returning data. 

One example is in the [`!!/block`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommunicate.py#L491) command:

    return False, "Invalid duration."
	
Another example is in [`!!/errorlogs`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommunicate.py#L513)

    return "The !!/errorlogs command requires 1 argument."
	
These two methods lead to weird handling in the two locations that `handle_commands` is called from. ([Location 1](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommunicate.py#L123), [Location 2](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommunicate.py#L157))

Method of handling 1:

    if type(result) != tuple:
        result = (True, result)
    if result[1] is not None:
        if wrap2.host + str(message_id) in GlobalVars.listen_to_these_if_edited:
            GlobalVars.listen_to_these_if_edited.remove(wrap2.host + str(message_id))
        message_with_reply = u":{} {}".format(message_id, result[1])
        if len(message_with_reply) <= 500 or "\n" in result[1]:
            ev.message.reply(result[1], False)
    if result[0] is False:
        add_to_listen_if_edited(wrap2.host, message_id)

Method of handling 2:

    r = result
    if type(result) == tuple:
        result = result[1]
    if result is not None and result is not False:
        reply += result + os.linesep
    elif result is None:
        reply += "<processed without return value>" + os.linesep
        amount_none += 1
    elif result is False or r[0] is False:
        reply += "<unrecognized command>" + os.linesep
[...]

PR #168, opened 2016-06-21
Phase 1 of handle_commands clean up

AWegnerGitHub:
This relates to #166. 

This is phase 1 - cleaning up `handle_commands` to be more manageable. This change replaces the massive function that existed with two `try`/`except` blocks. Once block is for the `sd <response>` commands and the other is for any other `!!/*` commands. 

Any required parameters are passed to the functions as part of the `*_parameters` dictionaries. 

All commands themselves have been put in `chatcommands.py`. This file is rather large, but once these are broken out into functions, I can work on eliminating several areas of duplication. 

One such duplication has already been removed - the logging to Metasmoke has been moved to a single function call within the appropriate commands, instead of each command needing to set up a Thread.

Prior to merging, I'd appreciate feedback. I have done testing, but more eyes are always good.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -2,12 +2,20 @@ from globalvars import GlobalVars
 from findspam import FindSpam
 from datetime import datetime
 from utcdate import UtcDate
-from datahandling import is_privileged
-from datahandling import will_i_be_notified
-from datahandling import get_all_notification_sites
-from datahandling import remove_from_notification_list
+from datahandling import add_blacklisted_user
 from datahandling import add_to_notification_list
+from datahandling import add_whitelisted_user
+from datahandling import fetch_lines_from_error_log
+from datahandling import get_all_notification_sites
+from datahandling import is_blacklisted_user
+from datahandling import is_privileged
+from datahandling import is_whitelisted_user
+from datahandling import remove_from_notification_list
+from datahandling import remove_blacklisted_user
+from datahandling import remove_whitelisted_user
+from datahandling import will_i_be_notified
 from chatcommunicate import post_message_in_room
+from parsing import get_user_from_list_command
 from parsing import get_user_from_url
 from spamhandling import handle_user_with_all_spam
 import random
@@ -19,7 +27,7 @@ import time
 # TODO: pull out code block to get user_id, chat_site, room_id into function
 # TODO: Order functions by groups
 # TODO: Return result for all functions should be similar (tuple/named tuple?)
-# TODO: Do we need uid == -2 check?
+# TODO: Do we need uid == -2 check?  Turn into "is_user_valid" check
 
 # Functions go before the final dictionary of command to function mappings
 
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
👍

chatcommands.py
```diff
@@ -0,0 +1,935 @@
...
+
+
+# --- Blacklist Functions --- #
+def command_add_blacklist_user(*args, **kwargs):
+    """
+    Adds a user to the site blacklist
+    :param kwargs: Requires that 'content_lower', 'message_url', 'ev_room', 'ev_user_id', and 'wrap2' is passed as kwarg
+    :return: A string
+    """
+    if is_privileged(kwargs['ev_room'], kwargs['ev_user_id'], kwargs['wrap2']):
```
kevin-brown: Is there a reason why you are collecting these in `**kwargs` instead of explicitly specifying the arguments?

chatcommands.py
```diff
@@ -0,0 +1,935 @@
...
+
+#
+#
+# System command functions below here
+# Each of these should take the *args and **kwargs parameters. This allows us to create functions that
+# don't accept any parameters but still use the `command_dict` mappings
+
+
+# --- Blacklist Functions --- #
+def command_add_blacklist_user(*args, **kwargs):
```
kevin-brown: There is no need to include `*args` if it's not being used.

chatcommands.py
AWegnerGitHub: Not each function takes the exact same arguments. By passing the dictionary, instead of explicitly setting each function's parameters, I can call it like I do here:

https://github.com/AWegnerGitHub/SmokeDetector/blob/branch_chatcommands_split/chatcommunicate.py#L222

```
    # Process additional commands
    command_parameters = {
        'content': content,
        'content_lower': content_lower,
        'ev_room': ev_room,
        'ev_room_name': ev_room_name,
        'ev_user_id': ev_user_id,
        'ev_user_name': ev_user_name,
        'message_parts': message_parts,
        'message_url': message_url,
        'wrap2': wrap2,
    }
    try:
        return cmds[command](**command_parameters)
    except KeyError:
        return False, None  # Unrecognized command, can be edited later.
```

The dictionary-switch statement (because Python lacks a true `switch`) allows us to [map the commands to desired function](https://github.com/AWegnerGitHub/SmokeDetector/blob/branch_chatcommands_split/chatcommands.py#L848). This allows us to quickly add aliases for commands (ie. `commands`, `help`, `info`)

chatcommands.py
AWegnerGitHub: Fair enough. It's not being used now, but I left it in case we have (or will have) positional parameters for some of these functions in the future. If we're opposed to keeping `*args` in the function signature, I'll pull it out.

chatcommunicate.py
```diff
@@ -197,459 +196,41 @@ def handle_commands(content_lower, message_parts, ev_room, ev_room_name, ev_user
...
+        'ev_room_name': ev_room_name,
+        'ev_user_id': ev_user_id,
+        'ev_user_name': ev_user_name,
+        'message_parts': message_parts,
+        'message_url': message_url,
+        'wrap2': wrap2,
+    }
+    try:
+        return cmds[command](**command_parameters)
+    except KeyError:
```
kevin-brown: Wouldn't it make more sense to check

```python
if command in cmds:
```

Instead of attempting to catch `KeyError`, which may be raised from within a command?

AWegnerGitHub: [No](http://stackoverflow.com/a/7604717/189134). 

and

> [EAFP](https://docs.python.org/2/glossary.html)
>    Easier to ask for forgiveness than permission. This common Python coding style assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false. This clean and fast style is characterized by the presence of many try and except statements. The technique contrasts with the LBYL style common to many other languages such as C.

chatcommands.py
kevin-brown: But you aren't passing in a dictionary, you're using [keyword argument unpacking](http://agiliq.com/blog/2012/06/understanding-args-and-kwargs/) to convert the dictionary to a set of keyword arguments passed into the function.

I don't disagree with using `**kwargs` to catch the extra unused arguments, but wouldn't it make sense to explicitly declare arguments which are being used within the function? That way anyone who uses the function elsewhere, or quickly wants to see what it takes, can know instead of having to guess what "any keyword arguments" really means.

AWegnerGitHub: I understand what you are asking now. Yes. That makes sense.

chatcommunicate.py
kevin-brown: Sure, I totally get the concept of EAFP and would normally be like "yeah, that makes sense" except in this case _this is hiding potential bugs in the commands_.

I roughly estimated the number of times a dictionary lookup is done in `chatcommands.py` by doing a quick search for `['`, which appears to have under 10 false positives. I got 185 hits for that, which means there **185 chances that a command raises a `KeyError`**, which would be ignored and the user would be notified that they used an unrecognized command.

So, sure. Here's the EAFP way of doing this, such that you don't have that problem.

```python
try:
    comm = cmds[command]
except KeyError:
    return False, None
else:
    return comm(**command_parameters)
```

And, for those who are interested, the LBYL way of doing it

```python
if command not in cmds:
    return False, None

return cmds[command](**command_parameters)
```

AWegnerGitHub: Again, good point. I'll adjust the calls for both the `cmds` and `subcmds`.

AWegnerGitHub:
Any additional comments on this PR?

Undo1:
We never did this? I'm on vacation right now, but my comment is to merge
it.

PR #171, opened 2016-07-09
Allow easy changing of SmokeDetector name and repository with environment variables

### LINKED ISSUE #169: Make SmokeDetector easier to run on an alternate account - Chat message prefix variables
I've been testing SmokeDetector more on an alternate account and would like to propose a modest change to make this easier for other developers.

I'd like to put this information in [globalvars.py](https://github.com/Charcoal-SE/SmokeDetector/blob/a50e7a5e8eedf60e3d1f1c04a1ff75551b1a0d1d/globalvars.py) to make it easier for developers to quickly change this data.

- Pull `[SmokeDetector](https://github.com/Charcoal-SE/SmokeDetector)`

By pulling this out of the multiple places it is in the code, we can allow users to link to their personal Repository and change the name from the official "SmokeDetector". This makes it easier to identify and if someone is curious, they can follow to an appropriate GitHub profile. It also removes the "authentic SmokeDetector" look that a test version may be running.

I'd suggest either pulling this into a single variable like `chatmessage_prefix` that contains the entire string, or two config variables one for `bot_name` and one for `bot_repository`.

This change would affect both [ws.py](https://github.com/Charcoal-SE/SmokeDetector/blob/a50e7a5e8eedf60e3d1f1c04a1ff75551b1a0d1d/ws.py) and [chatcommands.py](https://github.com/Charcoal-SE/SmokeDetector/blob/053da32a71a0808e4938c9625fa802bd2f1a0755/chatcommands.py)

A similar variable may be needed for `https://github.com/Charcoal-SE/SmokeDetector/commit/` links (in ws.py), but to me this isn't as important as the two listed above.

Undo1:
Closes #169.

Thoughts?

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -378,9 +378,9 @@ def command_help(*args, **kwargs):
     :param kwargs: No additional arguments expected
     :return: A string
     """
-    return "I'm [SmokeDetector](https://github.com/Charcoal-SE/SmokeDetector), a bot " \
+    return "I'm " + GlobalVars.chatmessage_prefix + ", a bot " \
            "that detects spam and offensive posts on the network and posts alerts to chat. " \
-           "[A command list is available here](https://github.com/Charcoal-SE/SmokeDetector/wiki/Commands)."
+           "[A command list is available here](" + GlobalVars.bot_repository + "/wiki/Commands)."
 
 
 def command_location(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


Undo1:
This should be a nonbreaking change (fails gracefully if the environment variable isn't set), so I'm going for it.

PR #179, opened 2016-07-24
Added new regex for nigger

angussidney:
Changed the regex for nigger so it will catch nigger, nigga, and now niggur.

See [this](http://chat.stackexchange.com/transcript/message/31246377#31246377) and [this](http://chat.stackexchange.com/transcript/message/31248637#31248637) for more context.


angussidney:
Also, should we add this regex to the list of blacklisted usernames?

Undo1:
Might as well. I've got it, thank you!

PR #186, opened 2016-08-09
Added Kyll from SOCVR to Charcoal HQ List

magisch:
Already privileged in SOCVR, Long time Member of SOCVR


NickVolynkin:
@ArtOfCode- shouldn't this be labeled `privileged-users`?

PR #188, opened 2016-08-09
+ 'more info'

NickVolynkin:
Lots of recent spam posts end with "more info>>>[some link](http://example.com)". 
E.g. https://metasmoke.erwaysoftware.com/post/36840


ArtOfCode-:
These should already be caught by "link following arrow in body"

NickVolynkin:
Oh, right. Close it then )

PR #202, opened 2016-08-28
Added angussidney to tavern

angussidney:
I'm just adding myself to The Tavern as well, so I can give feedback there too.

I also added some spaces so that all of the comments line up nicely :)

[Chat.meta.se profile for reference](http://chat.meta.stackexchange.com/users/285368/angussidney)


Undo1:
Wow, I actually got one.

PR #205, opened 2016-08-30
Stop versioning config.

NickVolynkin:
The `config` file was in `.gitignore` but still under version control.
That could cause unintended adding and commiting a change in `config`,
which, having been merged into master, would lead to merge conflicts
on next pull for everybody who changed their own `config`.

Added an explicit instruction in `config.sample` about copying its
contents to a new `config` file.

### DIFF
--- a/config
+++ b/config
@@ -1,4 +1,0 @@
-[Config]
-location=location_here
-
-metasmoke_host=http://127.0.0.1:3000
--- a/config.sample
+++ b/config.sample
@@ -1,3 +1,4 @@
+# Copy this to a new file named `config`
 [Config]
 location=location_here
 


NickVolynkin:
Oh, it seems that ignored but versioned config is required to run CI...

Undo1:
Yeah, the idea is to provide a way to run Smokey out of the box, while still keeping people (me) from accidentally commit real-life credentials.

NickVolynkin:
@Undo1 it doesn't keep you from commiting real-life credentials. Git still tracks `config` because it's already in the history.

Undo1:
I run `git update-index --assume-unchanged config` to fix that, but you're right that it doesn't keep *other* people from committing the file.

Right now, the guidance is "don't do that". I suppose we could make Smokey not depend on the config file being present, then remove it entirely.

NickVolynkin:
I encountered the same problem with running autotests on Jenkins. Solved it by having `jenkins.ini` and then `cp jenkins.ini config.ini` as part of environment setup. That's quite ugly, but I have no better idea. Maybe I should ask on SO. :)

NickVolynkin:
Let this PR stay here for a while, maybe I'll find a better solution, okay?

Undo1:
[globalvars.py:277](https://github.com/Charcoal-SE/SmokeDetector/blob/master/globalvars.py#L277) is the only place the config file is used. Maybe we could just make it assume some sane defaults when the file isn't present?

ArtOfCode-:
This has been here for over two weeks now without any activity, so I'm gonna close it. If there is a better solution, it can be reopened when necessary.

PR #207, opened 2016-08-30
Plop!

dorukayhan:
Added a command that replies with "Plop!".

The problem is, since I have no idea which function retrieves commands and executes things accordingly, I don't know where to call command_plop when someone says !!/plop...

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -298,6 +298,13 @@ def command_wut(*args, **kwargs):
     :return: A string
     """
     return Response(command_status=True, message="Whaddya mean, 'wut'? Humans...")
+    
+def command_plop(*args, **kwargs):
+    """
+    Returns a string equal to "Plop!" (This is a joke command)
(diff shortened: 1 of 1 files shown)


hichris1234:
Not only does this not pass CI, I really don't think more "fun" commands are a good use of our time.

ByteCommander:
And how would this be useful? Why would we want to add that? I don't see any reason. It isn't even funny - at least to me...

Undo1:
For the sake of education, you would place this in https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L1148.

Agree that it probably isn't something we want/need, though.

ArtOfCode-:
@dorukayhan Awesome that you want to contribute, but the others have said it - we've got enough fun commands these days. Your effort is probably better spent looking for websites we can blacklist, or patterns of spam that we can pick up.

PR #209, opened 2016-09-03
Add "sex sex bojiter" to the bad keywords

dorukayhan:
What if wingding comes back under a different name?


hichris1234:
bojiter is already a bad keyword, this would be caught too.

PR #215, opened 2016-09-12
improve the "no whitespace" regex

honnza:
`[a-z]+` means "only letters", and doesn't catch answers that include digits, punctuation or non-ascii. 
Motivated by http://chat.stackexchange.com/transcript/message/32262048#32262048

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -660,7 +660,7 @@ class FindSpam:
         {'method': all_caps_text, 'all': True, 'sites': ["pt.stackoverflow.com", "ru.stackoverflow.com", "es.stackoverflow.com", "ja.stackoverflow.com", "rus.stackexchange.com"],
          'reason': "all-caps {}", 'title': False, 'body': True, 'questions': False, 'username': False, 'stripcodeblocks': True, 'body_summary': False, 'max_rep': 1, 'max_score': 0},
         # No whitespace, punctuation, or formatting in a post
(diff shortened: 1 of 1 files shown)


Undo1:
I'm not a regex wizard, but the only issue I see off the bat is unformatted code answers; I.e. `Something.foo()` as text-only in an answer would be caught. Not high-quality, but not spam either.

honnza:
good point, but that one is too short anyways. How many false positives do we expect from this change? I doubt there would be any real amount to speak of.

Undo1:
Not sure, don't really have data on it. I wonder if we could get a rough idea from SEDE...

honnza:
I'd suggest testing this in production, if there is no better way :smile:

ByteCommander:
Maybe we could use something like this to whitelist some punctuation characters that prevent this pattern from matching:

    ur"^<p>\s*[^\s+-/=_()<>!?$%&]+\s*</p>\s*$"

honnza:
Only if this more aggressive regex starts generating FPs?

dorukayhan:
How about `^\S+$`?

honnza:
I don't think SE allows whitespace-only answers. Plus, there's the
omnipresent paragraph wrapper.

Undo1:
Heck, let's try it. We can watch for false positives as a result of this and revert if necessary. 

Thank you!

PR #216, opened 2016-09-12
modify "no whitespace" detection pattern

ByteCommander:
Still match if there are whitespace in the beginning or the end of the post only. Also do not test for punctuation or formatting.


ByteCommander:
Closing because it's a dupe of https://github.com/Charcoal-SE/SmokeDetector/pull/215

PR #217, opened 2016-09-18
Someone forgot to escape the full stop

angussidney:


Undo1:
It works either way (the `.` is just interpreted as "any character"), but this is technically 'right' so why not.

PR #228, opened 2016-10-01
Blacklisted few websites

Fortunate-MAN:
https://metasmoke.erwaysoftware.com/post/42140
https://metasmoke.erwaysoftware.com/post/42131
https://metasmoke.erwaysoftware.com/post/41846

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -450,7 +450,7 @@ class FindSpam:
         "raybiztech\\.com", "cegonsoft\\.com", "technomaniya\\.com", "instantassignmenthelp\\.com\\.au", "huintech\\.com",
         "dramaonline\\.pk", "gamingustaad\\.com", "carding-wutransfer\\.com", "nulife\\.co\\.in", "fixmyprinter\\.com",
         "iseepassword\\.com", "samedaypros\\.com", "zen-promo\\.com", "seyf-educ\\.com", "moneybanker\\.no",
-        "webhax\\.xyz",
+        "webhax\\.xyz", "techhacktweaks\\.com",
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -449,7 +449,8 @@ class FindSpam:
         "onlinecorrection\\.com", "dragracerv3game\\.com", "bananakong\\.net", "himzakaz\\.net", "dropcrack\\.com",
         "raybiztech\\.com", "cegonsoft\\.com", "technomaniya\\.com", "instantassignmenthelp\\.com\\.au", "huintech\\.com",
         "dramaonline\\.pk", "gamingustaad\\.com", "carding-wutransfer\\.com", "nulife\\.co\\.in", "fixmyprinter\\.com",
-        "iseepassword\\.com", "samedaypros\\.com", "zen-promo\\.com", "seyf-educ\\.com",
+        "iseepassword\\.com", "samedaypros\\.com", "zen-promo\\.com", "seyf-educ\\.com", "moneybanker\\.no",
```
Undo1: Note to self (and others, I suppose): Should we just blacklist domains in the `.no` TLD entirely? Need to pull data on it, but it doesn't sound like a terribly credit-worthy TLD.

ArtOfCode-: I'd go with no, pun *fully* intended, on that score.

`.no` is the ccTLD for Norway; I think we should be very careful about blacklisting entire ccTLDs. Country code TLDs are, I'd think, almost always going to have plenty of legit websites.

`.xxx`, on the other hand...

PR #234, opened 2016-10-03
Move blacklisted websites to a file

Undo1:
Need to do a little more testing on this, but I don't have any reason to believe it's not ready to go.

Thoughts? This could open the door to so much more flexibility.


findspam.py
```diff
@@ -277,182 +277,9 @@ class FindSpam:
...
-        "makehimmine\\.com\\.au", "shopicheck\\.com", "hivimoore\\.com", "blogmium\\.com", "soundmagic\\.us",
-        "onlinecorrection\\.com", "dragracerv3game\\.com", "bananakong\\.net", "himzakaz\\.net", "dropcrack\\.com",
-        "raybiztech\\.com", "cegonsoft\\.com", "technomaniya\\.com", "instantassignmenthelp\\.com\\.au", "huintech\\.com",
-        "dramaonline\\.pk", "gamingustaad\\.com", "carding-wutransfer\\.com", "nulife\\.co\\.in", "fixmyprinter\\.com",
-        "iseepassword\\.com", "samedaypros\\.com", "zen-promo\\.com", "seyf-educ\\.com", "moneybanker\\.no",
-        "webhax\\.xyz", "techhacktweaks\\.com", "fluffypancakerecipe\\.com", "appinary\\.com", "aprinled\\.com",
-        "abhitricks\\.com", "atompark\\.com",
-    ]
+
+    blacklisted_websites = (line.rstrip() for line in open("blacklisted_websites.txt"))
```
SmokeDetector: There's a non-zero chance that you don't know what you're doing.

This serves to block accidental merges until Undo over there has a chance to test this more.

ArtOfCode-:
I see no obvious issues here; let's do it.

AWegnerGitHub:
Why are we putting these in a file and not in a database? We can eliminate the need for Git access completely with 

 - A database/input UI (probably with permissions wrapped around it)
 - An API call to get the updated list (and cache said list on the server Smokey runs on, in case the DB isn't available)
 - A smokey command to update said list

By doing it this way, we have even greater flexibility because we can manage everything from MS (for example), not worry about GitHub permissions, and probably expand this to handle some of the other regular expression checks as well.

hichris1234:
Hmm, but then it would be hard to just setup and get started, and/or peer into which sites were blacklisted. A database seems a bit overkill for a list like this, although it would make logical sense.
On Oct 4, 2016 9:30 AM, A Wegner <notifications@github.com> wrote:Why are we putting these in a file and not in a database? We can eliminate the need for Git access completely with 

A database/input UI (probably with permissions wrapped around it)An API call to get the updated list (and cache said list on the server Smokey runs on, in case the DB isn't available)A smokey command to update said list

By doing it this way, we have even greater flexibility because we can manage everything from MS (for example), not worry about GitHub permissions, and probably expand this to handle some of the other regular expression checks as well.

ArtOfCode-:
@AWegnerGitHub @hichris1234 The idea is that we *will* put them in a database - the metasmoke database. The file on Smokey-side is the cache, so that if metasmoke's down then Smokey still has a list. We don't need another database Smokey-side when we already have one on metasmoke, and an API call to access it already built in.

AWegnerGitHub:
@ArtOfCode- Ah. So this is just setting up the cache file then? The DB and integration comes next?

ArtOfCode-:
@AWegnerGitHub Ideally, yeah. I built the API this morning; the database table in metasmoke has existed for a few months. Now all we need is for someone to write the code for Smokey to call the API and save the results to this cache file.

AWegnerGitHub:
Well then. I have nothing to add. Carry on :)

Undo1:
@AWegnerGitHub hichris summarized it pretty well - I do want to be able to fall back on manually adding things through GitHub if metasmoke breaks / 503's. Here's my endgame for this:

 1. People get special privileges on metasmoke to add to the blacklist. 
 2. When they do that, metasmoke sends a websocket request to Smokey with the addition/deletion (or just the whole file, whatever makes sense)
 3. Smokey edits the file, commits to GitHub (or PRs if a merge is involved/messy). This *should* keep continuity between metasmoke and what's on GitHub.

ArtOfCode-:
@Undo1 We can also have a manual update command that hits the API route.

Undo1:
Indeed, but that's down the road. For now, I'm merging this.

PR #246, opened 2016-10-10
Added "datagharch.com" to blacklisted Websites

magisch:
Appeared in 10 confirmed TP reports and is currently not caught by any filter on its own.

https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title=&body=datagharch&username=&why=&site=&feedback=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search


magisch:
That was a mistake - apparently its already added and I/Wrz just used !!/test wrong. closing.

angussidney:
You forgot to add the `.com` extension, and hence when you ran !!/test, smokey didn't recognise it as a domain

PR #256, opened 2016-10-13
URL repeated at end of long message

tripleee:
This is fairly prevalent in recent spam, and recently one instance was missed.


hichris1234:
I'm concerned about how accurate this would be. It seems like a lot of posts would have the same link twice in 300 characters.

tripleee:
Sorry about the silly errors, PR updated, tests passing.

ArtOfCode-:
Merged so that metasmoke etc can collect the stats to tell us how accurate this is. It can always be removed later if it's not doing what we want it to.

PR #258, opened 2016-10-17
Make it possible to direct coffee or tea to someone other than you

QPaysTaxes:
...this time for real.

Exactly what the title says: Before, you could only give yourself tea with !!/tea. Now, you can give someone else tea. Also applies to coffee, fancy that. For example, this exchange might happen:

> QPaysTaxes: !!/tea Olaf
> SmokeDetector: ↰@QPaysTaxes *brews a cup of Southern sweet tea for Olaf*

(Except it pings people. I just didn't want to ping some random person)

But this still also happens:

> QPaysTaxes: !!/tea
> SmokeDetector: ↰@QPaysTaxes *brews a cup of Southern sweet tea for @QPaysTaxes*

Is this needed? No. Does it solve any problem? No. But hey, it's not like it interferes with core functionality, and I'm _pretty_ sure it works. 

*I'd appreciate someone going through my code and making sure it works as I intend, because I really don't know if I got it right*

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -283,8 +283,8 @@ def command_coffee(ev_user_name, message_parts, *args, **kwargs):
     :param kwargs: No additional arguments expected
     :return: A string
     """
-    if message_parts and message_parts[0]:  # list not empty and string not empty
-        giving_to = message_parts[0][1:] if message_parts[0][0] == "@" else  message_parts[0]
+    if message_parts[1:] and message_parts[1]:  # list has the required element and it's not empty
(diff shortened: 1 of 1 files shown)


newbiedoodle:
I was thinking of changing it so that if you didn't specify a person, it said *brewing tea for **you*** instead of pinging you again, but if other people really want that, it's two lines.

AWegnerGitHub:
Changes or additions to the "fun" commands isn't something we really need, is it? 

In this case, it's making it potentially annoying by allowing you to ping another user via Smoke detector. As it stands, we already receive push back for pings on reports.

PR #295, opened 2016-11-11
Add Marlin Woods to username blacklist

bwDraco:
We're getting a flood of posts containing plagiarized content from unregistered accounts with the username "Marlin Woods" on Super User. Each post is on a separate unregistered account. See http://meta.superuser.com/questions/11908/flood-of-plagiarized-answers-from-marlin-woods.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -464,7 +464,7 @@ class FindSpam:
         #
         # Category: other
         # Blacklisted usernames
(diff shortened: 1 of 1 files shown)


Fortunate-MAN:
I don't think Smokey is for detecting plagiarized posts. AFAIK, this user is only plagiarizing content and not posting trolls or spam, which Smokey is intended to.

IMO, this shouldn't be merged as the user is **not** posting trolls or spam AFAIK.

Undo1:
Tend to agree that this wouldn't fall under Smokey's focus.

bwDraco:
Well, this is a bit of an edge case. It's getting referred to as trolling by others in the community, and the posts are disruptive because this is happening repeatedly and frequently, seemingly as to "play with" our moderation system by posting lots of inappropriate content that is hard to detect as such.

Undo1:
After reviewing this some more, I've come to the conclusion that this edge case is something we should detect. They're a persistent (ish) troll that is copying the question into the answer box, which is bad.

bwDraco:
For the benefit of readers, there was a typo in this code change which was fixed in [this pull request](https://github.com/Charcoal-SE/SmokeDetector/pull/296).

PR #297, opened 2016-11-12
Migrate Smokey user IDs to config

angussidney:
If I want to run Smokey under my own account (or a sock), I need to change the user IDs [here](https://github.com/Charcoal-SE/SmokeDetector/blob/master/globalvars.py#L248). However, if I do that, I can no longer pull from this repository via chat without fixing merge conflicts every time, as `globalvars.py` is version controlled.

There are three possible ways to deal with this:
 - Edit the upstream repository to use my user ID, however that would break the official bot (bad idea)
 - Continue to put up with the merge conflicts
 - Transfer user IDs to the config file, which is not version controlled. This is what I have done for this PR.

I have gone ahead and set up this change in a new branch, so it is ready to go if everyone agrees with me. But I will leave it to @ArtOfCode- to merge the PR, so that he can update the config files on his two smokey setups to make this work.

### DIFF
--- a/config.ci
+++ b/config.ci
@@ -8,6 +8,12 @@ location=Continuous Integration. If you're seeing this, create a config file wit
 # Set metasmoke_key if Smokey isn't running on the same machine (localhost)
 # that metasmoke is
 # metasmoke_key=gibberish_here
+ 
+# Set Smokey user IDs
+# Leave as default values for the normal smokey user
+charcoal_id=120914
+meta_tavern_id=266345
+socvr_id=3735529
 
 # Set GitHub keys
 # github_username=username@domain.com
--- a/config.sample
+++ b/config.sample
@@ -9,6 +9,12 @@ location=location_here
 # that metasmoke is
 # metasmoke_key=gibberish_here
 
+# Set Smokey user IDs
+# Leave as default values for the normal smokey user
+charcoal_id=120914
+meta_tavern_id=266345
+socvr_id=3735529
+
 # Set GitHub keys
 # github_username=username@domain.com
 # github_password=p@55w0rd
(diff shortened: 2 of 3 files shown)


hichris1234:
You should realize that as long as no one touches that line in globalvars, you shouldn't get a merge conflict. If you are getting a merge conflict, you're doing something wrong.

angussidney:
@hichris1234 wouldn't there be a merge conflict if *any* line in globalvars is edited? Or am I misunderstanding the way in which git works?

hichris1234:
You're misunderstanding how git works. As long as two commits don't touch the same line, it automatically merges your changes. But if two commits touch the same line, that's where you have merge conflicts.

angussidney:
A quick test on a test repo caused an error/merge conflict when I tried to pull from remote (both when the local changes were uncommited and commited). I was able to do it with uncommited code by using `git stash` and `git stash pop` after pulling.

However, since you would have more experience running a copy of SD I'm going to leave this closed until I run into any actual problems

hichris1234:
Oh, yeah, it has to be either committed or stashed. You can't pull from a remote with a dirty tree.

ferrybig:
I disagree that this is closed, commiting would break the !!/rev command and and the startup message, and would mean that the blacklist command pushes this commit to the online repo

Stashing wouldn't work because this requires modifications to the pull logic, and may cause inconsistent code results in case there is a merge conflict resolving it (the result aafter the last got pull is kept)

ArtOfCode-:
This can remain closed because it actually [isn't necessary at all](https://github.com/Charcoal-SE/SmokeDetector/blob/89aa9bdd0820f3505d8ed92ba51d90324fe34be0/ws.py#L49-L51) - Smokey sets the chat user IDs to the ID for the account it's running under when it logs in.

PR #299, opened 2016-11-16
Temporarily disabled the recovery message system

ardaozkal:
Something with websockets is broken and until that can be fixed, I propose disabling the error reporter because [it spams Charcoal HQ](http://chat.stackexchange.com/transcript/message/33551921#33551921).

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -202,7 +202,8 @@ while True:
             os._exit(4)
         ws = websocket.create_connection("ws://qa.sockets.stackexchange.com/")
         ws.send("155-questions-active")
-        # GlobalVars.charcoal_hq.send_message("Recovered from `" + exception_only + "`")
+        if exception_only != "WebSocketConnectionClosedException: Connection is already closed."
+            GlobalVars.charcoal_hq.send_message("Recovered from `" + exception_only + "`")
 
 now = datetime.utcnow()
 delta = UtcDate.startup_utc_date - now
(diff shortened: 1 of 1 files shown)


Undo1:
How about we just don't post those websocket errors?

AWegnerGitHub:
Instead of disabling it, why don't we report the issue to SE and see if
they can fix it?

hichris1234:
@Undo1 This is primarily the only error that is caught by that except block. While it's noisy right now, something is broken, so it's doing its job...

ArtOfCode-:
Not posting the errors is less than ideal because then we don't see errors... how about we post a *summary* of errors each hour - "X errors in the past hour" (or nothing if there aren't any)

dorukayhan:
Alternatively, we can make the error reporter not post a message if the error is `WebSocketConnectionClosedException: Connection is already closed`.

ardaozkal:
I think that we can just report this to SE and disable error reports in general or just for Socket Closed error in meantime.

ardaozkal:
Just sent a simple commit that checks if the issue is a websocket issue or not and reports only if it isn't. My python knowledge isn't the best so it might not work, tbh, but -yup, ci failed-

AWegnerGitHub:
I don't like hiding error messages. I don't think disabling them is the
right way to solve this.

A summary is a good alternative.

ardaozkal:
@AWegnerGitHub currently smokedetector is posting [a lot of error messages](http://chat.stackexchange.com/users/120914/smokedetector?tab=recent) and some people are simply moving them into the trash room. While a better option can be developed (like summary as you mentioned) or this bug can be fixed by SE, I think that disabling this exact error message won't do too much harm as it'll still show up in error logs. That said, I need to go now, have a nice day.

ArtOfCode-:
Socket errors are fixed now.

PR #300, opened 2016-11-17
bodyfetcher.py: give up if post doesn't have data

tripleee:
Recent ValueError failures from attempting to decode JSON which
wasn't there, apparently because SO is now killing off hung sockets
after a timeout.

### DIFF
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -65,7 +65,14 @@ class BodyFetcher:
 
     def add_to_queue(self, post, should_check_site=False):
         mse_sandbox_id = 3122
-        d = json.loads(json.loads(post)["data"])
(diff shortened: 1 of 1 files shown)


tripleee:
This is somewhat tentative, see comments in chat around http://chat.stackexchange.com/transcript/message/33557162#33557162

ferrybig:
At the moment, this is just a workaround around the actual problem. It still means we may miss posts because smokey is going to spam `
Recovered from WebSocketConnectionClosedException: Connection is already closed.` because the root cause of the problem is that we don't handle the `{'data': 'hb', 'action': 'hb'}` message correctly (it should respond back with a `hb` message

PR #308, opened 2016-11-23
According to Angussidney's suggestion

magisch:
To help track some more elusive one-off spams.


angussidney:
For reference: http://chat.stackexchange.com/transcript/message/33683992#33683992

PR #309, opened 2016-11-23
Blacklisted "molvi ji"

dorukayhan:
I just noticed that there are spam posts containing the string `molvi ji` as well as `baba ji` (case insensitive).

[Exhibits](https://metasmoke.erwaysoftware.com/reason/67)


hichris1234:
That would already be caught with just molvi: https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L13.

PR #324, opened 2016-11-30
findspam.py: alert for image from low-rep users

tripleee:
As requested by [Wrzlprmft](http://chat.stackexchange.com/users/68718) here:
http://chat.stackexchange.com/transcript/message/33819862#33819862


tripleee:
It should perhaps be emphasized that the code is set up to only trigger on {{academia.stackexchange.com}}

Issue #305, opened 2016-11-22
Add blacklist commands for keywords


angussidney:
Prompted by this: http://chat.stackexchange.com/transcript/message/33655049#33655049

I (and bwDraco) think that it would be great if we could implement a `!!/blacklist-keyword` command or similar.

Since we already have the `!!/blacklist` command implemented for websites, and the bad keyword regexes have already been separated into a text file, it should be fairly easy. All the infrastructure is there; we just need to copy/paste some code and change a few things.

While we're doing this, it might also be a good idea to rename the existing `!!/blacklist` command to `!!/blacklist-website`. I've noticed that a few people recently have misunderstood the use of the blacklist command, and changing its name should make that clear.

What are your thoughts?

If we reach an agreement that this would be a good idea, I could do this on the weekend (as long as no-one beats me to it).

bwDraco:
I second this.

To build on this, I'd also like to see a `!!/blacklist-username` command as well. A bit earlier (see #295), I had to manually fork the repo, edit the code, and submit a pull request in order to get a username added to the blacklist.

ArtOfCode-:
I third this.

With that said, we've had some discussion in CHQ on this topic. Consensus seems to be that we can move plaintext (or close to plaintext) lists to resource files, but regex patterns should stay in the code (such as pattern matching websites).

angussidney:
@ArtOfCode- Yeah, those were my thoughts. I originally was going to propose to do it for most blacklists, but then I realised that keywords and usernames were about the only other appropriate ones to move out into other files. 

And I agree with the idea that regexes should stay in code while plaintext can be moved into resource files.

hichris1234:
In 21 days, [5 commits, 2 of which involved no change \(ended up being reverted\)](https://github.com/Charcoal-SE/SmokeDetector/commits/master/bad_keywords.txt). With this level of activity...

ArtOfCode-:
With this level of activity, it's easy to forget what you're doing with the code, so adding a new robust command to do it for is a great idea! Just what I was thinking, @hichris1234.

hichris1234:
It's easy to forget how to add a line to a text file? Unfortunate times we live in...

ArtOfCode-:
Oh, definitely. I mean you've gotta remember which file it is, for a start, and our consistent naming schemes make that really difficult. And then you've gotta remember where the newline goes! And don't forget actually *typing*, who knows how to do that these days?

[...]

PR #325, opened 2016-11-30
Make findspam.FindSpam load the blacklisted username regex from a separate file instead of having it hardcoded

dorukayhan:
In [this reply](https://github.com/Charcoal-SE/SmokeDetector/issues/305#issuecomment-262158807) to #305, bwDraco suggested that we should implement `!!/blacklist-username` as well as `!!/blacklist-keyword` and `!!/blacklist-website`.

To do that, we have to do the following (assuming we want to keep the format of the username blacklist consistent with the website blacklist and bad keywords):

1. Export the username blacklist to its own text file
2. Make [findspam.FindSpam](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L201) use that text file instead of a hardcoded regex to figure out which usernames are blacklisted
3. Add the `!!/blacklist-username` command

Since 1 and 2 mess with parts that require no interaction with users, I decided to do them and leave the username blacklister to those who actually know Python.

### DIFF
--- a/blacklisted_usernames.txt
+++ b/blacklisted_usernames.txt
@@ -1,0 +1,12 @@
+^l(?:ol){2,}$
+^troll$
+tejveer ?iq
+ser?vice pemanas?
+\bnigg[aeu][rh]?
+\b(fuck(er|ing)?|penis)\b
+^w.ng.?d.ng$
+dlqudals
+^[a-z ]+juri(?:n|na|ns|sa|ya|yam|ym)$
+^[a-z]+jiibond$
+Marlin Woods
+xiguai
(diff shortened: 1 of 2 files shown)


dorukayhan:
Great. I refactor a thing and CI fails.

Undo1:
Maybe try `open("blacklisted_usernames.txt", "Ur"` (capital `U`).

dorukayhan:
Turns out CI was failing because of a whitespace-only line! *facepalm*

angussidney:
I was going to get to this today, but it looks like you beat me to it. Thanks, what you have done is the harder part, believe it or not.

As for the command, that is pretty simple. I wrote the blacklist function in `gitmanager.py` so that it accepts an argument for the type of blacklist, all that needs to be done is to add another `elif` [here](https://github.com/Charcoal-SE/SmokeDetector/blob/master/gitmanager.py#L20) and the new command below [here](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L276).

PR #336, opened 2016-12-06
allow "www" in the misleading link filter

NobodyNada:
The misleading link filter caught a post because that post linked to `http://example.com` but with the link text `http://www.example.com`.  I added a check for that in `misleading_link`.

Hopefully I can pass `flake8` this time...


ArtOfCode-:
Rule #1 of Smokey pull requests: you can never pass flake8.

PR #345, opened 2016-12-09
Blacklist oraclefusionfinancials\.in

tripleee:
Manually creating a PR while the !!/blacklist-website command is down.


ArtOfCode-:
I'm only seeing one true positive for this on metasmoke. Let's wait for it to turn up again before blacklisting.

tripleee:
I wanted to be preemptive because it's a morph of the UPPERCASE ONLINE TRAINING IN INDIA spammer.

PR #373, opened 2016-12-19
Add autoflagged? subcommand

angussidney:
This will allow users to reply to a report to determine whether or not it has been autoflagged.

Also, by the looks of it, I can no longer push to master, I have to make a pull request. I believe this will break the blacklisting commands.


Undo1:
Oh, that's because of the review requirement. Would like for that to be only on pull requests, might be possible to enforce through CI in a cleaner way. 

Disabling it for now.

PR #377, opened 2016-12-20
findspam.py: Nested blockquotes

tripleee:
Many of these spams have been detected only by the domain name heuristic, which could fail if they post with a domain name which doesn't contain terms on the blacklist.  To strengthen the filter, add a rule to identify the conspicuous use of nested blockquotes.

Example: https://metasmoke.erwaysoftware.com/post/50594


hichris1234:
This will apparently catch [~150 true positives](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title=&body_is_regex=1&body=%3Cblockquote%3E%5B%5B%3Aspace%3A%5D%5D*%3Cblockquote%3E%5B%5B%3Aspace%3A%5D%5D*%3Cblockquote%3E%5B%5B%3Aspace%3A%5D%5D*%3Cblockquote%3E&username=&why=&site=&feedback=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search). I could've sworn we already had something like this, but I guess not. I'll merge it in.

PR #397, opened 2016-12-27
Some minor code fixes

teward:
(1) While *extremely* unlikely, there is a tiny risk, if we change something about blacklist types and then forget to update `gitmanager.py`, that we'll have something blow up due to a no-value variable.  Let's have an "Else" case that will prevent that.

(2) Class methods tend to have `cls`, not `self`, in their arguments.

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -9,7 +9,7 @@ import json
 # noinspection PyRedundantParentheses,PyClassHasNoInit
 class GitManager:
     @classmethod
-    def add_to_blacklist(self, **kwargs):
+    def add_to_blacklist(cls, **kwargs):
         blacklist = kwargs.get("blacklist", "website")
         items_to_blacklist = kwargs.get("items_to_blacklist", [])
         username = kwargs.get("username", "")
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
(2) is stylistic, and I *think* (not 100% but fairly sure) that `self` is the more widely used.

ArtOfCode-:
^ TIL. Most of our stuff is instance methods, so should be self.

teward:
@ArtOfCode- @AWegnerGitHub If that's the case, I don't think they should be tagged as `@classmethod` then.  In either case, I've reverted that specific commit, and fixed up other parts.

teward:
Note that I've pushed a commit up for autopull that will include an IDE suppression on that; I run everything unfortunately through an IDE, and it was whining every time I committed, so I've just suppressed (like I did with other code changes).

AWegnerGitHub:
I'm on mobile so I can't check everything right now, but...we do have some
class methods. Those can use 'cls'. Everything not decorated as a class
method should use self.

PR #419, opened 2016-12-29
Upgrade 'testo' regex

teward:
Suggested by tripleee on chat: http://chat.stackexchange.com/transcript/message/34415527#34415527


tripleee:
Maybe also add `e?` before the `x`; this gets me many more hits.

teward:
@tripleee Push made, patience is a virtue it takes time to type on a tablet xD

PR #446, opened 2017-01-10
Thanks to @triplee, Matches more tps, no fps

magisch:
24 TP matches / 0 FP matches in the last 4 months.

Thanks to triplee


tripleee:
Maybe retitle to clarify that this updates the recent "lux derma" blacklisting regex.

PR #451, opened 2017-01-12
findspam.py: Update link inside nested blockquotes

tripleee:
Experimentation reveals six nested blockquotes or more are all TP.  Add the requirement to have a link alone inside the blockquotes.

This updates the rule introduced in PR#377 and tweaked in PR#416.

I have updated the rule description, I hope that's acceptable.  This no longer looks merely for nested blockquotes, so the old description would be misleading (but I'm wondering whether this will make Metasmoke regard this as a new heuristic).


tripleee:
For the record, yes, this created a new reason #113 which currently has a modest but accurate 16/16 TP score. https://metasmoke.erwaysoftware.com/reason/113

PR #452, opened 2017-01-12
finspam.py: bad_link_text: typo fix

tripleee:
If I can guess the intent here, there's an alternation operator | missing before the fragment "(^| )(porno? "... -- without that, there is no way there could be a match on ^


ArtOfCode-:
You've mixed your other PR into here as well, but no matter.

PR #456, opened 2017-01-16
findspam.py: Pattern for http://%20 in href

tripleee:
This is a common formatting error in spam currently (386 TP hits, 1 FP, 1 NAA)


ArtOfCode-:
Remember confirmation bias. Have you checked this against normal posts too?

tripleee:
I tried to search in SEDE but it times out as soon as I get a query which I hope will actually work.

On a smaller site like English, I can run this, and get zero results: https://data.stackexchange.com/english/query/616177/urls-with-http-20-pattern

tripleee:
Zero on Mathematica (too small to be useful, I guess), Ask Ubuntu, Server Fault, Super User, and Mathematics.  Still times out on Stack Overflow. (The SEDE front page arranges sites by size, so I picked the top four after SO, and a smaller one by mistake -- Mathematica autocompleted when I wanted to select Mathematics.)

AWegnerGitHub:
The onsite search shows 1,400+ posts with the pattern `http://%20`

[Search Link](http://stackoverflow.com/search?tab=newest&q=url%3a%22http%3a%2f%2f%2520%22)

PR #462, opened 2017-01-18
Should be an indexerror not a valueerror

AWegnerGitHub:
see http://chat.stackexchange.com/transcript/message/34852020#34852020 for reference

Wrong exception was being caught


teward:
I've seen ValueError types in historic error captures at this line.  It would be better to capture both types if we're going to change the capture.

PR #473, opened 2017-01-20
http://%20 antipattern fix for repeated URL rule

tripleee:
Spammers occasionally paste in a space before the URL and that currently bypasses this rather strict rule.  Edit it to tolerate this variation.


tripleee:
Hmm, it dislikes formatting I didn't change, again?

tripleee:
Thanks for the heads-up; updated.

PR #480, opened 2017-01-24
command_pending error references unset variable

tripleee:
The ValueError exception attempted to use the variable we just failed to assign.
Fix by displaying the raw content in the error message instead.


tripleee:
This code still contains a slightly pesky assumption that `content` is always subscriptable, didn't check whether this is guaranteed

PR #481, opened 2017-01-25
bad_keywords.txt: Add Nigerian phone number pattern

tripleee:
As discussed here: http://chat.stackexchange.com/transcript/message/34997711#34997711

234/235 TP (one NAA): https://metasmoke.erwaysoftware.com/search?body=%28%5E%7C%5B%5EA-Za-z0-9_%5D%29%5C%2B234%5B0-9%5D%7B8%2C%7D%28%24%7C%5B%5EA-Za-z0-9_%5D%29&body_is_regex=1&commit=Search&feedback=&feedback_filter=naa&reason=&site=&title=&user_rep_direction=%3E%3D&user_reputation=0&username=&utf8=%E2%9C%93&why=

The proposal is Glorfindel's but we agreed I would make the PR. This is somewhat more far-reaching than your routine blacklisting request so I think we want to have this reviewed by the authors.

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -240,3 +240,4 @@ raw\Wpower\Wxl
 pantothen
 argireline
 brick\Wmuscle
+\+234[0-9]{8,}
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -240,4 +240,4 @@ raw\Wpower\Wxl
 pantothen
 argireline
 brick\Wmuscle
-\+234[0-9]{8,}
+\+234\d{8,14}


tripleee:
Tangentially notice also the spurious second NAA result, what's up with that? 

http://chat.stackexchange.com/transcript/message/34997774#34997774

tripleee:
This SEDE query finds 21 potential FPs in all of Stack Overflow: https://data.stackexchange.com/stackoverflow/query/619660/nigerian-phones-take-2

ArtOfCode-:
Does this not get caught by the phone number in `{}` reason?

tripleee:
Nope; as noted in the very first comment in the chat discussion I linked, that only gets applied to the title.

tripleee:
I am aware of no particular reason to prefer `\d` over `[0-9]` but sure.  (It complicates matters if you want to use Metasmoke's somewhat less capable MySQL regex engine with these patterns.)

ArtOfCode-:
@tripleee I suggested `\d` because semantics. It's always good to be semantically correct where we

tripleee:
This was broken; fix in PR#484

PR #484, opened 2017-01-27
bad_keywords:txt: Fix plussed phone numbers

tripleee:
These were not working, because the leading plus is consumed by the word boundary condition in the code, so we can't match that directly, but we can say "if we just skipped a plus sign" with a lookbehind.

I included a couple of comments in the text file which are simply never going to match anything; but down the line, we might want to add code to skip comments in the file, or just remove these lines.

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -240,7 +240,9 @@ raw\Wpower\Wxl
 pantothen
 argireline
 brick\Wmuscle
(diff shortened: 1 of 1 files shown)


tripleee:
See also discussion about design guidance at http://chat.stackexchange.com/transcript/message/35038988#35038988

PR #493, opened 2017-02-01
Add special-case handler for Math.SE on 'has few characters'

teward:
Math.SE triggers a lot of false positives on this filter.

We should consider special-casing that site so that it has to have 5 or more and 15 or less unique characters for Math.SE.

This is free to be modified as necessary to adjust the threshold, or for a better approach/solution.


Undo1:
This seems to be a reasonable threshold for now, we can adjust as needed.

markyi370:
If we made this modular, we can apply it to other sites too.

PR #499, opened 2017-02-02
Metasmoke: Track 'last ping time', reboot if time > 60 seconds

teward:
**This is not thoroughly tested, and needs review before inclusion!**

We don't currently track the 'last time' a ping from MetaSmoke was processed.  Therefore, we have some odd cases where SmokeDetector will 'die off' and not ping.

By tracking the 'last pinged' time, and check if that time is greater than 60 seconds into the past.  If it is, then die off with exit code '10' and write to the ErrorLogs accordingly, then wait 5 seconds, and restart SmokeDetector with a brand new set of sockets.  The 5 second wait is for 'networking must settle' waits, just in case.

This should allow for 'not pinging' cases to 'try' and recover by a forcible restart to SmokeDetector, and will kill existing sockets and use brand new ones, in case of uncaught network interruptions.

### DIFF
--- a/metasmoke.py
+++ b/metasmoke.py
@@ -164,6 +164,8 @@ class Metasmoke:
             print "Metasmoke location not defined; not reporting"
             return
 
+        metasmoke_key = GlobalVars.metasmoke_key
+
         try:
             payload = {
                 'feedback': {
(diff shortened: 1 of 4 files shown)


metasmoke.py
```diff
@@ -57,6 +58,17 @@ def init_websocket():
                     time.sleep(10)
 
     @staticmethod
+    def check_last_pingtime():
+        while True:
```
Undo1: Instead of looping forever and ever, you might consider only running this check once every few seconds and making a new `threading.Timer`, like we do in `send_status_ping`.

PR #500, opened 2017-02-03
Run git.config safely

tripleee:
Ignore exit code 1 when checking whether global user.name and user.email
are set or not.  Also, fix copy/paste error in user.email check.


tripleee:
Monologue at http://chat.stackexchange.com/transcript/message/35189117#35189117 in reference to repeated crashes earlier in the transcript.

PR #510, opened 2017-02-06
Get rid of the anchors in wingding's username regex

dorukayhan:
The name of the "user" who posted https://metasmoke.erwaysoftware.com/post/56146 explains why I made this commit better than I could.


Undo1:
Feels like a pretty broad pattern without the anchors. Could we instead allow a few (2-3) characters on either side?

angussidney:
@Undo1 done, ready to merge if you're happy with it

Undo1:
Looks good to me, we'll watch for false positives.

PR #511, opened 2017-02-06
Blacklisting nazi(s) on SciFi as per moderator request: 

Glorfindel83:
Blacklisting nazi(s) on SciFi as per [moderator request](http://chat.stackexchange.com/transcript/message/35247701#35247701)


Undo1:
Skeptical about this one (partially due to http://scifi.stackexchange.com/search?q=nazi), but we'll let it run and watch the results. I might move it to its own reason.

Glorfindel83:
It works only on 1-rep users, that should significantly reduce the number of FPs.

PR #522, opened 2017-02-10
Infer missed IDs from previous API calls

Undo1:
Don't merge yet, but feedback is welcome.

See http://chat.stackexchange.com/transcript/message/35339676#35339676 for context. Here's the theory:

 1. When we make an API call, for example, for post IDs `[1,2,3,4]`, we store the *highest* ID (`4`)
 2. When *another* API call for that site is triggered (for `[7,8,9,10]`), we:
    1. take the highest ID from the new batch (`10`), 
    2. compute all IDs between the previous highest ID (`4`) and the new highest (`10`) = `[5,6,7,8,9,10`]
    3. Merge with the requested IDs (`[7..10]`) as they may contain an edited post (imagine if `0` was a valid ID and requested) and uniqueify

The idea is that if we make an API call for `[1,2,3]`, then one for `[5,6,7]`, we *probably* intended to get `4` too but the websocket didn't give it to us for whatever reason. 

Thoughts?

### DIFF
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -132,7 +132,7 @@ class BodyFetcher:
 
         if site in self.previous_max_ids and max(new_post_ids) > self.previous_max_ids[site]:
             previous_max_id = self.previous_max_ids[site]
-            intermediate_posts = range(previous_max_id, max(new_post_ids))
+            intermediate_posts = range(previous_max_id + 1, max(new_post_ids))
 
             # We don't want to go over the 100-post API cutoff, so take the last
             # (100-len(new_post_ids)) from intermediate_posts
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -12,6 +12,7 @@ import requests
 # noinspection PyClassHasNoInit,PyBroadException
 class BodyFetcher:
     queue = {}
+    previous_max_ids = {}
 
     special_cases = {
         "math.stackexchange.com": 15,
@@ -62,6 +63,7 @@ class BodyFetcher:
 
     api_data_lock = threading.Lock()
     queue_modify_lock = threading.Lock()
+    max_ids_modify_lock = threading.Lock()
 
     def add_to_queue(self, post, should_check_site=False):
         mse_sandbox_id = 3122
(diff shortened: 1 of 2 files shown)


angussidney:
This seems to be a great idea; but I have one concern: we've come close to running out of API quota a few times recently, will this have any impact on our API usage?

Undo1:
@angus It won't touch request count; it just tried to bundle data we missed into the free space in our existing requests. (we never have a full request, the highest site is 15 and most are 2)

AWegnerGitHub:
A little further down in the transcript, Undo provided some Super User IDs:

http://chat.stackexchange.com/transcript/message/35341044#35341044

I ran those through the API. There are 98 IDs in the initial list. 72 of those were returned via the API. [52 are answers and 20 are questions](http://chat.stackexchange.com/transcript/message/35341148#35341148)

The remaining 26 IDs are provided in the transcript: http://chat.stackexchange.com/transcript/message/35341180#35341180

---

My questions, before determining if this is good or bad:

 1.) Why are we missing a quarter of the initial list provided? What happened to these 26? I don't have the rep on SU to see the timelines for those posts. 
 2.) On SO, we already miss posts due to the web socket not showing all SO posts. If we aren't working around that via the API already, I think this is one potential spot for higher API calls. 
 3.) What's the impact going to be to processing time/speed of these "extra" posts?

---

One area of concern:

The sites with low queues are going to break this if/when the following occurs:

 - Current post ID is 5000
 - A user edits a post from last week on Drupal - post ID 4700 (roughly 43 post a day is 300 posts)
 - A new post is created, Smokey checks it as normal and sees the new ID is 5001. We now pull 300 extra posts causing at least three extra API hits plus time for smokey to chug through up to 300 posts (minus any that have been deleted).

This is going to bog down even more on busier sites or when edits occur on much older posts.

SmokeDetector:
It's capped at 100/req, we will never have extra API calls from this.
Priority goes to the requested IDs first, and we throw out the oldest
intermediate IDs as necessary to get to <= 100 IDs.

Undo1:
... That was me. 

Additionally, I don't believe this will make any difference at all on SO; it runs custom logic anyway (takes the last 50, all the time)

Undo1:
That does bring up a few things to test, though. I need to check what happens if the new IDs are all lower than our previous max id, and make sure we never allow a max ID to shrink. Both easy.

Undo1:
After running this for a couple hours on my Pi, it (1) doesn't seem to cause any undue additional load (except up front, and my Pi still handles it fine. It also doesn't create any more API requests.

Most of the additional IDs appear to be answers, but I did see a couple missed questions show up too (especially on the higher volume non-SO sites). 

Merging for now; we can always revert if needed. I'm still open to feedback on this.

PR #524, opened 2017-02-12
Enable ability to go into 'Attach Debugger' mode w/ PyCharm debuggers

teward:
Currently, I, @teward, am probably the only person that would actually be running Smokey within such a debugger instance, but this would enable others to do so as well if they wished to try and debug odd crash errors.

That said, I want to make sure we're OK with adding this.  Otherwise we can just torch the branch, and I'll make local-tree changes each time I need to debuggerize this thing.

------

The changes in this branch I'd like pulled in allows us to run PyCharm 'debuggers' against SmokeDetector, without the evils observed in https://github.com/Manishearth/ChatExchange/issues/135 by simply having the Python process that spawned with `ws.py` attach to a PyCharm debugger at a specified host/port combination before doing anything.  It will then, accordingly, suspend everything until the Debugger attaches, so we don't have errant `ws.py` threads running when we don't want it to.

This permits enabling the PyCharm debugger attachment functionality ***only*** by defining `PYDEVD` in the environment variables with a value of 'True', 'Yes', or 'Enable', and the ability to override the Host (default value is 'localhost') and Port (default value is '20500') values when we enable debugger attach capability so you can have it attach to an off-environment debugger which can still track what's going on.  Therefore, it will only ever be a concern *if* a specific SmokeDetector instance is explicitly told to use the debugger functions, either by ws.py directly or by nocrash.py.

For example, SmokeDetector on `teward/Aurora` would be able to reach the dedicated debugger instance that I would already have running/listening on my other system, Daedalus, here on my local network, as it is permitted past via firewall rulesets; any 'crash' on Aurora would be caught, lock the threads and all processes, and then 'freeze' so we can get an idea of what's got what values from where.  Since this would then timeout on pings and communication with MetaSmoke, the autofailover system would kick in to bring up a standby instance to 'active' should such 'debugging' instances freeze up due to the 'suspension' of activity caused by an uncaught error if we add debugging capture points locally or on the Git tree everywhere.

------

From the first commit's full message, this is a detail of the changes, and the long details of how this all works:

First, we introduce a new `Debugging` class in `debugging.py`; this handles determining whether or not we need to launch the Debugger, based on environment variables, and constructs different ways of getting the arguments in ways you can manipulate what's being launched later (a la `ws.py`)

Secondly, we modify `nocrash.py` a little.  We make it have a 'copy' of the current environment variables already made available to it, from launch.  We then go and check, via the import of the `Debugging` class, whether or not we're enabling attaching to the debugger.  If `Debugging` is enabled, then add values to the copied environment variables set, defining a secondary set of environment variables that we will ultimately pass down to `ws.py`.

Thirdly, we modify `ws.py`.  This is where the most important modifications would happen.  Given that `nocrash.py` spawns `ws.py`, when using a debugger such as that in PyCharm it is impossible to attach to `nocrash.py` when it's already running and also track activity in `ws.py`.  This makes it impossible to 'break' when you want it to (such as an unhandled exception in `ws.py`) to get an idea of what was happening, where, and what triggered a given error.  (It also makes it impossible to get a memory snapshot of what variables have what values, etc.)  In this case, since we'll be using the same debugging 'server', we have `ws.py` run its own checks on the environment passed down to it, via the `Debugging` class import, and determine early on in the code whether we're attaching to a debugger or not.  This check happens before we even begin to launch `ws.py`, and prints to the command line that it's attaching the debugger.  Failure to import or find the corresponding package or module will trigger a failure case and totally exit entirely via exit code 6; failure to attach to debugger fails and exits via the same exit code.

There's a downside here though.  We have to include a copy of PyCharm's `pycharm-debug.egg` package in the local tree.  This adds a little bit to the files, but is mandatory if we want to be able to include 'debugging' capabilities we can easily switch on and off.  So, fourth item of business is to make sure that `pycharm-debug.egg` is properly included in the repository, which means making a "do not ignore" entry in `.gitignore` for that file.  (This also adds a comment a few lines earlier about the .DS_Store file, but that's mostly irrelevant)


teward:
I won't be deleting the branch, because this way it's not hard to keep debugger capacity later if we absolutely need it.

PR #544, opened 2017-02-23
"Git not supported on Windows" Warnings (Fix #540)

### LINKED ISSUE #540: Smokey tests fail to run on Windows
Smokey Detector includes a Windows incompatible library: `sh`. This is clearly noticed when running tests locally. 

[`sh` will not support Windows](http://amoffat.github.io/sh/sections/faq.html#will-windows-be-supported)

[`pbs` will no longer be supported](https://pypi.python.org/pypi/pbs)

	ImportError while importing test module 'H:\Development\SmokeDetector\test\test_chatcommunicate.py'.
	Hint: make sure your test modules/packages have valid Python names.
	Traceback:
	test\test_chatcommunicate.py:3: in <module>
		from chatcommunicate import *
	chatcommunicate.py:10: in <module>
		import chatcommands
	chatcommands.py:13: in <module>
		from gitmanager import GitManager
	gitmanager.py:1: in <module>
		from sh import git
	h:\Anaconda\lib\site-packages\sh.py:36: in <module>
		support." % __version__)
	E   ImportError: sh 1.12.9 is currently only supported on linux and osx. please install pbs 0.110 (http://pypi.python.org/pypi/pbs) for windows support.

--- 

Is there an alternative we can utilize that will allow developers in both environments to work without the above error?

teward:
Rather than a Dockerfile approach to fixing #540, let's just check if we're on Windows or not, and if not just don't import the `sh` library and for any `git.` calls preempt the calls by checking the platform instead; if we're on Windows, it'll just fail out and never reach a `git` function, and also never reach the `from sh import git` functions either if Windows is the platform we're on.

(Inspired by an approach I made to my `lxc-bootstrap` script for an LXC container automated bootstrapping approach, where I tested if we're on Linux and if not return "NotImplementedError").


teward:
In testing this, we can properly run the test code.  We cannot, however, launch Smokey in a way that works with Windows.  Testing on Win7 and Win10, I got the same issue about not being able to get the `usr` cookie.  That suggests some Windows incompatibility, so I've added a "quit" condition early in `ws.py` that will exit if you try and actually run SmokeDetector in a Windows environment.  This does not, however, affect the test suites, so all is good.

teward:
Tested on four separate Windows instances, this prevents the `sh` import problems.  Note remaining issues such as that in #555.

Issue #488, opened 2017-01-29
Improvements to Smokey's reports in chat


Glorfindel83:
There is some room for improvement of Smokey's chat messages. For instance, the quite valuable MetaSmoke link is sometimes dropped because of the length of the message. To make optimal use of the 'real estate', we should consider that the messages are seen both in general chatrooms by users who have hardly heard of SmokeDetector, and by hardcore spam fighters in Charcoal HQ.

Some discussion about this can be found [here](http://chat.stackexchange.com/transcript/message/35077541#35077541).

IIRC, feedback via chat depends on correctly parsing the report message, so one can't just change the message format without changing that logic too. Also, I've seen some chat messages in the test part of the repository, which also need to be adjusted.

Here is a list of possible improvements. Some of them should be 'permanent' changes, others should only be changed if there's not enough space.

 - [x] The first link ([SmokeDetector]) isn't useful. In Charcoal HQ, you can find the link to GitHub in the starred Project Links message, or in the chatroom's description. It should be dropped there; for *other* chatrooms, it should probably link to http://charcoal-se.org/#whats-smokey because that's much more informative for a newcomer than the GitHub repository itself.
 - [ ] We can remove the spaces, so that the first part looks like [SmokeDetector|MS] *if necessary*.
 - [ ] Some of the reasons can be made shorter (at least, the way they're displayed in chat), e.g. "website" -> "site", "repeated URL at end of long post" -> "repeated URL", *if necessary*. [See here](http://chat.stackexchange.com/transcript/message/35077601#35077601) for more examples.
 - [ ] Reasons can be grouped, e.g. "bad URL, bad keyword (x2), blacklisted/pattern-matching website (x4), repeated URL" *if necessary*.
 - [x] List can be shortened (e.g. Reason 1, Reason 2, Reason 3, Reason 4, Reason 5, +2 more)
 - [x] The site, usually written out in full like `graphicdesign.stackexchange.com`, can be shortened to `graphicdesign.SE` *if necessary*.
 - [x] Add autoflagged information (if that's even possible - maybe something for the long term) to the messages (in Charcoal HQ only). This is also something a userscript could do.

Any thoughts?

DavidPostill:
Agree to all suggestions

ArtOfCode-:
[...]

PR #550, opened 2017-02-25
Change repo link to description in report prefix

angussidney:
As per the first point of #488, this replaces the link to the repository with a shortened URL to the 'What's Smokey' section of our website. This would be more useful to newcomers than the repo link, and you can still access the repo via the pinned projects in CHQ.

Note that the prefix for the startup message remains unchanged (still pointing to the repo), as it is more likely that people would like to be linked to the repo then.

https://goo.gl/eLDYqh -> http://charcoal-se.org/#whats-smokey (sadly, git.io only works for Github pages sites under the .github.io domain.)

If there aren't any concerns, I'll merge this tomorrow.


j-f1:
You can still use git.io — just use https://charcoal-se.github.io/#whats-smokey (https://git.io/vyTcn).

PR #552, opened 2017-02-25
Updated two methods which work on major spam targets

Glorfindel83:
Updated two methods which work on major spam targets - the list of sites was a little outdated.


Glorfindel83:
Not really, it's just that there hasn't been much spam on those sites in
the last months. Therefore, I chose to replace them with the current spam
magnets.
I suppose we could have MS keep track of a 'Top-N' list of spam targets,
and refer to those in findspam.py.

PR #556, opened 2017-02-26
Windows is not supported for runtime, so "exit" (fix #555)

### LINKED ISSUE #555: Smokey Fails to Run on Windows
Excluding the test failures issue in #540 (and the now-merged fix in #544 that will fix the failure in the 'tests' sections which was necessary for these tests and running into this issue), we have a reproducible runtime failure in Windows environments.  That is, we get errors such as these repeatedly:

```
Git support not available in Windows.
teward/Aurora
https://metasmoke.erwaysoftware.com
Git support not available in Windows.
LoginError: failed to get `usr` cookie from Stack Exchange OpenID
2017-02-26 18:45:08.642000 UTC
  File "ws.py", line 97, in <module>
    GlobalVars.wrap.login(username, password)

  File "C:\Users\Administrator\Documents\SmokeDetector\ChatExchange\chatexchange\client.py", line 138, in login
    self._br.login_se_openid(email, password)

  File "C:\Users\Administrator\Documents\SmokeDetector\ChatExchange\chatexchange\browser.py", line 154, in login_se_openid
    "failed to get `usr` cookie from Stack Exchange OpenID")
```

I have gotten this error four separate times - in Windows 10, in Windows 7, in Windows Server 2012, and in Windows Server 2016 (which was on EC2 just to make sure it wasn't my environments).  In all four cases, this error has occurred, preventing Smokey from running.

This suggests to me that SmokeDetector is incapable of running on a Windows environment as a result of something in ChatExchange.  I'd like to propose allowing us to run the "tests" on Windows, but actually *exit* if someone tries to actually execute SmokeDetector on a Windows environment, given that I've had these ChatExchange-originating errors in four separate Windows test environments.

teward:
These changes only apply when SmokeDetector is actually being run, not during CI tests.

Windows and ChatExchange don't get along (see #555 for details).  So, we can't use Windows.  Exit and post an error about this as well.

(NOTE: This will not affect Windows Subsystem for Linux, aka Bash on Windows, which shows a "Linux" platform line.)

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -12,6 +12,20 @@ install_thread_excepthook()
 # test it thoroughly.
 
 import os
+
+# Platform Check: If Windows, we can't run, see GitHub issue #555.
+# Do it here, so we can die off early.
+import platform
+if "windows" in str(platform.platform()).lower():
+    print "OSError: Windows is not supported at this time."
+    with open("errorLogs.txt", "a") as errlogs:
+        errlogs.write("ERROR: Windows is not supported at this time.")
+    errlogs.close()
+    # noinspection PyProtectedMember
+    os._exit(6)
+
+# Debugger Attachment Here - we only work with this if specified in
+# the environment variables that we want to run in debugging mode.
 from debugging import Debugging
 if Debugging.enabled:
     # noinspection PyBroadException


ArtOfCode-:
Do we *need* this? Certainly a while ago it was entirely possible for Smokey to die off and nobody with creds to be around to boot it up again, in which case someone without creds would just boot up a standby for an hour. If we're doing that, it really doesn't matter if we're running on Windows for an hour - no, we won't have git commands or probably websockets, but that's nowhere near essential in a standby. If someone can make it work with Windows for an hour, I'm happy to let them try instead of flat-out not supporting Windows.

teward:
@AWegnerGitHub ttested this in Windows and it's able to run.  Somehow.  Closing this PR, nuking the branch.

PR #560, opened 2017-02-27
Try and better handle logon failures and "Invalid Literal"

teward:
This issue has plagued us in the past.  Logon may randomly fail, and we get an "invalid literal" error when trying to parse a string as an int.  Our first attempt at this capture didn't fix the core issue.

This is a major refactor to the logon process, *but* it will give us a better way of handling things.

Firstly, split the individual logons and value acquisitions into separate little segments of code.

Secondly, put a `for` loop around each segment, and track how many times we've iterated over the logon attempts, waiting 1 second between each failed attempt.  This also limits us to 10 attempts by 'for' loops.  We then try to get the logon data and values that are valid, but when it fails it waits a second, then goes to the next item in the loop and tries again.  The moment we've got a valid logon and we don't get a value error, we can continue on.

Thirdly, if we've tried 10 times and still are failing, die off hard with a "RuntimeError" that will kill and reset the entire Smokey process (via the 'unhandled exception' hook).  There is one such "Die" case for each chat wrapper.  This is a much more painful approach to this, but will better allow us to handle chat death errors on 'invalid literal' and 'attempt' to logon again after waiting a short while.  There's probably better approaches to this, but this should work to evade those invalid literal errors.

This must be reviewed ***before*** being merged, for sanity.

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -98,30 +98,33 @@ load_files()
 filter_auto_ignored_posts()
 
 # chat.stackexchange.com logon/wrapper
-while True:
+for chatlogoncount in range(1,10):
     try:
         GlobalVars.wrap.login(username, password)
         GlobalVars.smokeDetector_user_id[GlobalVars.charcoal_room_id] = str(GlobalVars.wrap.get_me().id)
         break  # If we didn't error out horribly, we can be done with this loop
     except ValueError:
+        time.sleep(1)
         continue  # If we did error, we need to try this again.
 
 # chat.meta.stackexchange.com logon/wrapper
-while True:
+for metalogoncount in range(1,10):
     try:
         GlobalVars.wrapm.login(username, password)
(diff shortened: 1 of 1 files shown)


AWegnerGitHub:
Is this something we could pass upstream to CE?

teward:
@AWegnerGitHub I don't think so.  The error we run into, I've tried tracing the cause, but it goes so deep into ChatExchange that it's literally black magic that is holding everything together.  It's probably something @ProgramFOX or someone else will have to *really* dig into, and it's never a consistently repeatable error enough to narrow down exactly *where* it breaks.

j-f1:
@teward I *think* the problem is that the login has failed, but you’re not getting an error. If you `print()` the value of `user_name` [here](https://github.com/Manishearth/ChatExchange/blob/5e4feb2fbd63bb96bb4a6c7a8ece31fef8b45363/chatexchange/browser.py#L231), it will probably be “log in”.

teward:
@j-f1 There's an assertion error if you aren't logged in - i ran into it in testing, so I'm not very certain about that.

PR #564, opened 2017-03-01
Attempt to detect spammers who make their e-mail harder to detect

Glorfindel83:
Attempt to detect spammers who make their e-mail harder to ---read--- detect, e.g. `info @ charcoal-se . org`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ errorLogs.txt.*
 
 # Cache
 test/.cache/
+.cache/
 
 # Config
 config


j-f1:
A couple of suggestions:

- Maybe use `\s*` instead of `\s?` to detect multiple spaces
- Match `at` as well  as `@` (case insensitive)
- Match `dot` as well as `.` (case insensitive)
- Match `` `@` `` and other formatting methods

teward:
Would this capture this, which I just observed on a spam message?  `"antonio.iosdevmate" &Gmail`

Glorfindel83:
I have tried to implement @JF's suggestions, and came up with the following:

`(?<![=#\/])\b[A-z0-9_.%+-]+\W*(@|[aA][tT])\W*(?!(example|domain|site|foo|\dx)\W*(\.|[dD][oO][tT])\W*[A-z]{2,4})[A-z0-9_.%+-]+\W*(\.|[dD][oO][tT])\W*[A-z]{2,4}\b`

[Regex 101](https://regex101.com/r/5vzFow/1)

I'm hesitant to implement the example of Thomas; it looks too much like
valid code to me, for instance:

`if (name == "first.last" & gmailEnabled) {`

I also noticed the three regexes are slightly different. I think it makes
sense to have one global `EMAIL_REGEX` constant, which can be modified
before calling `regex.compile`, but I'd rather leave that to somebody who
*is* proficient in Python.

angussidney:
I'd like to see some data before we merge this. Is anyone experienced enough with the SEDE to get some?

Glorfindel83:
I think I am, but SEDE is too slow and doesn't support regexes, just the general SQL `LIKE` syntax.

dorukayhan:
> SEDE is too slow and doesn't support regexes

We can work around that by unconditionally `SELECT`ing as many `Posts` as possible, downloading the resulting dataset, and then running regexes over it.

ArtOfCode-:
That... would break SEDE.

dorukayhan:
@ArtOfCode I mean, we can `SELECT * FROM Posts`, download the results as a .csv file, and test the regexes on that .csv file. That wouldn't break SEDE as it makes no online interaction other than a mere `SELECT`ion.

SmokeDetector:
I've done it before, you can get any table 50k at a time from SEDE (or just
download the data dump).

artofcode:
@dorukayhan: that would be @ArtOfCode- , not me ;-)

ArtOfCode-:
o/ namesake :)

But yeah, @dorukayhan SEDE *really* doesn't like large numbers of records. You're likely to have trouble getting lots of posts from it.

PR #568, opened 2017-03-07
Fix leak of api_request_lock

NobodyNada:
`api_get_post` has error handling in case it is passed an invalid URL or the API does not return any `items`.  However, this error handling does not release `api_request_lock`, causing a deadlock the next time `api_get_post` is called.  A similar bug occurs in `bodyfetcher.py`.

This bug is *a* cause of `!!/report` breaking, but it may not be *the* cause.

### DIFF
--- a/apigetpost.py
+++ b/apigetpost.py
@@ -30,6 +30,7 @@ def api_get_post(post_url):
         time.sleep(GlobalVars.api_backoff_time - time.time() + 2)
     d = parsing.fetch_post_id_and_site_from_url(post_url)
     if d is None:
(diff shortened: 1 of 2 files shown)


Undo1:
If this was a cause, I'd expect to see our post scan rate on https://metasmoke.erwaysoftware.com/smoke_detector/8/statistics to drop before a !!/report is called. Need to look for that in the future, and retroactively.

NobodyNada:
It looks like there are drops in [Aurora's scan rate](https://metasmoke.erwaysoftware.com/smoke_detector/6/statistics) just like @Undo1 predicted:

stats

PR #594, opened 2017-03-17
add_to_blacklist(): explicitly push to master

tripleee:
This attempts to work around the "git push.default simple" symptom.

See also http://stackoverflow.com/questions/21772279/why-is-pushing-to-matching-the-default-in-git/21772695#21772695

Example transcript:
http://chat.stackexchange.com/transcript/message/36096835#36096835

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -93,7 +93,7 @@ class GitManager:
         if code_permissions:
             git.checkout("master")
             git.merge(branch)
-            git.push("master")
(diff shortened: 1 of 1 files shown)


angussidney:
It should work, according to the sh docs: http://amoffat.github.io/sh/sections/passing_arguments.html#passing-arguments

I'm going to merge, and if it breaks something, we're in no worse of a situation than we were beforehand

PR #598, opened 2017-03-17
Measure post queue timings

ArtOfCode-:
This should measure how long a post has been in the bodyfetcher queue at the point that the API call for its site is made. It stores the timings locally in `bodyfetcherQueueTimings.p` so that stats can be analysed. Also created a Python script to do that analysis.

PR'd so I can get a code review, and cc @Undo1 since he was talking about doing this.


teward:
Just a note that this broke some things, too, so we had to revert it in commit 0177abf, but it's now fixed.  (dict and lists aren't interchangeable, and we have to do some manipulation to make things work proper)

ArtOfCode-:
@teward I nuked the bodyfetcher queue manually, because yeah it broke things.

PR #599, opened 2017-03-18
Added whole_post and  'answer similar to existing answer on post' reason

quartata:
Rules can now have a new `whole_post` key that if true will cause the entire `Post` object to be passed to the method. Currently it's a little messy since due to how `test_post` is structured you have to include `title` `username` and `body` keys even though they're being ignored anyways (it tries to access them for generating the regex why without checking to see if the rule even is a regex...)

Additionally this also has the 'answer similar to existing answer on post' rule, which uses `difflib` like username is similar to website. This is to catch spammers that merely copy and paste another answer and add a bunch of spam links to it. I'll try to get tp/fp data as soon as I can.

### DIFF
--- a/classes/Post.py
+++ b/classes/Post.py
@@ -163,6 +163,10 @@ class Post:
         return int(self._owner_rep)
 
     @property
+    def parent(self):
+        return self._parent
+
+    @property
     def post_id(self):
         return unicode(self._post_id)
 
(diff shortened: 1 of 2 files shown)


quartata:
For anyone wondering, I ignored code golf because the posts are really short and often are very similar when they're in the same language.

PR #602, opened 2017-03-19
Added 'mostly dots' rule

quartata:
We've been getting a rash of spam posts that try to be "clever" like this: http://chat.stackexchange.com/transcript/message/36131864#36131864

We catch them anyways since our phone number detection isn't dumb, but I figured it wouldn't hurt to throw in an extra rule to catch them. This rule will trip if the ratio of dots to characters in the post is greater than 40%.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -407,6 +407,19 @@ def strip_urls_and_tags(string):
     return regex.sub("</?.+?>|\w+?://", "", regex.sub(URL_REGEX, "", string))
 
 
+# noinspection PyUnusedLocal,PyMissingTypeHints
+def mostly_dots(s, site, *args):
+    body = strip_urls_and_tags(s)
+    body_length = len(body)
+
+    dot_count = len(regex.findall("\.", body))
+
+    if body_length and dot_count / float(body_length) >= 0.4:
+        return True, u"Post contains {} dots out of {} characters".format(dot_count, body_length)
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -407,6 +407,19 @@ def strip_urls_and_tags(string):
...
 
 
+# noinspection PyUnusedLocal,PyMissingTypeHints
+def mostly_dots(s, site, *args):
+    body = strip_urls_and_tags(s)
+    body_length = len(body)
+
+    dot_count = len(regex.findall("\.", body))
+
+    if body_length and dot_count / float(body_length) >= 0.4:
```
ArtOfCode-: Why `float(body_length)`?

quartata: Because Python 2 does integer division when both are ints.

ArtOfCode-: /me is too used to Python 3.

teward: We can always *import* `__future__` but we'd have to redo a large portion of our math.  I only recently starting using the division aspect of `__future__` in one of my other scripts, but so far I'm a fan of it.  I'm not a fan of some other aspects of Python 3 though.  Yet.

ArtOfCode-: `from __future__ import braces`

teward: `from __future__ import division` would work too, but... eh.

PR #605, opened 2017-03-20
Allow joke !!/alives in CHQ

angussidney:
Currently, there are several joke responses to the !!/alive command which only show in SOCVR and the Tavern. I'd like to make CHQ great again by bringing them back to Charcoal HQ.

Does anyone have any objections? It's quite obvious by the fact that Smokey's posting a message that the bot must be alive, so there shouldn't be any concerns from that perspective.


ArtOfCode-:
Meh, sure. We don't really need new joke stuff, but I'm okay with existing stuff being used across the board

PR #606, opened 2017-03-20
Strip ANSI color codes from !!/gitstatus

quartata:


kevin-brown:
Normally you'd disable this with `git -c color.status=false status` but I have no idea how you'd do that with the Python library you're using.

quartata:
I'm not sure how either which is why I did this. I'll try and find out more about this module because using a regex is a bit clunky.

j-f1:
@quartata You can also use `git config color.ui false` to disable color everywhere.

Undo1:
Took @kevin-brown's suggestion. I'm not aware of any other places where we have issues with color, otherwise we'd want to consider setting it globally like @j-f1 says.

quartata:
Or @Undo1 can just do that. Didn't realize it was just a wrapper over the actual command...

Undo1:
Yep, it's just `sh`. You could do `from sh import foo`, and then `foo.blah("baz")` should be equal to `foo blah baz`.

PR #610, opened 2017-03-21
Added codecov support

quartata:
Be sure to add any other files we want coverage on: simply do `--cov=<module name>` on `pytest`


ArtOfCode-:
That's better coverage than I'd expected...

quartata:
@ArtOfCode- It's only tracking select files.

PR #612, opened 2017-03-22
Unblacklist couchsurfing\.com (8 FPs, 0 TPs)

Glorfindel83:
https://metasmoke.erwaysoftware.com/search?body=couchsurfing


Undo1:
Suppose there's a way to only show those Codecov reports if the coverage actually changed?

j-f1:
@Undo1 It seems there’s an option called `require_changes` that, when set to `true`, won’t comment unless there’s changes.

PR #619, opened 2017-03-26
Lowered threshold for similar_answer

quartata:
Previously I had used the same threshold for this rule as "username similar to website," which was *way* too high considering the difference in sizes between usernames and posts. After a little bit of experimentation, I've decided 70% is a good number to try.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -10,6 +10,7 @@ from urlparse import urlparse
 from helpers import all_matches_unique, log
 
 SIMILAR_THRESHOLD = 0.95
+SIMILAR_ANSWER_THRESHOLD = 0.7
 EXCEPTION_RE = r"^Domain (.*) didn't .*!$"
 RE_COMPILE = regex.compile(EXCEPTION_RE)
 COMMON_MALFORMED_PROTOCOLS = [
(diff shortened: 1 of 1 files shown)


quartata:
@Undo1 The codecov settings could use some tweaking.

findspam.py
```diff
@@ -397,7 +397,7 @@ def similar_answer(post):
             sanitized_answer = strip_urls_and_tags(other_answer.body)
             ratio = similar_ratio(sanitized_body, sanitized_answer)
 
-            if ratio >= SIMILAR_THRESHOLD:
+            if ratio >= 0.7:
```
ArtOfCode-: Why can we not just set `SIMILAR_THRESHOLD` to 0.7?

teward: +1 to @ArtOfCode-'s suggestion - we should not be hard-coding thresholds into conditionals if we have a variable we can use instead and then just alter.

quartata: @ArtOfCode- @teward Because `similar_username` uses that constant and it would be way too low for it.

ArtOfCode-: So... time for a new constant, then

quartata: Which I did.

PR #634, opened 2017-04-11
findspam.py: Add .us to popular scam TLDs

tripleee:
Generally, where we already had .in and .info, also capture .us


tripleee:
Triggered by a recent gmailcustomerservicenumber.us spam: https://metasmoke.erwaysoftware.com/post/64040

PR #635, opened 2017-04-12
findspam.py: Refactor net-worths.net regex

tripleee:
celebrity prefix is optional, add optional dashes to capture new variant.
https://metasmoke.erwaysoftware.com/post/64148


j-f1:
How about using `\W?` to catch other separators?

tripleee:
@j-f1 I don't think other separators are likely in a domain name.  Or do you have something specific in mind?

PR #645, opened 2017-04-18
Adding emoji alias for naa

Cerbrus:
Silly? Yes.
Practical? Actually, yea it is :D


Cerbrus:
Sadface... can't use that character in the python file...

PR #655, opened 2017-04-20
Add (?#review) comments to preemptive listings

tripleee:
For documentation, add a regex comment (which does not affect the matching behavior; it just gets quietly ignored) on recently added phrases which should be reviewed in a few days.

If we agree that this is a good idea, this should be codified and documented properly.

I tried to initiate a discussion at http://chat.stackexchange.com/transcript/message/36827286#36827286 but I have not seen any responses yet.  Also previously http://chat.stackexchange.com/transcript/message/36802878#36802878

> that was a solid 24/24 TP rate but I'm thinking we should simply blacklist on first sight when it's clear from context that it's yet another head on the Indian pharma spam hydra

I tagged the ones I added preemptively http://chat.stackexchange.com/search?q=preemptive&user=&room=11540 but this is obviously less than ideal.

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -363,6 +363,6 @@ neuro\Wxr
 test\Wshred
 phallyx
 nuvapelle
-jullen(?#review)
+jullen
 nos\Wex(?#review)
 platinum\Wbeaute
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -358,12 +358,12 @@ testx\Wcore
 lumidaire
 embova
 core\Wskincare
-leallure
+leallure(?#review)
 neuro\Wxr
 test\Wshred
 phallyx
 nuvapelle
 
-jullen
-nos\Wex
+jullen(?#review)
+nos\Wex(?#review)
 platinum\Wbeaute
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -358,7 +358,7 @@ testx\Wcore
 lumidaire
 embova
 core\Wskincare
-leallure(?#review)
+leallure
 neuro\Wxr
 test\Wshred
 phallyx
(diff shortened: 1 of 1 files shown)


angussidney:
The question is, do we really want to be pre-emptively listing keywords? If we're in a position to say that this keyword will most likely be in a lot of spam posts, then we can almost guarantee that these spam posts are going to be caught by a lot of *other* reasons/keywords too. 

I can see this changing from 'this looks like a product name, it's only the first time we've seen it, but it's similar to other product names so let's blacklist it' to 'there's only five words in this post that haven't been blacklisted yet, which one should I do next?'

---

If we do decide to continue with this, but add comments to the regex, who is going to come back and remove them? When do we determine that 'yep, this was the right choice'?

tripleee:
The basic observation here is that the Indian pharma spammer seems to be spinning up new domain names and product names almost daily.  When we see spam which is otherwise similar to previously seen spam, but a new product name or domain name brings it close to the autoflagging threshold, my instinct is to blacklist the newcomer so as to avoid misses in the near future.  The reason we have multiple heuristics in the first place is that none of them are assumed to be perfect, and so by reasoning of "defense in depth" we want to do what we can to avoid false negatives from a high-volume spammer.

Case in point, the term "jullen" which I preemptively blacklisted earlier today just got its second hit: http://chat.stackexchange.com/transcript/message/36828152#36828152

(... and so I will update the pull request to remove the `(?#review)` tag on that.)

tripleee:
The broader question of "how do we keep blacklists focused, clean, and correct" is an important one, but this merely attempts to offer a crutch for the case where you are unsure whether what you want to do is going to be a good idea in the long run.

I also want to have a separate policy discussion about whether preemptively blacklisting new terms used by the Indian pharma spammer is something we want to tolerate or perhaps even encourage, but agreement on the basic mechanism here is really independent of the outcome from that discussion.

tripleee:
Pretty much oboleted by the !!/watch-keyword command

PR #660, opened 2017-04-21
blacklist male\Wenhancement

tripleee:
Required some refactoring.  I can't believe this wasn't already blacklisted.

(There are about five false positives, Skeptics / Health / Meta; but the volume
of spam easily outweighs the tiny risk.)

Metasmoke TP rate 1190/1193 currently:
https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title=&body=fitwaypoint&username=&why=&site=&feedback=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -118,7 +118,7 @@ Herbalife
 Accumass
 purple rhino male enhancement
 Dating Coaching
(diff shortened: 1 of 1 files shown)


Undo1:
We'll want to watch this for false positives, but if it works then good catch!

PR #661, opened 2017-04-21
blacklist eye\Wnovela

tripleee:
Refactored skin\Wnovela to also cover this.


tripleee:
Just the one hit for now, so technically you could call this preemptive.

https://metasmoke.erwaysoftware.com/post/65449

PR #662, opened 2017-04-21
findspam.py: New reason "Bad pattern in URL"

tripleee:
For the time being, it specifically matches URLs ending with "-reviews"
(with a couple of oddball variations from old spam).  This is a common
MO for the Indian pharma spammer.  Most of these spams are detected just
fine, but they are mutating brand names etc. incessantly and might one
day succeed in getting past the usual filters.  This is added as just
one more layer of "defense in depth".

It should be relatively easy to see how to add more fragments to the
rule in the future.

There are two new test cases to check both branches of the regex at
the heart of this rule.

Current Metasmoke report for a similar query suggests 3494/3494 TP rate:
https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title=&body_is_regex=1&body=%28%5E%7C%5B%5EA-Za-z0-9_%5D%29%5BA-Za-z0-9_%5D%2B-reviews%28-%28canada%7C%28and%7Cor%29-scam%29%29%3F%2F%3F%3C%2Fa%3E&username=&why=&site=&feedback=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -277,7 +277,7 @@ def bad_link_text(s, site, *args):   # suspicious text of a hyperlink
 # noinspection PyUnusedLocal,PyMissingTypeHints
 def bad_pattern_in_url(s, site, *args):
     patterns = [
-        r'[^"]*-reviews(?:-(?:canada|(?:and|or)-scam))?/?',
+        r'[^"]*-reviews?(?:-(?:canada|(?:and|or)-scam))?/?',
     ]
     matches = regex.compile(
         r'<a href="(?P<frag>{0})"|<a href="[^"]*"(?:\s+"[^"]*")*>(?P<frag>{0})</a>'.format(
--- a/findspam.py
+++ b/findspam.py
@@ -275,19 +275,17 @@ def bad_link_text(s, site, *args):   # suspicious text of a hyperlink
 
 
 # noinspection PyUnusedLocal,PyMissingTypeHints
-def bad_fragment_in_link(s, site, *args):
-    fragments = [
+def bad_pattern_in_url(s, site, *args):
+    patterns = [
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
Metasmoke query is good. Have you run this against an unbiased dataset too?

tripleee:
Nope, I'm thinking let's just unleash this and check for false positives for the next few days or weeks.  Maybe somebody could try to google some FP candidates but it's past bedtime here for me so not me, not now.  I can try next week I suppose.

tripleee:
I found two FPs on Stack Overflow for this pattern.  https://data.stackexchange.com/stackoverflow/revision/662129/826523/posts-with-suspicious-reviews-pattern-in-url

My query is running into timeouts as usual, "top 2" got me [these](http://stackoverflow.com/questions/40727692/wp-customer-review-form-customization) [two](http://stackoverflow.com/questions/23319/should-you-bug-the-code-used-in-code-reviews) but I'll see if it can find any more. Somebody who actually possesses some SQL skills might be able to pull out more.

tripleee:
[Two](http://stackoverflow.com/questions/43476654/approve-changes-in-github) [more](http://stackoverflow.com/questions/32262295/adding-a-reviewer-to-a-github-pull-request/41028889#41028889). https://data.stackexchange.com/stackoverflow/revision/662129/826528/posts-with-suspicious-reviews-pattern-in-url

tripleee:
Thanks to @rene, a total of 29 FPs on Stack Overflow: https://api.stackexchange.com/docs/advanced-search#order=desc&sort=activity&url=*-reviews&filter=!9YdnSPuG8&site=stackoverflow&run=true

angussidney:
We can always revert this if anything goes wrong. 29 FPs all time doesn't seem too bad.

PR #666, opened 2017-04-24
Blacklist novellus alone

tripleee:
"Novellus Skin" was blacklisted but it occurs in spam with a lot of other words, like "Novellus Face Cream", "Novellus Naturals", etc.


tripleee:
Uh, sorry, I seem to have crashed into another branch somehow.  I'll submit a separate pull request with just this change.

PR #668, opened 2017-04-24
classes/Post.py: Don't pull in GlobalVars

tripleee:
For modularity, avoid doing anything with globals from this module.

Instead, possibly raise a ValueError for the caller to handle somehow.

### DIFF
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -332,7 +332,11 @@ class BodyFetcher:
                 continue
 
             post['site'] = site
-            post_ = Post(api_response=post)
+            try:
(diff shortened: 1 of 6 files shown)


teward:
While I am all for the modularity, I'm hesitant to just return an error - it's more than likely force a SmokeDetector restart in any invalid API reply.

tripleee:
The code attempts to trap the places in the code where the exception could be raised.

There are various exceptions which could be raised everywhere in the code, so if the client should not die on any of those, you need to put in a handler for that, which will then solve this perceived problem as well.  But I'm submitting a separate PR which avoids raising an explicit exception.

PR #669, opened 2017-04-24
Update bodyfetcher.py

AWegnerGitHub:
This adds an additional message when an `api_quota` missing message occurs. It will show the last `url` that was sent.


bodyfetcher.py
```diff
@@ -306,6 +306,7 @@ def make_api_call_for_site(self, site):
                 if GlobalVars.api_backoff_time < time.time() + 12:  # Add a backoff of 10 + 2 seconds as a default
                     GlobalVars.api_backoff_time = time.time() + 12
             message_hq += " Backing off on requests for the next 12 seconds."
+            message_hq += " Previous URL: `{}`".format(url)
```
teward: This can be added simply to the prior message, without its own line.

    message_hq += " Backing off on requests for the next 12 seconds. Previous URL: `{}`".format(url)

That's such an example.  (I don't think we need to add a dedicated `+=` if we can just use the existing one?

PR #670, opened 2017-04-25
Remove breast enlargement phrase from bad keywords

tripleee:
<del>Refactor "breasts? enlargement" to also work with e.g. a dash between the keywords,
as commonly seen e.g. in URLs.</del>


tripleee:
Looks like https://metasmoke.erwaysoftware.com/post/65957 is already covered, actually.  Maybe just take out this pattern.

tripleee:
You'll want to rebase to remove the silly first commit.  Thanks.

PR #675, opened 2017-04-26
README.md: update pip3 instructions

tripleee:
Either run both with --user or with sudo; explain the difference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,12 @@ git clone https://github.com/Charcoal-SE/SmokeDetector.git
 cd SmokeDetector
 git submodule init
 git submodule update
-sudo pip3 install -r requirements.txt --upgrade
+pip3 install -r --user requirements.txt --upgrade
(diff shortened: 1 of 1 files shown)


angussidney:
The commands are like they are for a reason: the package in `user-requirements.txt` has to be installed for only the current user so that it can update itself automatically without throwing errors, while the others should be installed system wide.

You can run `install_dependencies.sh` to do both in one hit

tripleee:
"Should be installed" system-wide why?  If anything, these should all be installed in a virtualenv so that you don't need `sudo` access and can update everything at your leisure.  I'm thinking the instructions should be easy to follow for a newcomer but explain variations for those who might want more control and understand what they are doing.

ArtOfCode-:
As angus said - we have `requirements.txt` for packages that should be system-level installed for whatever reason, and `user-requirements.txt` for packages (mainly `tld`) that need to be user-level installled.

PR #678, opened 2017-04-27
don't depend on GlobalVars

tripleee:
Instead, depend on helpers.log() for reporting parsing problems.

This makes the error less visible, but avoids the potential risks of using an exception (see discussion on [PR#668](/Charcoal-SE/SmokeDetector/pull/668), which this supersedes).

Also coincidentally fix the `api_response=` handling so that a `None` value will not override a default (notably, passing Metasmoke results with `"score": null` would fail further down in the code when you attempt to manipulate them).

### DIFF
--- a/classes/Post.py
+++ b/classes/Post.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 import json
-from globalvars import GlobalVars
+from helpers import log
 import html
 from typing import AnyStr, Union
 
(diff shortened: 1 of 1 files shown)


tripleee:
Since you *also* merged #668, the `log` call should be replaced with `raise PostParseError`, don't you think?

ArtOfCode-:
I inclined on not doing that, because raising the PostParseError obscures the details of the ValueError, which is the *real* error. And also, as Thomas said, logs are less likely to crash Smokey than exceptions.

tripleee:
Thanks, all good then.

PR #690, opened 2017-05-03
Extend richestcelebrities pattern

tripleee:
We have seen .bio and .wiki as well as .org -- simply blacklist anything with this base.


angussidney:
I'm transferring this to pattern matching websites, since it's now more of a pattern rather than a specific blacklisted website.

PR #693, opened 2017-05-03
findspam.py: what_is_this_pharma_title()

tripleee:
Idiosyncratic title formatting, starting with lowercase "what is this".

Metasmoke results are approximate because it doesn't support case-significant searching. 
 I get 101/114 TP rate with this, but the actual result is closer to 90/90 (however with some low-hanging fruit being missed by the rule): https://metasmoke.erwaysoftware.com/search?body=&body_is_regex=1&commit=Search&feedback=&reason=&site=&title=%5Ewhat+is+this+%28%5BA-Z%5D%7Chttp%3A%2F%2F%29&title_is_regex=1&user_rep_direction=%3E%3D&user_reputation=0&username=&utf8=%E2%9C%93&why=

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -211,11 +211,9 @@ def pattern_product_name(s, site, *args):
 
 
 # noinspection PyUnusedLocal,PyMissingTypeHints
-def bad_pattern_in_title(s, site, *args):   # title "what is this Xxxx?"
-    matches = regex.compile(r'^(what is this (?:[A-Z]|http://)').findall(s):
-    if matches:
-        return True, u'Bad pattern in title {}'.format(
-            ", ".join(["".join(match) for match in matches]))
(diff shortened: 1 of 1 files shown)


angussidney:
I think a better reason name for this would be 'Pattern-matching title'. More general reason names should be used so that we can match different patterns down the line under the same reason name.

tripleee:
Like Metasmoke, Google and Stack Exchange do not allow me to search for this specific capitalization pattern.  I have googled for `inurl:/what-is-this-*` and found only matches with proper case capitalization (but I guess any non-spam post would quickly get reformatted by an astute editor at least on the high-traffic sites).

tripleee:
I'm not a fan of this premature refactoring, and there are other rules which should be refactored as well if we want to be consistent; but here goes.

Discussion at http://chat.stackexchange.com/transcript/message/37131127#37131127

PR #695, opened 2017-05-04
gitmanager.py: add a critical section with a lock

### LINKED ISSUE #679: Blacklisting twice quickly breaks Smoke Detector
I guess the automation tries to work in the same directory where the previous merge is in progress.

Checking out into a temporary directory and removing that when the merge is done should work around this issue.

There have been multiple situations where Smokey needed operator intervention to get back on track; most recently, http://chat.stackexchange.com/transcript/message/37019763#37019763

tripleee:
This way, only one blacklisting request can be in progress at any time.  Further requests will simply block on attempting to acquire the lock, and thus wait nicely in queue.

Fixes #679

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -7,6 +7,7 @@ import requests
 import time
 import json
 import regex
+from threading import Lock
 if 'windows' in str(platform.platform()).lower():
     # noinspection PyPep8Naming
     from classes import Git as git
@@ -16,8 +17,11 @@ else:
 
 # noinspection PyRedundantParentheses,PyClassHasNoInit,PyBroadException
 class GitManager:
-    @staticmethod
-    def add_to_blacklist(**kwargs):
+
+    gitmanager_lock = Lock()
+
+    @classmethod
+    def add_to_blacklist(cls, **kwargs):
         blacklist = kwargs.get("blacklist", "")
         item_to_blacklist = kwargs.get("item_to_blacklist", "")
         username = kwargs.get("username", "")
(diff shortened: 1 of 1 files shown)


angussidney:
Code looks good to me; have you tested it? If not, I'll give it a try when I get the chance.

tripleee:
Once the `--global` glitch is fixed it'll be easier to test. Then, you could even add a unit test.

I'm thinking I should still fix the one remaining race condition; expect another push here soon (but I have meetings for several hours no, so "soon" is later than I'd like).

tripleee:
... Oh, it's merged now. https://github.com/Charcoal-SE/SmokeDetector/pull/694

teward:
What 'race conditions' remain?  If you refer to the 'deploy' branch, that must be returned to before we can do anything else - no way around that - that is, during the 'lock' we need to be able to return to the 'deploy' branch as normal; all merges should be based on that after all.

So, please explain 'race conditions' in context, I don't see your 'race condition' here unfortunately.

tripleee:
Returning to the `deploy` branch currently happens outside the locked critical section, which means it could (attempt to) overwrite the work directory of another instance which is currently inside the critical section.

tripleee:
Also for the record, I described the race condition in more detail in the related issue https://github.com/Charcoal-SE/SmokeDetector/issues/679#issuecomment-299099042

tripleee:
The design of `gitmanager.py` is such that it is pretty much impossible to test it without complex, elaborate, and error-prone arrangements to have the `master` and `deploy` branches refer to something else than the real thing.  I would hope this could be merged to Zephyr or something for testing if you are unwilling to merge it without some sort of system testing.

angussidney:
I'm happy for this to be merged, and if Something Bad Happens™ we can always revert it and try again.

teward:
I'd be happy to merge this and revert it, but firstly, let's merge in my Windows git-compatibility layer, though it should have no impact on this, i'd like to get that in and do testing on it and improvements on it, so that that isn't impacted by this directly.  It shouldn't affect any *actual* Smokey instances, but testing is hard because it reverts to the master and deploy branches then dies on Windows (part of my 'Make it Windows Compatible' evil heh)

tripleee:
A quick test run of this functionality is demonstrated here:

http://chat.stackexchange.com/transcript/message/37345341#37345341

I challenge you to rapidly blacklist two entries with the current code base to see if it works.

tripleee:
Something was funky with the previous code.  I have force-pushed a single commit which just adds the lock and wraps the critical region in a `try: acquire ... finally: release`

Issue #490, opened 2017-01-31
Whitelist for websites


tripleee:
There should be a way to exempt some web sites from the broad swath of regex which triggers on some legitimate sites but still catches a lot of spam.

For example, in the `tcl` tag on Stack Overflow, many questions and answers link to `tcl.tk` which I surmise is the official documentation for the TK toolkit.  These are obviously legit, and should be exempt from the [regex filter which triggers on everything in the `tk`, `pk`, and `ir` country top-level domains.](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L504)

Undo1:
Add them to the 'sites' array.

tripleee:
Undo my understanding is that the `sites` member of the rule dict entry lists SE sites where the rule should be exempted, not URLs which should not be matched by the regex ... Are you referring to a different variable, or do I misunderstand how it works?  I don't want to disable the rule on SO, I want links to `tcl.tk` to be exempt on all sites from all rules (there are multiple rules for sites in the `tk` ccTLD).

Undo1:
Yeah, I misunderstood.

angussidney:
Basic spec: search all existing rules for whitelist variants (and add it to rules which don't have them already), then refactor to refer to a global file containing the whitelist and [string format it into the check](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L696). Bonus points for adding a whitelist command.

PR #700, opened 2017-05-04
Add jsfiddle.net and codepen.io

CalvT:
I've gone with the full domain names to make sure it only is valid for the proper sites. This is adding them to the whitelist. codepen.io (0 tp / 2 naa / 19 fp), jsfiddle.net (5 tp / 4 naa / 25 fp)


j-f1:
@CalvT 
> ./findspam.py:87:121: E501 line too long (139 > 120 characters)
([build line 216](https://travis-ci.org/Charcoal-SE/SmokeDetector/builds/228669597#L216))

angussidney:
I'm going to merge this for now, but I think in future it might be a good idea to start work on a global whitelist (#490) which doesn't only work for LAEOA

PR #702, opened 2017-05-05
Implement a Windows wrapper class for "Git"

teward:
This should permit any of the 'git' functions we already use to work mostly out of the box, with one or two exceptions where we need to do a platform check in the code base, to work.  This makes Smokey Windows-compatible with all functions.


ArtOfCode-:
Looks okay to me, but @AWegnerGitHub might be a better person to review this

teward:
Yeah we are going to do further checks today or this weekend with full Smokey creds.  We know this works fine for got status though - we need to test actual blacklisting now, though.

AWegnerGitHub:
I stuck myself on this as a reviewer. As mentioned above, we're planning on testing in the next few days.

teward:
I'm merging this now, and will fix it later if things break.  Testing on the git branch it's on is causing some headaches, and code reverts which blow things up when testing blacklisting functions, etc. so it'd be better to have this and then remove it later if it doesn't work, similar to what we've done for other 'radical changes' we've implemented.

PR #706, opened 2017-05-05
test/test_blacklist_integrity.py: Simple checks for blacklist files

tripleee:
Fail on empty line, no linefeed, or DOS CRLFs in any of the files.


tripleee:
Er, it fails because it works!

PR #708, opened 2017-05-05
Give Cai privs in Charcoal

CalvT:
This is preemptive, they've been requested but not approved yet


CalvT:
Done by angussidney separately

PR #710, opened 2017-05-05
Test blacklist integrity duplicates

tripleee:
Run a check against duplicates entries as well while we are already scanning the blacklists.

Incidentally, also refactor to put the function inside a `test_blacklist_integrity()` function so it works like the other `test/*.py` scripts.

### DIFF
--- a/test/test_blacklist_integrity.py
+++ b/test/test_blacklist_integrity.py
@@ -2,12 +2,19 @@
 
 from glob import glob
 
-for bl_file in glob('bad_*.txt') + glob('blacklisted_*.txt'):
-    with open(bl_file, 'r') as lines:
-        for lineno, line in enumerate(lines, 1):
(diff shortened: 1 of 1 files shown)


blacklisted_usernames.txt
```diff
@@ -22,3 +22,4 @@ fujihdmanufacturer
 Skybridge Infotech
 SenPerfect
 laserscheap
+first
```
j-f1: Are you sure you want to do this?

tripleee: Uh, no, thanks for catching that!

tripleee:
Sorry, the Git test commits had bleeded in here -- will supersede with another PR.

tripleee:
Force pushed the correct history, should be good to go now.  Sorry for the mess.

PR #746, opened 2017-05-12
Add a !!/watch-keyword command

tripleee:
Matches on this keyword list will trigger a rule "potentially bad keyword" which is separate from "bad keyword" and is intended to have a lower barrier than outright blacklisting.

Basically, if you want to be alerted when "something" comes up, Smokey can now do that.

Items on the watched keyword list are annotated with a date stamp to allow an administrator to prune the list periodically (down the line we might want to add code to do this automatically, too).

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -267,10 +267,11 @@ def command_blacklist_help(*args, **kwargs):
     Returns a string which explains the usage of the new blacklist commands.
     :return: A string
     """
-    return Response(command_status=True, message="The !!/blacklist command has been deprecated. "
-                                                 "Please use !!/blacklist-website, !!/blacklist-username "
-                                                 "or !!/blacklist-keyword. Remember to escape dots "
-                                                 "in URLs using \\.")
+    return Response(command_status=True,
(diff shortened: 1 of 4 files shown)


tripleee:
Those CI failures seem rather draconian, is there any way to tweak them down from "failure" to "warning" or something?

ArtOfCode-:
No, but we can always just override and merge anyway :)

angussidney:
What's the difference between this and blacklisted keywords? Both will trigger a report on their own, so it's not like this one has a lower potential impact than blacklisted keywords.

tripleee:
The primary intent is to allow us to `!!/watch-keyword` when we don't have sufficient evidence to directly blacklist something.  I keep on making mental notes, bookmarks, notes to self etc on things which I am reasonably convinced are going to crop up again soon, but it's a hassle.  Being able to say "please remind me to go ahead and blacklist" is the primary feature here; we should definitely take care to ensure that we don't have things both in a watchlist and blacklisted at the same time (probably programmatically, eventually).

I'd be fine with artificially giving this a zero weight if that seems to be necessary and/or helpful.

PR #753, opened 2017-05-16
ws.py: don't visit other rooms with charcoal-hq-only

tripleee:
The functionality in chatexchange_extension.py is partial; need
to also prevent ws.py from setting up connections to various rooms
to really make it use Charcoal only.


tripleee:
The reason I am proposing this is that my sock puppet doesn't have chat privileges on SO so it crashes when it tries to join SOCVR, even when running with the `charcoal-hq-only` setting.

PR #754, opened 2017-05-16
chatcommunicate.py: Ignore anything > 3 words

tripleee:
Avoid triggering on discussions where people link to SD and say "why"
or "true" as part of a longer sentence


tripleee:
Not a viable approach after all -- the `test-a` command and friends needs to operate on messages of arbitrary length.

PR #757, opened 2017-05-17
Indian pharma spammers

tripleee:
This adds a number of recurring spam domains to the blacklisted web site list.

Each domain is a separate commit, with brief statistics in the commit message.

Every one has at least one Metasmoke hit with a weight below our default autoflagging threshold.

### DIFF
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -1153,3 +1153,4 @@ megacleanseradvice\.com
 worldmuscleking\.com
 gomusclebuilding\.com
 musclegainfast\.com
+healthygcpro\.com
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -1148,3 +1148,4 @@ researchomatic\.com
 nutritionfit\.org
 oralhealthplus\.com
 supplementlab\.org
+supplementstest\.org
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -1146,3 +1146,4 @@ freecustomwritings\.com
 cyberservices\.com
 researchomatic\.com
 nutritionfit\.org
+oralhealthplus\.com
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -1151,3 +1151,4 @@ supplementlab\.org
 supplementstest\.org
 megacleanseradvice\.com
 worldmuscleking\.com
+gomusclebuilding\.com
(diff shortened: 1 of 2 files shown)


tripleee:
A few do not have recent hits below autoflagging threshold.  I'm hoping this would stimulate discussion to agree on another refinement to the policy for when we can blacklist something (6 months? many recent hits even above the threshold?)

blacklisted_websites.txt
```diff
@@ -1071,3 +1071,22 @@ healthflyup\.com
...
+oralhealthplus\.com
+drhelpnutrition\.org
+supplementlab\.org
+supplementstest\.org
+megacleanseradvice\.com
+worldmuscleking\.com
+quicksupplementfact\.com
+maleenhancementshop\.info
+gomusclebuilding\.com
+healthyorder\.org
```
angussidney: The majority of these are already pattern-matched, and some are already blacklisted (like this one). This will need the be reviewed and pruned down before merging.

tripleee: Pattern-matched I think is a weaker heuristic than outright blacklisted.  The ones which are already blacklisted should be pruned, I agree.  Not sure if I can fix this today.  Thanks for the feedback!

angussidney:
I'd really like to have a discussion over our blacklisting standards soon, since [our standards](https://charcoal-se.org/smokey/Guidance-for-Blacklisting-Websites) which we proposed a while ago seem to have slipped a fair bit since we started autoflagging. I'd like to have this discussion before we merge this PR.

I'll write up a separate issue when I get the chance.

tripleee:
I removed `healthyorder.com` which was already captured by a broader pattern.  `maleenhancementshop.info` was also in watched, so is removed from there.  It still triggers "bad keyword in body" and all of these trigger "pattern-matching website".

My thinking with explicitly blacklisting these is that it communicates to ourselves that we know exactly who these guys are, and that we want every post with a link to their web site to trigger.  The overlap with pattern-matching website should be resolved the other way around if we don't want both to trigger, IMHO. Part of my reasoning is that you can't exactly predict what the patterns are going to match, and so when developing or refactoring the patterns for the pattern-matching rule, it's easy to accidentally lose one of these domains. (By the same token, perhaps the broad regex in the blacklist which already matched `healthyorder` should be refactored to a plain list of matching domains, or moved to the pattern-matching rules?)

Having an explicit enumeration of bad domains is helpful for further development, too, though you could argue that this should be a separate resource, perhaps then used as one of several used to build a run-time blacklist.

tripleee:
Pending on https://github.com/Charcoal-SE/SmokeDetector/issues/771

tripleee:
Updated, many of these domains had been watched or blacklisted in the interim so I started over and cherry-picked the commits I wanted from my original branch.  Force-pushed and ready to merge.

PR #769, opened 2017-05-19
Fixup commit message, PR title, etc

tripleee:
The watch_keyword handling was simply reusing the blacklist messages.
Refactored for clarity to indicate what was done and fix e.g. the generated Metasmoke query for the PR for new watches.

Not entirely happy with the wording in the generated messages, but this should fix the biggest issues.


ArtOfCode-:
Are there issues here I'm not aware of or can it be merged?

tripleee:
I was unsure of the protocol but I interpret the above as "please go ahead and merge".

PR #779, opened 2017-05-23
Watched is experimental

tripleee:
Mark all the "potentially bad keyword in {answer,body,title,username}" rules as experimental. Refactor some of the related code in `spamhandling.py` slightly.

### DIFF
--- a/spamhandling.py
+++ b/spamhandling.py
@@ -138,7 +138,7 @@ def handle_spam(post, reasons, why):
         log('debug', GlobalVars.parser.unescape(s).encode('ascii', errors='replace'))
         if time.time() >= GlobalVars.blockedTime["all"]:
             datahandling.append_to_latest_questions(post.post_site, post.post_id, post.title)
-            if reason not in GlobalVars.experimental_reasons:
+            if set(reasons).intersection(GlobalVars.experimental_reasons) != set(reasons):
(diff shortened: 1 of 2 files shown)


tripleee:
I suppose `experimental_reasons` could be a `set` instead of a `list` for simplicity (I don't think it's used anywhere else?) but I'm hoping someone could weigh in on whether my set-related fix makes sense in the first place.

spamhandling.py
```diff
@@ -84,7 +84,7 @@ def handle_spam(post, reasons, why):
     reason = ", ".join(reasons[:5])
     if len(reasons) > 5:
         reason += ", +{} more".format(len(reasons) - 5)
-    reason = reason[:1].upper() + reason[1:]  # reason is capitalised, unlike the entries of reasons list
+    reason = reason.title()  # reason is capitalised, unlike the entries of reasons list
```
j-f1: This [isn’t the same thing](https://devdocs.io/python~3.6/library/stdtypes#str.title). It’ll return “Bad Keyword In Answer, Another Reason, Why Not” instead of “Bad keyword in answer, another reason, why not.”

tripleee: Duh.  Thanks for catching this!  Rebased and force pushed without this silly commit.

PR #821, opened 2017-06-07
Adds new character_utilization_ratio function to check if a single

AWegnerGitHub:
Adds new character_utilization_ratio function to check if a single character is over used in a post.

Initially only active on judaism.stackexchange.com


AWegnerGitHub:
If we aren't using the codeclimate checks, can we remove it from CI?

Undo1:
!!/aprove

Undo1:
!!/approve

Undo1:
Ignoring the CodeClimate thing for now. Need to figure out what we're doing with it.

PR #822, opened 2017-06-08
Link to chat message in add_to_blacklist; link to commit in chat reply

### LINKED ISSUE #807: FR: Add more links to the blacklisting process
It seems like the reply message should say something like this:

> Blacklisted pottmaxxx: 1aec925

Also, I think the commit message should include a link to the message that ordered the blacklist so people can see any context.

j-f1:
Fixes #807.


ArtOfCode-:
!!/approve

PR #853, opened 2017-06-17
"lose weight" was causing too many false positives on Pets.SE

Glorfindel83:
"lose weight" was causing too many false positives on Pets.SE, including an autoflagged one today.
https://metasmoke.erwaysoftware.com/search?body=lose+weight&site=212

As per [moderator request](https://chat.stackexchange.com/transcript/message/38185045#38185045):

> I think we should remove the word weight gain and/or loss from the pets.se


tripleee:
!!/approve

PR #864, opened 2017-06-21
remove \w{9}rev\.com --autopull

tripleee:
This looks like an accidental commit.  The commit message when this was introduced indicates that a simple addition to a previous pattern was intended to be added, and nothing more.

https://github.com/Charcoal-SE/SmokeDetector/commit/fae73d8c3


tripleee:
Reminder to self:

    git blame -w d36f72272^ -- findspam.py | grep 'rev\\\.'

and recursively repeat until you find the real commit.  https://stackoverflow.com/questions/5098256/git-blame-prior-commits was helpful for discovering this, but you can't use the line number directly, because it of course changed multiple times over history.

tripleee:
@normalhuman can you please review?

ArtOfCode-:
!!/approve

PR #865, opened 2017-06-21
Fix formatting of watched_keywords.txt lines 205 and 206

NobodyNada:
A couple of lines in `watched_keywords.txt` had broken formatting; line 205 had a trailing space (I'm not sure if it's a big deal or not, but I removed it), and line 206 had two spaces instead of a tab (which caused an error message when loading the keywords).

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -202,8 +202,8 @@
 1496921326	Glorfindel	alphonsacollege\.com
 1496927422	Mithrandir	surejob\.in
 1496949638	Ferrybig	hyderabadfairy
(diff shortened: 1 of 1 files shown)


NobodyNada:
@codecov-io [That's because the error handler doesn't get run](https://codecov.io/gh/Charcoal-SE/SmokeDetector/compare/d015612912618e0582d239545c2a544d7fd80fcf...4e49d2a01bb7d0b69db2164ca3a04697e3caec16/src/findspam.py)

tripleee:
!!/approve

PR #867, opened 2017-06-21
Require confirmation if a blacklist pattern is already caught

NobodyNada:
This came up yesterday, when someone tried to blacklist a keyword that was in the old `bad_keywords_nwb`.  I added a check, similar to `!!/test`, before each blacklisting operation so that confirmation is required if blacklisting a keyword that is already caught.

Since the pattern to blacklist is a regex and not just a keyword or URL, this is not perfect, since the blacklist *pattern* is run through the filters and not necessarily what the pattern was intended to catch.  I replace `\W` with a space and `\.` with a period, so it should work for most blacklist/watchlist operations.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -291,6 +291,70 @@ def command_blacklist_help(*args, **kwargs):
                     "!!/blacklist-keyword, or perhaps !!/watch-keyword. "
                     "Remember to escape dots in URLs using \\.")
 
+def check_blacklist(string_to_test, is_username):
+    # Test the string and provide a warning message if it is already caught.
+    if is_username:
+      question = Post(api_response={'title': 'Valid title', 'body': 'Valid body',
+                  'owner': {'display_name': string_to_test, 'reputation': 1, 'link': ''},
+                  'site': "", 'IsAnswer': False, 'score': 0})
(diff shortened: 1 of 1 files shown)


NobodyNada:
Hmm; I'm unsure what to do about the codeclimate review.  The "Too many parameters" ones seem mostly unavoidable, and the "Similar code" ones are incorrect (it's complaining about the docstrings), leaving only "Method too long."

ArtOfCode-:
"Too many parameters" can be solved either by splitting a method into multiple (which also solves "method too long"), or by using `**kwargs`.

NobodyNada:
Thanks!

ArtOfCode-:
Apparently I can also make the CC status check pass manually, so... that's a thing

NobodyNada:
Alright; would you still like me to make those changes?  I probably would have used `**kwargs` had I known about it then; that's pretty cool!

ArtOfCode-:
Yeah, go ahead on `**kwargs`. I won't make you pass the codecov/patch check, though ;)

NobodyNada:
Alright; thanks!  `check_blacklist` would be be pretty trivial to test; I can add a quick test if you like.

From Jonathan

ArtOfCode-:
!!/approve

PR #871, opened 2017-06-22
Fix !!/watch

NobodyNada:
`GitManager` expected a different item format when watching instead of blacklisting.  I changed it to format  the new item itself.

I haven't tested it because I accidentally ruined my dev environment, but it should work fine.

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -7,6 +7,7 @@ import requests
 import time
 import json
 import regex
+import datetime
 from threading import Lock
 if 'windows' in str(platform.platform()).lower():
     # noinspection PyPep8Naming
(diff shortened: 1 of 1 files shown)


tripleee:
You need to `import datetime` in order to be able to use `datetime.now()`

tripleee:
You need to `import datetime` in order to be able to use `datetime.now()`

gitmanager.py
```diff
@@ -88,6 +89,8 @@ def add_to_blacklist(cls, **kwargs):
             # Set up parameters for watch vs blacklist
             if blacklist_file_name in ['watched_keywords.txt']:
                 op = 'watch'
+                now = datetime.now().strftime('%s')
+                item_to_blacklist = "\t".join([now, username, item_to_blacklist])
                 item = item_to_blacklist.split('\t', 2)[2]
```
tripleee: It would be more elegant to save the original `item` from `item_to_blacklist` so you don't have to `split` what you just `join`ed.  But we can fix that up later.

PR #873, opened 2017-06-22
nocrash.py: add errorlog.txt logging

tripleee:
Hopefully reveal something more about why we are restarting.


angussidney:
!!/approve

PR #907, opened 2017-06-30
updating test-reload watch

doppeldeveloper:
`text\Wreload` catches `test.reload` code, so i'd like to make this a bit more specific. all matches so far are hyphen or space.


tripleee:
!!/approve

PR #916, opened 2017-07-03
Morning routine 20170703

tripleee:
Batching a number of edits to avoid multiple restarts.

### DIFF
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -1241,3 +1241,4 @@ gctrm\.com
 gtznf\.com
 gnula\.mobi
 gelkeyfim\.com
+naijadailyfeed\.com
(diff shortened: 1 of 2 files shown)


tripleee:
!!/approve

ArtOfCode-:
Approvals are there to have more than one pair of eyes on a PR, so self approving kinda defeats that. PullApprove should have ignored it, but maybe the config was wrong.

angussidney:
@ArtOfCode- but similarly, Tripleee could have made these changes directly to master, but batched them together in order to save resources, so another approval isn't really necessary.

tripleee:
I agree that the PR was sort of gratuitous in this situation, though. I did my morning routine this morning just by merging locally. (Though a PR would have helped avoid a problem I managed to introduce, by running the test suite before merging. But I'll keep that in mind to include in my local routines from now on.)

PR #961, opened 2017-07-15
Better handle history 404 errors, so we don't have to trigger an error.

teward:
We have a couple cases where message history on deleted messages are pulled, and trigger 404 errors that may lead to harder crashing where they aren't captured proper.

Thus, let's add some 404-error protection for message history 404 errors.  If such an error happens we return nothing.  If it's any other type of 404 error in the message handling, raise that error.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -62,7 +62,7 @@ def print_chat_message(ev):
     log('info', message + colored(" - " + ev.data['user_name'], attrs=['bold']))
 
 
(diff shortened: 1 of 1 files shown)


tripleee:
Can you point to a scenario where this could be tested?

teward:
@triplee Kinda hard actually.  This was observed when you try and pull chat history links which apparently Smokey does, and then can't see because it was deleted or such.

PR #962, opened 2017-07-16
Switch palyaçom.com from the blacklist to the watchlist

NobodyNada:
It's only had 3 hits, all from the same user.  I mean to put it on the watchlist, but I'm tired and I did the wrong thing without noticing.


ArtOfCode-:
!!/approve

PR #963, opened 2017-07-16
Properly handle dict loading with error handling, etc. - ROUND 2

teward:
Round 2 for this.

Basically, we have the advanced load_pickle, and it works if there's pickles existing.  If there's no pickle files existing, it seems to properly handle "empty" or "invalid" pickles and returns the proper empty list or dict objects.  This seems to work fine, and TCoM was running and restarted a handful of times without a repeat of the issues saw yesterday by the same approaches.

### DIFF
--- a/datahandling.py
+++ b/datahandling.py
@@ -36,6 +36,7 @@ def _load_pickle(path, encoding='utf-8'):
             os.remove(path)
             raise
 
+
 # methods to load files and filter data in them:
 def load_files():
     if os.path.isfile("falsePositives.p"):
--- a/datahandling.py
+++ b/datahandling.py
@@ -13,71 +13,57 @@ import math
 import regex
 
 
+def _load_pickle(path, encoding='utf-8'):
+    with open(path, mode="rb") as f:
+        try:
+            return pickle.load(f, encoding=encoding)
+        except UnicodeDecodeError:
+            f.close()
+            os.remove(path)
+
+            if "apicalls" in path.lower():
+                return {}
+
+            if "latestmessages" in path.lower():
+                return {GlobalVars.meta_tavern_room_id: [], GlobalVars.charcoal_room_id: [],
+                        GlobalVars.socvr_room_id: [], '111347': []}
+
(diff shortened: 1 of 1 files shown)


datahandling.py
```diff
@@ -13,71 +13,58 @@
 import regex
 
 
-# methods to load files and filter data in them:
+def _load_pickle(path, encoding='utf-8'):
+    with open(path, mode="rb") as f:
+        try:
+            return pickle.load(f, encoding=encoding)
+        except UnicodeDecodeError:
+            f.close()
```
tripleee: With a context manager (`with open()...`) you don't need to (and indeed cannot) explicitly `.close()` -- the context manager does that for you, even when there is an error.

datahandling.py
```diff
@@ -13,71 +13,58 @@
...
+                return {}
+
+            if "latestmessages" in path.lower():
+                return {GlobalVars.meta_tavern_room_id: [], GlobalVars.charcoal_room_id: [],
+                        GlobalVars.socvr_room_id: [], '111347': []}
+
+            if "bodyfetcher" in path.lower():
+                return {}
+        except Exception as e:
+            raise e
```
tripleee: Catching an exception just so you can `raise` it is useless.  Just take out the `except Exception` branch.

j-f1: What’s wrong with this?
```python
try:
    try:
        try:
            try:
                try:
                    try:
                        result = 1 / 0
                    except Exception as e:
                        raise e
                except Exception as e:
                    raise e
            except Exception as e:
                raise e
        except Exception as e:
            raise e
    except Exception as e:
        raise e
except Exception as e:
    raise e
```

teward: Ultimately, nothing, but since tripleee's harping on that, I've updated the code to remove this.

datahandling.py
teward: Fixed.

teward:
@tripleee Can you review again and adjust your review status please?  I made additional changes.

ArtOfCode-:
!!/approve

PR #965, opened 2017-07-18
findspam.py: bad_ns_for_url_domain()

tripleee:
Identify rogue name server (for now, concentrate on the Indian pharma spammer's favorite namecheaphosting.com) in domain names in URLs.

requirements.txt: pull in dnspython

findspam.py: post_links(): refactor into a separate function so that multiple methods can invoke it.


tripleee:
Notice that this is likely to incur significant additional latency, enough to dominate over all our regular expression rules combined.  I have had no way to test this yet, but reviewing recent spam indicates that those Indian pharma spams which didn't get autoflagged would be triggered by this.

angussidney:
Additionally, I'm hesitant about blacklisting all domains resulting to this nameserver. After all, Namecheap supposedly manages [7 million domain names](https://en.wikipedia.org/wiki/Namecheap), and they *do* partner with GitHub to offer free domains to students as part of the [GitHub Education Pack](https://education.github.com/pack), which makes them look somewhat legitimate.

j-f1:
Would it be possible to cache the results of the probe in a pickle so each domain only gets checked once?

tripleee:
I've been thinking about caching, yes; but the caching nature of DNS itself means that repeated queries will only touch the local name server until the TTL times out.
If that's insufficient, implementing a longer local TTL or storing results in a database (or sure, a picke) could be useful.

findspam.py
```diff
@@ -316,6 +321,20 @@ def bad_pattern_in_url(s, site, *args):
...
 
+def bad_ns_for_url_domain(s, site, *args):
+    for domain in set([get_domain(link) for link in post_links(s)]):
+        try:
+            ns = dns.resolver.query(domain, 'ns')
+        except dns.exception.DNSException as exc:
+            # TODO: log exception?
+            continue
+        nameservers = [server.target.to_text() for server in ns]
+        if any([ns.endswith('.namecheaphosting.com.') for ns in nameservers]):
```
ArtOfCode-: This is an ideal candidate for expansion. Instead of only `.namecheaphosting.com.`, we'd ideally make this a blacklist of almost-always-bad nameservers, and report the post for any of them.

tripleee: I completely agree; but let's see if this is feasible at all before generalizing it, shall we? Devising a extension mechanism would complicate this proof of concept, potentially needlessly if it turns out to be too heavy and/or FP-prone. I can't think of another provider that we could realistically blacklist at this point, anyway.

Undo1: Agree with @tripleee on this one. Do the simplest thing that might work first; from here we have two easy paths: (1) rollback, or (2) add a rule field for it. Seems like a pretty good idea to me.

Undo1:
I do have reservations about blacklisting a whole name server, but I don't see a reason we shouldn't try it. It won't touch autoflagging, the remaining concern is performance. How long does it take to get these results? DNS is quick, so I doubt it's significant, but want to check.

angussidney:
@Undo note that this will affect autoflagging because it's a new rule. It'll need to be added to be experimental rules list and have its weight pinned down in MS before we test it out.

Undo1:
@angussidney Metasmoke already pins reasons with fewer than 20 posts to 0: https://github.com/Charcoal-SE/metasmoke/blob/b68c535f456bf5c7b6ed17e3c9f6eeda6525db0a/app/helpers/reasons_helper.rb#L18

ArtOfCode-:
And after that, if it's terrible and catches loads of FP's, it'll have a tiny weight anyway.

tripleee:
The opposite is more likely - the rule could quickly rack up 20 TP hits from India but it might later degrade significantly if they switch wholesale to a new provider, or spread themselves thin across multiple providers like they already do with domain names and IP addresses (but this is somewhat less likely because then they lose the probably large bulk registration discount they are likely to get by centralizing domain registrations with just a few providers).

tripleee:
For what it's worth, I am piggybacking a test of this into the deepsmoke experiment I'm running on an Azure host provided by Undo. So far (500-ish messages scanned) no false positives at least, though subjectively I think I can perceive some latency from the DNS queries.

There's a separate chat room where I post occasional observations and updates; https://chat.stackexchange.com/rooms/64277/deepsmoke-a-charcoal-subsidiary

tripleee:
The test run is still in progress, but it produced its first and so far only match a little while ago. This was a TP -- it correctly flagged a previously unseen domain as suspicious.

https://chat.stackexchange.com/transcript/message/39615953#39615953

So far, the test run has scanned a bit over 2,000 messages today. The match was on post 1694 in the current run (though it scanned a dozen or so before it was restarted earlier today).

tripleee:
The experiment ran out of quota but it got 5/5 TP on 3957 messages.  There is obviously some latency (informally, seems to me more than for Deepsmoke) but I think it's bearable.

With that, can we proceed to discuss merging this? Or do we need more latency metrics?

Undo1:
Latency is probably fine, although I'm interested in the 'ran out of quota' part - any details there?

It'd be interested at some point to have a built-in way to push these longer checks off to another thread, but probably not necessary before merging.

tripleee:
It consumed the 10k API quota, no big mystery there. I guess I might have some inefficiencies in the code (it's all on Github in the {{deepsmoke-demo-double-down}} branch in my private clone if you want to have a look).

Undo1:
Ah, okay. It'd be fine to use the 20k key we have; it's all the same project we were granted it for, and I have no qualms about using it for R&D.

tripleee:
As an aside, regarding Angus' comment up there about 7 million domains, that's probably whois registrations, not customers who buy DNS service. This is looking for the latter. (Of course, that in no way rules out the risk that we could still face significant false positives, if not now then maybe surprisingly and invonveniently when some big site changes DNS providers.)

j-f1:
You could add a whitelist for domains that are hosted by namecheap but aren’t spam.

tripleee:
Sure, that too!

angussidney:
@Undo1 @SulphurDioxide @teward @NobodyNada could you please install the `dnspython` library on your instances and then update my requested changes review or leave a comment here to let us know that you've done so.

SulphurDioxide:
Henders/EC2 has `dnspython` now

tripleee:
Tangentially, should we think about making the whole runtime available as a Docker container or something like that? It won't directly help here but then at least you could ensure that the environment upgrades will be one-click operations.

Undo1:
I don't have mine running right now and won't for a while, don't worry about my instances.

teward:
@tripleee Docker sounds like a good idea but let's wait until we have a centralized solution for all the pickles and such before we delve into that.  Docker also has an added risk - git credentials can be tricky unless key auth is used and Docker isn't super kind there.

ferrybig:
I'm against docker support, this would mean dropping support for all Windows versions expect windows 10 pro and higher again.

tripleee:
*Having* Docker is a convenience for those who can use it, not an impediment for those who can't. But that discussion probably belongs in a separate ticket.

NobodyNada:
NobodyNada/RPi and qcpi have `dnspython`

tripleee:
Thomas indicates that his instances are now updated (Solar Flare now, Lunar Eclipse later tonight):
https://chat.stackexchange.com/transcript/message/39727495#39727495

tripleee:
Could we please go ahead with this now? I think we have enough test results to give it a whirl in production. (Angus already offered to do that on Friday but I was away for the weekend so we postponed it.)

tripleee:
As agreed in chat on Friday, proceeding to merge this now.

PR #968, opened 2017-07-18
removed musixmatch\.com

superplane39:
Isn't a spam site - I've used it myself. It was watched for being used as an example of a site that had a lot of lyrics, so https://metasmoke.erwaysoftware.com/post/72014 probably shouldn't have been deleted as spam.


tripleee:
!/approve

tripleee:
!!/approve

Issue #949, opened 2017-07-14
Posts which trigger only experimental reasons should only generate notifications in Charcoal HQ


tripleee:
There is a code path which is supposed to do this, but it's apparently not working.  Here is an experimental-only hit which was evidently reported in SObotics:

https://chat.stackexchange.com/transcript/message/38729208#38729208

tripleee:
Looks like it actually works precisely the wrong way around!

https://metasmoke.erwaysoftware.com/post/76990 only triggered an experimental (watch) reason. It was [not reported in Charcoal](https://chat.stackexchange.com/transcript/message/38869987#38869987) but [was reported in SOCVR](https://chat.stackoverflow.com/transcript/41570?m=32819119#32819119).

ArtOfCode-:
Sounds like someone over-negated a boolean....

tripleee:
Closing as hopefully fixed.  We'll have to wait for a low-score post to see whether it behaves as expected but if not, just reopen.

tripleee:
This one was reported in Charcoal and I believe only in Charcoal, so this seems to be properly done:

https://chat.stackexchange.com/transcript/message/38894750#38894750

Tangentially, that's an FP related to https://github.com/Charcoal-SE/SmokeDetector/issues/948

PR #970, opened 2017-07-20
Attempt to fix Charcoal only for all experimental (#949)

tripleee:
Looks like the old logic would *exclude* Charcoal and post to all other rooms (sic).

Also wrapped some horribly overlong lines in the subsequent code.

Fix for https://github.com/Charcoal-SE/SmokeDetector/issues/949 (I hope).


ArtOfCode-:
!!/approve

PR #972, opened 2017-07-21
findspam.py: whitelist djangogirls.org

tripleee:
This is a high-profile support community for the Python web framework Django.
Currently at 0/8 false positives in Metasmoke.


tripleee:
Miffed to find that those 8 false positives are the *only* hits ... What's up with this pattern?

https://metasmoke.erwaysoftware.com/search?body_is_regex=1&body=%5Cb%5B%5Cw-%5D%7B6%7Dgirls%5Cb

PR #973, opened 2017-07-21
Add private room for SagarV and Smoke Detector on meta chat

sagarvd01:
added a new private room only for reports. no other discussion.

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -200,7 +200,7 @@ GlobalVars.specialrooms = [
     {
         "sites": ["meta.stackexchange.com"],
         "room": GlobalVars.wrapm.get_room("1121"),
-         "unwantedReasons": []
+        "unwantedReasons": []
     },
     {
         "sites": ["bitcoin.stackexchange.com"],
--- a/ws.py
+++ b/ws.py
@@ -197,7 +197,7 @@ GlobalVars.specialrooms = [
         "room": GlobalVars.wrap.get_room("388"),
         "unwantedReasons": []
     },
-    {   
+    {
         "sites": ["meta.stackexchange.com"],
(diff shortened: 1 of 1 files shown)


tripleee:
Could you document here the rationale for adding another room on Meta? Smoke Detector is already in the [Tavern](https://chat.meta.stackexchange.com/rooms/89/tavern-on-the-meta) (and posts are periodically archived to [Chimney](https://chat.meta.stackexchange.com/rooms/1037/chimney)). If this is just for your personal needs, you can set up Smokey to ping you in chat with `!!/notify`, or you can install the spam alert userscript.

Glorfindel83:
@tripleee to be fair, !!/notify is rather broken.

tripleee:
For the record, the Travis failure is just the run-of-the-mill `flake8` whitespace flak.

AWegnerGitHub:
I'd also like to know the rational behind this.

ws.py
```diff
@@ -197,6 +197,10 @@ def check_socket_connections():
         "room": GlobalVars.wrap.get_room("388"),
         "unwantedReasons": []
     },
+    {   "sites": ["meta.stackexchange.com"],
```
tripleee: `flake8` dislikes this, but the fix should be trivial -- just add a newline after the opening brace and reindent.

ws.py
```diff
@@ -197,6 +197,10 @@ def check_socket_connections():
...
+        "room": GlobalVars.wrap.get_room("1121"),
```
tripleee: I *think* you will need to use `wrapm` rather than `wrap` to get the correct room.

AWegnerGitHub:
Rejected due to lack of explanation both here and in CHQ

PR #974, opened 2017-07-21
Add a "Pattern-matching email" reason

NobodyNada:
I added a "Pattern-matching email" reason, which catches emails containing patterns like "spell," "hack," and "temple."  There were a few of these patterns already in `bad_keywords.txt`, so I moved them into the new reason.


angussidney:
!!/approve

PR #983, opened 2017-07-27
Fix typo

NobodyNada:
I meant to blacklist "Geneticore Boost," but I'm on my phone and I blacklisted "Genericore Boost" instead.


Undo1:
!!/approve

PR #1030, opened 2017-08-23
Prevent "this isn't complex" etc. from being triggered.

Glorfindel83:
The `t\Wcomplex` regex in the blacklist is prone to false positives; fix it by requiring the word boundary before the first character to not be a single quote (which is also the conventional apostrophe in plain ASCII).


tripleee:
!!/approve

PR #1037, opened 2017-08-25
.pullapprove.yml: add NobodyNada

tripleee:
Chat link: https://chat.stackexchange.com/transcript/message/39612692#39612692


tripleee:
I had to sync the collaborators list on PullApprove before it would accept the change, but it's all good now. I'm mainly leaving this comment as troubleshooting help in case somebody needs to repeat this process sometime in the future.

PR #1052, opened 2017-08-31
findspam.py: bad_link_text(): fix regex

tripleee:
Refactor and (I hope) fix a couple of apparent errors in the long regex.


tripleee:
Besides wrapping the regex into more logical chunks (don't wrap related phrases across line breaks in the source), I changed a couple of `(^| )` and `( |$)` into `\b` (which should also match e.g. when preceded or followed by punctuation) and replaced a "wild" wildcard `.*` with a slightly less FP_prone `(?:\b.{1,20}\b)?`

angussidney:
!!/approve

tripleee:
This is waiting for a final merge.  Anyone ...?

Issue #1047, opened 2017-08-30
Dockerfile still refers to Python 2.7


tripleee:
The Dockerfile should be updated to Python 3.

Its purpose should probably be clarified, too.  It would be nice if it had all the functionality to run a production Smokey instance.

tripleee:
I'm not sure what exactly is required to run a prod instance. It should `git checkout deploy` and have the configuration file with the keys.  Anything else?

Undo1:
Credentials are the biggest thing. Maybe we could have an encrypted config file, with the password distributed to trusted folks as we do now with individual creds?

angussidney:
@Undo1 do you mean have everything pre-entered but encrypted, and then provide a password to unlock it? That won't necessarily work all the time, given that MS keys are separate and we don't always want to use the main Smokey account. 

A simple interactive program which directly prompts for SE username/pass (via CE), MS Key, GH creds, and location would allow more flexibility.

Undo1:
Good catch. Would be possible to split the file, but that's likely more trouble than it's worth.

tripleee:
The merge a week ago solved this.

PR #1055, opened 2017-09-01
Update Dockerfile for Python 3 etc (#1047)

tripleee:
Update Dockerfile to use Python3 and correctly pull in dependencies.  (#1047)
Also, create a separate SmokeDetector user instead of running as root.

* Remove obsolete MAINTAINER direcive
* Ubuntu provides python3-pip so no need to separately `wget` pip
* Install SmokeDetector with a virtualenv so that `tld` can be updated by SmokeDetector
* The image starts up with `su - SmokeDetector` by default
* Optimize by using a single `RUN` directive

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,22 @@
 FROM ubuntu
-MAINTAINER CaffeineAddiction
 
-RUN apt update && \
-    apt install build-essential python2.7 python-dev git wget -y && \
-    wget https://bootstrap.pypa.io/get-pip.py && \
-    python2.7 get-pip.py && \
-    cd ~ && \
-    env GIT_SSL_NO_VERIFY=true git clone https://github.com/Charcoal-SE/SmokeDetector.git && \
-    cd SmokeDetector && \
-    git submodule init && \
-    git submodule update && \
-    pip install -r requirements.txt --upgrade
-
-RUN apt remove -y --purge build-essential python-dev wget && \
+RUN DEBIAN_FRONTEND=noninteractive && \
+    apt -y update && \
+    apt install -y build-essential \
+        python3 python3-dev python3-pip python3-venv git && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,20 +1,22 @@
...
-
-RUN apt remove -y --purge build-essential python-dev wget && \
+RUN DEBIAN_FRONTEND=noninteractive && \
+    apt -y update && \
+    apt install -y build-essential \
+        python3 python3-dev python3-pip python3-venv git && \
+    adduser --disabled-password --force-badname SmokeDetector \
+        --gecos SmokeDetector && \
+    su - SmokeDetector sh -c '\
+        env GIT_SSL_NO_VERIFY=true git clone https://github.com/Charcoal-SE/SmokeDetector.git && \
```
j-f1: Why `GIT_NO_SSL_VERIFY`?

tripleee: I just copy/pasted it from the old `Dockerfile` but you get weird errors from `git` if you don't have the full certificate chain for SSL verification. I'd imagine installing the `ca-certificates` package would fix that more properly. Thanks for inquiring; I'll see if I can get rid of that.

teward: @tripleee Any updates on that?  Because I'd rather use valid SSL, not invalid SSL.  (`ca-certificates` is important after all to have so Smokey when it runs doesn't blow up on bad SSL)

tripleee:
Sorry for dropping the ball on this. I have now updated the `Dockerfile` and pushed the resulting image to https://hub.docker.com/r/tripleee/smokedetector/

The resulting image is still quite large. I'd be tempted to refactor it to use Alpine instead of Ubuntu but that's a separate effort, and also depends on what the image is supposed to be used for.

tripleee:
I struggled to find out how to mark the requested change as implemented. I guess the protocol is for @teward to approve the change.

PR #1057, opened 2017-09-01
findspam.py: bad_keywords_nwb: fix invalid escape sequence \W

tripleee:
For consistency, prefer r"...\W..." over "...\\\\W..." elsewhere in bad_keyword_nwb too
(i.e. use Python raw strings with single backslashes than regular Python strings where
regex backslashes have to be doubled).


angussidney:
!!/approve

PR #1058, opened 2017-09-01
GlobalVars.parser.unescape(): avoid deprecation warning

tripleee:
HTMLParser.unescape() was supposed to have been removed in Python 3.5
(but was allegedly still forgotten in).  The canonical replacement is simply
html.unescape().  Hack it in to the parser object so the existing code can
continue to work without deprecation warnings (though it's only actually
used in 3 places; we should really avoid using a global for this eventually).

Ref: https://stackoverflow.com/questions/2087370/decode-html-entities-in-python-string

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -4,6 +4,7 @@ import os
 from datetime import datetime
 from chatexchange_extension import Client
 from html.parser import HTMLParser
+from html import unescape
 from hashlib import md5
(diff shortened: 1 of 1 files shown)


tripleee:
Also chat link https://chat.stackexchange.com/transcript/message/39757762#39757762

PR #1079, opened 2017-09-12
nocrash.py: check git user name

tripleee:
Fail if wrong or unset unless running with no-git-user-check

### DIFF
--- a/nocrash.py
+++ b/nocrash.py
@@ -65,7 +65,8 @@ if 'no-git-user-check' in persistent_arguments:
 else:
     git_name = git.config('--get', 'user.name', _ok_code=[0, 1])
     if git_name != "SmokeDetector":
-        logging.error('git config user.name "{0}" is wrong'.format(git_name))
+        logging.error('git config user.name "{0}" is wrong; '
+                      'use no-git-user-check to ignore'.format(git_name))
         exit(122)
     git_mail = git.config('--get', 'user.email', _ok_code=[0, 1])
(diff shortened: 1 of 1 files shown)


nocrash.py
```diff
@@ -59,6 +60,19 @@ def error(message):
     logging.error('[NoCrash] {}'.format(message))
 
 
+if 'no-git-user-check' in persistent_arguments:
+    persistent_arguments.remove('no-git-user-check')
+else:
+    git_name = git.config('--get', 'user.name', _ok_code=[0, 1])
+    if git_name != "SmokeDetector":
+        logging.error('git config user.name "{0}" is wrong'.format(git_name))
```
NobodyNada: Maybe mention that the error can be bypassed; e.g. `git config user.name "{0}" is wrong; use no-git-check to ignore`

tripleee: Yeah, I convinced myself that short source lines would be more important than useful feedback, which is obviously not a good balance.  Hang on.

teward:
Y'know, if we're just checking statically if the credentials are set correctly, why don't we just let nocrash.py *set* the credentials if they don't exist, just for the Smokey repository?

tripleee:
I was considering that, but permanently overriding the user's preference e.g. in a development instance would be a huge misfeature IMHO. Granteoe, you could just say don't use nocrash if you aren't running a production instance.

ArtOfCode-:
!!/approve

PR #1081, opened 2017-09-13
!!/unwatch suganthan

Glorfindel83:
!!/unwatch suganthan - it's causing too many false positives (in usernames)


tripleee:
!!/approve

PR #1082, opened 2017-09-14
Exclude bad NS from git check

tripleee:
Don't require -force to blacklist or watch something which is triggered by these rules.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -292,7 +292,7 @@ def command_blacklist_help(*args, **kwargs):
                     "Remember to escape dots in URLs using \\.")
 
 
-def check_blacklist(string_to_test, is_username, is_watchlist):
+def check_blacklist(string_to_test, is_username, is_watchlist, filter_dns=False):
     # Test the string and provide a warning message if it is already caught.
     if is_username:
         question = Post(api_response={'title': 'Valid title', 'body': 'Valid body',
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
What's the rationale here? Why should this reason exempt a domain from the usual blacklisting requirements?

j-f1:
> Why should this reason exempt a domain from the usual blacklisting requirements?

Because a spammer could change the domain to be on another NS, but blacklisting will still be effective.

tripleee:
I agree that that's a useful discussion to have, but it's a broader issue than this one. If we agree that "bad NS" is solid enough on its own, maybe this PR should simply just be rejected for now. But then shouldn't we have more granular control across the board? I'd like for blacklisted X (including blacklisted NS) to be split into "really blacklisted X" (which should have basically 100% accuracy and a weight of 100, or even pinned to an even higher weight) and "suspicious X" (which can collect any number of speculative terms which may accrue FPs occasionally, and have the weight calculated accordingly).

ArtOfCode-:
That's what blacklisting and watching are for, @tripleee. The idea behind blacklisting requirements is that it's only going on the blacklist if we're absolutely sure it's spam.

tripleee:
Indeed, but there are terms and regexes within those lists which were added before we had a defined policy, or in violation of the policy, or which ended up producing false positives for whatever reason. What I'm trying to suggest is to make this three-way: watch (speculative heuristic, no substantial weight), suspicious (thought to be good heuristic, let it contribute to the weight according to the formula) and really black (basically proven good enough to autoflag by itself, though we may want to be slightly more conservative than that).

ArtOfCode-:
That sounds like excess complexity to me. Surely it'd be better to clean up the existing lists? Three lists makes an already complicated system even more effort to learn about.

quartata:
Why was this closed? I think all of this discussion is missing the point: checking NS is for discovering domains that *should* be blacklisted -- we shouldn't then discourage people from blacklisting them by making them use `-force`.

ArtOfCode-:
Closed because lack of activity. That doesn't mean no discussion.

PR #1100, opened 2017-09-23
Propose a 'support phone number' bad title kword

teward:
[This is the relevant MetaSmoke detection search](https://metasmoke.erwaysoftware.com/search?utf8=✓&title_is_regex=1&title=support%5CWphone%5CWnumber&body=&username=&why=&site=&feedback=&autoflagged=&reason=&user_rep_direction=>%3D&user_reputation=0&commit=Search), focusing on Title Only.  68/69 TPs, 1 FP, and 1 NAA.

After seeing a phone number detection in a post on SO where the capture pattern was *only* in the title for a suspicious phone number, I thought we might consider adding an additional 'support phone number' keyword detection.

While this adds another hard-coded keyword for title-only matching, it's more often than not, I think, only going to be seen in spam posts.  This could also be expanded to match other well known tech support 'scam' patterns that are primarily in the title, but I am only focusing on one currently.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -727,6 +727,10 @@ class FindSpam:
          'stripcodeblocks': False, 'body_summary': False, 'max_rep': 1, 'max_score': 0},
         {'regex': r'(?i)[\w\s]{0,20}help(?: a)?(?: weak)? postgraduate student(?: to)? write(?: a)? book\??',
          'all': True, 'sites': [], 'reason': 'bad keyword in {}', 'title': True, 'body': False, 'username': False,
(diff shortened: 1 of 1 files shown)


angussidney:
By the looks of it, these types of titles are already caught by this keyword:

https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L47

teward:
I’m not so sure about that - it didn’t capture the recent detections today as spam at all with that reason, only triggered the phone number detection.  This may need further investigation.



*Sent from my iPhone.  Please excuse any typos, as they are likely to happen by accident.*

angussidney:
https://chat.stackexchange.com/transcript/message/40308716#40308716

PR #1120, opened 2017-10-03
!!/rev includes !!/location output

tripleee:
Help sort out which instance is responding when there are several running


ArtOfCode-:
!!/approve

Issue #1028, opened 2017-08-21
Option to run Smoke Detector chatless/metasmokeless/gitless


tripleee:
For testing, it would be beneficial if Smoke Detector could be run in a completely local mode, i.e. with output only to the console, and no connection to various back-end services which it normally communicates with.

Ideally, the services it uses should be individually selectable, but for a start, a way to disable them all would encourage experimentation and perhaps even allow more people to participate in its development because setting up a private instance is both less visible (and so less embarrassing if you make silly mistakes, or have weird experiments which would pollute the chat room inordinately) as well as less technically intimidating.

---

- [x] Chatless
- [x] Metasmokeless
- [ ] Gitless

ArtOfCode-:
Obligatory sarcastic "edit your hosts file"...

But in seriousness, we're almost there already. Git and metasmoke are already entirely optional - just don't put them in config if you don't want them. Chat is less optional, but someone (Thomas) made a start on a no-chat mode that could be finished up.

teward:
I'm happy to try and build a chat-less mode, but it'll be a little tricky to build.  You can assign this to me in the mean time - we have such heavy chat integration for things, that I'll have to do some overhauling just to get this part working.  I also don't think I have any work on the no-chat mode, I only had fixes for 'failed to login to chat' issues.

quartata:
Chatless is trivial in the NG PR with an empty rooms.yml.

angussidney:
The 'gitless' point (which I'm assuming means using git for read-only tasks) should be complete when #1280 is merged.

quartata:
I'd say we're already decently Metasmoke-less as well

Issue #1131, opened 2017-10-04
What do we need in a central blacklist tracker?


Undo1:
As brought up in https://github.com/Charcoal-SE/metasmoke/issues/257, there's a desire to move blacklists out of GitHub and into some other tool. Whether that tool is metasmoke, another app, or some other alien-tech-powered solution, we'll want to know what it should *do*. Here's my first take:

 - Store blacklist entries categorized by type ([watch|blacklist] [website, keyword,username], etc.)
 - Have an easy-to-use way to add entries
    - From a web UI, instances, and/or metasmoke
 - Have reasonably good uptime
 - (?) See what changed since the last update?

Thoughts? Am I missing the mark? What else do we need?

BrockA:
Looks good to start.
Maybe also search/sort/graph by hits , and/or false positives, per entry?

j-f1:
It would be useful to have an independent web UI so blacklisting can still go on if MS is down. Maybe a page on https://charcoal-se.org?

AWegnerGitHub:
I've been looking at the links you provided [here](https://github.com/Charcoal-SE/metasmoke/issues/257#issuecomment-334050275) and thinking about this today. 

I think the blacklist would need 

 - `type`: blacklist-keyword, blacklist-username, blacklist-websites, watchlist
 - `text`: the pattern to blacklist. 
 - `created_at`: Just a simple timestamp of when it was added. I think we're just going to have to populate the list with the date when this is implemented initially
 - `modified_by`: Who to blame when a pattern fails; last person to touch this entry
 - `updated_at`: Obvious

From a front end, I think it could really be as simple as - select type, entry pattern. The rest can be automatically supplied by metasmoke. 

---

Since I've been thinking about this, I also think we need a table for notifications, since we lose those between instances right now. The layout of that table would be the same as we store it now, which I think it `user_id`, `site`, `chat_server`?

---

For your last bullet, the simplest method I came up with was maintaining a third table of last updated timestamps for each of the other tables. Every time one of the other tables is added/updated/deleted from, this table has the key associated with that table updated to the current time. 

```
list                timestamp
blacklists      2017-10-05 00:44:27
notifications 2017-10-04 12:15:00
```

Alternatively, we could do it based on table/key so the blacklist entry above could be broken down to something like 

```
list                type          timestamp
blacklists      keyword   2017-10-05 00:44:27
blacklists      user          2017-10-01 15:13:11
[...]

PR #1135, opened 2017-10-07
Backporting NG chat system

### LINKED ISSUE #852: !!/allspam should report all the user's posts to metasmoke
While [discussing !!/allspam](https://chat.stackexchange.com/transcript/message/38181920#38181920), @angussidney noted:

> Using !!/report with multiple arguments is preferred however, as then we get a copy of the post on MS, we can give feedback ,etc

While this is true, it's also more cumbersome (and not doable if the user has more than _x_ posts, as !!/report is rate limited). Instead, !!/allspam should be rewritten so that it reports all the user's posts to metasmoke.

Alternatively, we could make a new command !!/report-all &lt;link to user> for this.

quartata:
After much blood, sweat, tears <s>and blood</s>, I have successfully backported [NG's](https://github.com/Charcoal-SE/SmokeDetector-ng) `chatcommunicate` to the original SmokeDetector.

## TL;DR

- For non-developers:
   - SmokeDetector can run outside of Charcoal HQ (can run in any room however you like, really)
   - Adding SmokeDetector to new rooms no longer requires code changes (can be done temporarily until next restart with a command and made permanent with a simple config change)
   - Making SmokeDetector accept commands in rooms is a one-line config change (before it required significant code changes in several places, and there was instability for months when we added SmokeDetector to SOBotics)
   - It's possible to control what reasons get reported in your room without code changes.

- For developers:
    - New interface for chat commands, a simple decorator that replaces the massive dictionary and tons of parameters that go unused
    - Chatcommunicate is significantly shorter and cleaner -- so hopefully adding new chat-related functionality will be easier.
    - Retrieving report data from a message is easier.

### Room system:

A new configuration file, `rooms.yml`, has been added. This is the structure:

```
<chat host>:
  <room id>:
    commands: true/false,
    msg_types:
      - <role string...>
      - ...

    privileges: 
      - <user id...>
      - ...
```

This compactly defines all of the rooms SmokeDetector interacts with. The `commands` and `privileges` keys, as you might suspect, define whether SmokeDetector listens for commands in the room and who can run privileged commands respectively. The only thing that's new, really, is `msg_types`. The `msg_types`, in essence, defines the kinds of messages the room will receive. A few of the special ones:

- `debug`: this is what Charcoal HQ or your main room would have set. Rooms with this set will receive startup messages and other debugging information (basically anything that would have previously been done with `GlobalVars.charcoal_hq.send_message`).

- `all` or `site-<site>`: This dictates what sites will have their spam reported to the room. Rooms with `all` will receive reports from any site, while `site-<site>` (e.g `site-askubuntu.com` or `site-codegolf.stackexchange.com`) will receive reports from those particular sites.

- `no-<reason>`: This room won't receive reports with the specified reason (e.g `no-all-caps title`)

- `metatavern`: A special role I made for the Tavern on Meta, currently it just receives messages about blocks (for Town Halls).

- `experimental`: Only rooms with this role will receive reports consisting of only experimental reasons (watchlist).

To leverage these room roles, a flexible new function called `tell_rooms` has been added and replaces almost all of the previous `GlobalVars.special_rooms` or `GlobalVars.charcoal_hq` machinery:

```
tell_rooms(message, (roles to have...), (roles not to have...))
```

In essence, this broadcast the message to *any* room with one of the "roles to have," as long as it does not have *any* of the "roles not to have." An example:

```
tell_rooms("Hai", ("all", "site-askubuntu.com"), ("no-offensive answer detected"))
```

This will tell rooms with either `all` or `site-askubuntu.com` as long as they don't have `no-offensive answer detected`.

As a shorthand, there exists also `tell_rooms_with` and `tell_rooms_without` when there's only one role to have/not have:

```
tell_rooms("test", ("debug",), ())                   -> tell_rooms_with("debug", "test")
tell_rooms("meta tavern sucks", (), ("metatavern",)) -> tell_rooms_without("metatavern", "meta tavern sucks")
```

### Command system:

Defining a chat command is as simple as appending a decorator to your function:

```
@command(types..., reply=False, whole_msg=False, privileged=False, arity=None, aliases=None, give_name=False)
def <name_of_command>(...):
```

Most of the keyword arguments are not needed and can be omitted to have the default behavior. An explanation of each:

- `types...` - the most important part. These are a list of "functions" that are used to preprocess your arguments (and also define how many arguments your command takes). For instance, a command with `@command(int, int)` will take two strings and run them through `int`, coercing them to numbers. The most common "types" are Python's built-in `int` and `str` as well as chatcommunicate's `message`, which is used for reply commands that take a ChatExchange message object.

- `reply=<bool>` - if this is `False`, this command is a standard command (`!!/<name>`). If this is `True`, this is a subcommand (reply to a message, or use shorthand `sd commands...` notation). Subcommands are expected to have to take only one argument, and that argument will be the ChatExchange message object representing what message was replied to.

- `whole_msg=<bool>` - if this is `True`, the message containing the command will be passed in as the first argument. This is useful for retrieving data about who ran the command (`msg.owner`)

- `privileged=<bool>` - if this is `True`, only privileged users can execute this command.

- `arity=(<min arity>, <max arity>)` - an arity override, dictating the minimum and maximum arguments this function can take. Missing arguments will be `None`.

- `aliases=[strings...]` - other names this command can be referred to by (`true` vs `tp`, for instance).

- `give_name=<bool>` - if this is `True`, the alias used will be passed in as a keyword argument `alias_used`.

### Report data:

A nice bonus of all this is that I've made it so that `chatcommunicate` stores a dictionary mapping the message IDs of reports to a tuple containing the post URL and owner URL. This eliminates the ugly `fetch_post_url_from_msg_content` in most cases (if the message isn't in the dictionary for whatever reason, it'll run it automatically to parse it out). Instead, one can simply use the following to retrieve this information:

```
get_report_data(message)
```

where `message` is a ChatExchange message. Note that this dictionary is pickled along with last messages in the new `messageData.p`.

---

This PR also includes the new-and-improved allspam command. Please see https://github.com/quartata/SmokeDetector/pull/3 for details. 

---

Fixes #852 
See also: #1028


rooms.json5
```diff
@@ -0,0 +1,112 @@
+{
+    "stackexchange.com": {
+        "id": 120914
+
+        "11540": { # Charcoal HQ
+            msg_types: ["debug", "all"],
```
j-f1: These have to be `//`-style comments. If you want `#`-style comments, I’d recommend YAML.

rooms.json5
```diff
@@ -0,0 +1,112 @@
...
+                "268731"    # Sagar V
+            ]
+        },
+    },
+
+    "meta.stackexchange.com": {
+
+    },
+
+    "
```
j-f1: You’ve got an extra quote here.

rooms.json5
```diff
@@ -0,0 +1,112 @@
...
+                "164187",   # Okx
+                "202619",   # John Militer
+                "262693",   # suraj
+                "11287",    # Martin Sleziak
+                "88588",    # NVZ
+                "281674",   # paper1111
+                "279119",   # Tetsuya Yamamoto
+                "307652",   # Ajay Brahmakshatriya
+                "238145",   # Owen Hines
+                "268731"    # Sagar V
```
j-f1: If you put a trailing comma here, it will make diffs prettier when we add someone new.

chatcommunicate.py
```diff
@@ -1,275 +1,355 @@
...
+                _commands["reply"][alias] = cmd
+        else:
+            _commands["prefix"][func.__name__] = cmd
+
+            for alias in aliases:
+                _commands["prefix"][alias] = cmd
+
+        return f
+
+    return decorator
```
j-f1: Could you check if the only argument passed to `@command` is a function, and if so, `return decorator(type_signature[0])`? This way you could

```py
@command
def foo(...):
  # ...
```

quartata: Unfortunately the way the decorator works is that you pass in "types" which are really functions (int, float, str, message...) If you had something like @command(int) that would be indistinguishable from @command

AWegnerGitHub:
List of questions @quartata:

1. Can various `msg_types` be mixed? For example, `['site-askubuntu.com', 'no-all-caps title', 'no-Link at end of body']`
2. Are any options incompatible in `msg_types`? 
3. Does capitalization matter on the above reasons?
4. Can you explain `metatavern` a bit more? What does "currently it just receives messages about blocks (for Town Halls)" mean? The block command should work in all rooms.
5. What is passed if `whole_msg` is set to `False`?
6. What is the layout of `messageData.p`? This is something that could eventually be migrated to the central tracker and stuff mentioned in #1131. 
7. Can you provide an example function definition using `arity` and how it's called? I'm having a hard time seeing what this would be used for.

AWegnerGitHub:
On another note, why do we need json5? I'm not a fan of adding it because it makes any json files we create non-standard.

quartata:
1. Yes, of course. It wouldn't be very useful otherwise.
2. I suppose `all` and `site-<site>` don't make too much sense together, but it won't double post or anything. The point of the roles is that the only semantic meaning is in when you call `tell_rooms`.
3. It shouldn't? I haven't tried it uppercase...
4. When SmokeDetector is blocked in a room, it sends a message to Charcoal HQ about it. I made it so that this message is also sent to rooms with `metatavern` in addition to `debug`, since usually unblocking is done because of the Town Hall.
5. Just the preprocessed arguments (message parts after being run through your types).
6. It's a named tuple with two dictionaries (https://github.com/quartata/SmokeDetector/blob/master/chatcommunicate.py#L26). The first maps a tuple (chat host, room ID) to a deque of ChatExchange messages. The second maps message IDs to a tuple (post URL, owner URL).
7. A good example is `!!/errorlogs` -- when called with no arguments, it returns 50 lines, but when called with a number it returns that many lines. You set `arity=(0, 1)` to say that it take at minimum 0 arguments and at most 1 argument: https://github.com/quartata/SmokeDetector/blob/master/chatcommands.py#L547

JSON5 is needed for comments in JSON (primarily for the privileges list). AFAIK we don't actually use JSON anywhere else (we use INI for config).

j-f1:
@quartata What do you think about using YAML instead? It was designed from the beginning to be used for config files, while JSON was intended (I think) for data transfer.

quartata:
@j-f1 YAML could be good, although now that I'm thinking about it I could just use `ast.literal_eval` if I want to keep the Python dictionary syntax

angussidney:
I don't think evaluating the file is a good idea. I agree with JF: YAML seems like a good idea, expecially since we already use it for all of our CI providers

Undo1:
+1 for YAML. Haven't been able to look through the rest yet.

quartata:
Switched to YAML, will update PR description

rooms.yml
```diff
@@ -1,103 +1,396 @@
...
+      - 262693   # suraj
+      - 11287    # Martin Sleziak
+      - 88588    # NVZ
+      - 281674   # paper1111
+      - 279119   # Tetsuya Yamamoto
+      - 307652   # Ajay Brahmakshatriya
+      - 238145   # Owen Hines
+      - 268731   # Sagar V
+
+  3877:
```
j-f1: Which room is this?

angussidney: Raiders of the Lost Downboat, the AskUbuntu room, I'll add a comment to the code

quartata:
@codecov-io bad bot

ArtOfCode-:
@codecov-io good bot. @quartata sorry, I'm going to be that guy - we need plenty of good tests on this. Major change, high breakability, yada yada, automated tests good.

chatcommunicate.py
```diff
@@ -139,6 +139,9 @@ def on_msg(msg, client, room):
             if room_data.last_report_data != ():
                 _last_messages.reports[message.id] = room_data.last_report_data
 
+                if len(_last_messages.reports) > 50:
```
j-f1: You might want to change this to a `while` to keep the count down if multiple reports accidentally get added, sending the total above 51.

quartata: @j-f1 This would only happen if another thread added to it. I'll add a lock instead

AWegnerGitHub:
Completely informational: This pull request was mentioned as one of the things that will make implementing a central database easier. This was mentioned specifically on my [Helios proposal](https://gist.github.com/AWegnerGitHub/f17e4f65c089712f888de429323cd86b), but I imagine these changes will be helpful regardless of which option is picked.

angussidney:
Just an update here on GH so that it's properly documented:

This PR is basically finished in terms of functionality. Currently, @quartata and I are slowly working to increase the coverage by adding new tests for the various commands, so that codecov will pass. Any help is appreciated, because tests are the only thing that is stopping this from being merged. 

Additionally, if someone could test this on Windows, that would be awesome. There isn't any reason why it shouldn't work, but it would be a good idea to test it anyway.

angussidney:
Over the weekend I'm going to fix a bug with the allspam command, but after that we should be free to merge (hopefully on Monday). 

If anyone has any comments/concerns, now's the time to add them.

ArtOfCode-:
How are we looking wrt merging?

quartata:
Need to make a test for !!/hats, then need to fix allspam

ArtOfCode-:
Is that stuff that has to happen on this PR, or can we merge it so that that work can happen in parallel with the other stuff that needs to use this code?

Bear in mind `!!/allspam` gets used very rarely.

j-f1:
@quartata Can you fix the merge conflict?

quartata:
I guess we're all ready...

PR #1151, opened 2017-10-15
Remove test --autopull

bwDraco:


bwDraco:
Done manually through the SmokeDetector account, so no need.

PR #1159, opened 2017-10-18
Move URLs to website blacklist --autopull

bwDraco:
Some URLs were erroneously put in the keyword blacklist, so I'm moving them to the website blacklist.


tripleee:
!!/approve

PR #1165, opened 2017-10-21
Fix sd delete the easy way

KazWolfe:
- Don't call force-delete from delete. Instead, just take care
  of it then and there.


ArtOfCode-:
You have a Flake error

PR #1166, opened 2017-10-23
Fix watch pattern mcfly -> mcfly\.aero to avoid fps --autopull

ByteCommander:


ByteCommander:
!!/approve

superplane39:
!!/approve

tripleee:
The reason we require approval is so that every commit gets reviewed by a second pair of eyes. Thus, you are not supposed to `!!/approve` your own PR, and the code won't accept it.

PR #1170, opened 2017-10-25
Make mostly dots calculate length *before* stripping code blocks

quartata:
Currently mostly dots strips code blocks first, then calculates the ratio of dots to characters in the post. This causes a lot of FPs when posts have long code blocks but not much else. This change makes it so that the length of those code blocks is still factored in, just not any dots in them.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -541,6 +541,9 @@ def mostly_dots(s, site, *args):
     body = strip_urls_and_tags(s)
     body_length = len(body)
 
+    body = regex.sub("(?s)<pre>.*?</pre>", "", body)
+    body = regex.sub("(?s)<code>.*?</code>", "", body)
+
     dot_count = len(regex.findall(r"\.", body))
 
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Even if your post only has one sentence outside of a code block, it still shouldn't be mostly dots... Am I missing

quartata:
@ArtOfCode- That's how it used to work (because it used `stripcodeblocks`, so it never got to see the code block). This fixes that.

j-f1:
@ArtOfCode- I think that @quartata is saying that we shouldn’t be catching posts that looks like this:

```md
    class MyClass(SuperClass):
        # pretend there’s lots of code here
    print(MyClass().foo('bar'))

.......................
```

SE says they can’t post a question with just code, but they don’t have anything to say. This makes it a bad question, but it’s not actually spam and Smokey shouldn’t catch it.

ArtOfCode-:
!!/approve

PR #1178, opened 2017-10-29
Scan for R/A posts using Perspective

quartata:
We (read: Undo at my bequest) recently obtained an API key for [Perspective](https://www.perspectiveapi.com/), Google's ML service that lets you scan comments with a variety of pre-trained models trained to detect toxic comments. This is an experimental (read: I haven't tried it yet) reason that uses it to scan question and answer bodies and reports any that score higher than 50% (rather aggressive, will probably be increased as we try it out).


findspam.py
```diff
@@ -1095,6 +1120,13 @@ class FindSpam:
          'max_rep': 20, 'max_score': 0},
     ]
 
+    # Toxic content using Perspective
+    if GlobalVars.perspective_key:  # don't bother if we don't have a key, since it's expensive
```
AWegnerGitHub: What does"rather expensive" mean? Are we going to have performance issues?

quartata: No. It's one POST, which is expensive compared to the other things we do but it's not that slow. There's just no point in running it if we don't have the Perspective key in the config (it's a secret, so it might not be there if someone is say running a test instance).

NobodyNada:
I've been pretty out of the loop lately (been busy with school) so I don't know whether this has been discussed, but in case anyone didn't know, Petter has been trying out Perspective with HeatDetector in SOBotics.

IIRC it worked fairly well.  It wasn't perfect, and I think Petter's existing regex and NLP stuff worked quite a bit better, but it'll give better results if you set a threshold higher than 50%.  It felt like it was *very* keyword-based, and it frequently caught false-positives like "I'm an idiot."

I also remember Petter holding off from sending feedback to Perspective because he was worried about licensing, since Google stores the post text when they're sent feedback.

That's my fuzzy recollection;  I haven't been in chat much in the past few months.  If you want more details, you can ask him -- he's pingable in Charcoal, you can probably find him in SOBotics, and his GH profile is [jdd-software](https://github.com/jdd-software).

quartata:
I agree that 50% threshold will probably be too low to start with but I wanted to start low and see just how many FPs we get -- the Holocaust troll scores around 0.58 on some of the posts I tested which is my motivation for pulling off a lower threshold.

I've talked with Petter quite a bit actually -- among other things, he recommended trying the SEVERE_TOXICITY model which I'll give a try once Undo mails me the key.

ArtOfCode-:
Any reason this shouldn't be merged?

quartata:
@ArtOfCode- Default threshold is a little too low. I bumped it up to .70 locally, but was also working on an improvement that would lower the score based on how many first-person pronouns there were (which would hopefully allow for a lower overall threshold)

quartata:
We were also previously failing on coverage, but I suspect that coveralls will be more lenient (this is an old codecov check)

angussidney:
Have you run a test of this to see how it goes when put up against real SE data?

quartata:
Yes, it works well. 0.85 is a pretty conservative threshold, but going any lower runs into issues with stuff like "I am stupid <...>" (which I have a fix for, but for now I just want to have it up to catch the most severe stuff). I have it in as an experimental reason for now anyways.

angussidney:
!!/approve

quartata:
don't merge this yet, we need to give everyone the key

ArtOfCode-:
@quartata If you want to send me the key via Keybase, I can distribute it on if you like

PR #1183, opened 2017-10-30
Enhance "Already recently reported" by adding MS link

### LINKED ISSUE #990: Enhance "Already recently reported" smokey response.
When you try and !!/report something that has been previously reported Smokey replies as follows:

> @Glorfindel Post 1: Already recently reported

It would be useful to modify the Smokey reply to include **a link to the recent report** so we can easily find it and take a look at it.

paper1111:
When`!!/report` is used on something that is previously reported, the response is 

> @XXX Post X: Already recently reported

With this PR, the message would be

> @XXX Post X: Already recently reported [MS]

With a link to the MS report.

Closes #990

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1411,9 +1411,12 @@ def command_report_post(ev_room, ev_user_id, wrap2, message_parts, message_url,
                 (post_data.post_id, post_data.site)):
             # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
             # this re-report might be attempting to correct that/fix a mistake/etc.
-            se_link = parsing.to_protocol_relative(post_data.post_url)
-            ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(se_link)
-            output.append("Post {}: Already recently reported [[MS]({})]".format(index, ms_link))
+            if GlobalVars.metasmoke_host is not None:
+                se_link = parsing.to_protocol_relative(post_data.post_url)
+                ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(se_link)
+                output.append("Post {}: Already recently reported [[MS]({})]".format(index, ms_link))
+            else:
+                output.append("Post {}: Already recently reported".format(index)
             continue
         post_data.is_answer = (post_data.post_type == "answer")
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -1411,7 +1411,9 @@ def command_report_post(ev_room, ev_user_id, wrap2, message_parts, message_url,
                 (post_data.post_id, post_data.site)):
             # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
             # this re-report might be attempting to correct that/fix a mistake/etc.
-            output.append("Post {}: Already recently reported".format(index))
+            se_link = parsing.to_protocol_relative(post_data.post_url)
+            ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(se_link)
+            output.append("Post {}: Already recently reported [[MS]({})]".format(index, ms_link))
```
angussidney: Could you add a check to see whether `GlobalVars.metasmoke_key` exists before appending the MS link? No point adding it unless if there is actually a report on MS

ArtOfCode-:
You've got a CI failure

paper1111:
Oops... I don't have time now, will fix tonight

paper1111:
@ArtOfCode- I've fixed the syntax error

ArtOfCode-:
!!/approve

chatcommands.py
```diff
@@ -1411,8 +1411,14 @@ def command_report_post(ev_room, ev_user_id, wrap2, message_parts, message_url,
...
             # this re-report might be attempting to correct that/fix a mistake/etc.
-            output.append("Post {}: Already recently reported".format(index))
-            continue
+            if GlobalVars.metasmoke_key is not None:
+                se_link = parsing.to_protocol_relative(post_data.post_url)
+                ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(se_link)
+                output.append("Post {}: Already recently reported [[MS]({})]".format(index, ms_link))
+                continue
+            else:
+                output.append("Post {}: Already recently reported").format(index)
```
angussidney: This is incorrect syntax for .format. It needs to be called on the string itself. Fixed in 4d576d8

PR #1185, opened 2017-10-30
Include an SE search link in blacklist PRs

NobodyNada:
When reviewing a blacklist/watchlist PR, I often do a [global Stack Exchange search](https://stackexchange.com/search) for the keyword or URL, just to be sure the keyword isn't used in non-spam posts.  I modified the auto-generated blacklist PR message to add an SE search link.

While I was at it, I also changed the MS search link to use `urllib`'s `quote_plus`, which converts spaces to pluses and percent-escapes other special characters.

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -157,7 +157,7 @@ class GitManager:
 
                 payload = {"title": u"{0}: {1} {2}".format(username, op.title(), item),
                            "body": u"[{0}]({1}) requests the {2} of the {3} {4}. See the Metasmoke search [here]"
-                                   "(https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93{5}{6}), and the"
+                                   "(https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93{5}{6}) and the"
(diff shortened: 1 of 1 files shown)


gitmanager.py
```diff
@@ -156,11 +157,12 @@ def add_to_blacklist(cls, **kwargs):
 
                 payload = {"title": u"{0}: {1} {2}".format(username, op.title(), item),
                            "body": u"[{0}]({1}) requests the {2} of the {3} {4}. See the Metasmoke search [here]"
-                                   "(https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93{5}{6})\n"
+                                   "(https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93{5}{6}), and the"
```
j-f1: Can you 🔥 the `,` here? I think it would read better as `See the metasmoke search [here] and the Stack Exchange search [here]`.

NobodyNada: I agree; that reads a little better, and I think the comma was grammatically incorrect.  Thanks!

ArtOfCode-:
!!/approve

PR #1191, opened 2017-11-02
Git lockdir

tripleee:
Smokey gets wedged if you blacklist or watch something while a previous Git operation is still running in CI.  Prevent that by creating a local lock directory which is removed when you pull.

The `gitmanager_lock` is probably redundant if this is accepted, but I am no longer entirely sure whether there were some other failure scenarios which it covered differently...


ArtOfCode-:
!!/approve

PR #1215, opened 2017-11-08
Neither SE nor regular Metasmoke supports regex search

tripleee:
Search links in the generated PR text were still wrong. Use the same expression for both Metasmoke and StackExchange search. (Both are still pretty brittle, because we are trying to convert a regex to a plain search string.)


ArtOfCode-:
!!/approve

PR #1216, opened 2017-11-09
Gitmanager refactor

quartata:
This does two things to help improve Smokey's Git workflow:

- Adds a pre-commit hook that checks to see if we're on deploy, and refuse to commit if so. This is to prevent people from accidentally committing to deploy and creating nasty local changes.
- Simplify the blacklisting process and avoid checking out or modifying the local master branch (in case it's not up to date or has local changes):
    - Fetch origin/master from Github
    - Check it out into a detached HEAD
    - Make blacklist commit
    - If the user has code privileges, push directly to master from detached HEAD
    - Else, checkout a new temporary branch and make a PR.

Note that at no point do we commit directly to the local master or even check it out. This also eliminates any merging, which would run into issues if the changes couldn't be fast-forwarded.


quartata:
This also eliminates the notorious "HEAD isn't at the tip of origin's master branch" error.

tripleee:
I'm here but I'm waiting for the revamp which avoids detached head for privileged users.

https://chat.stackexchange.com/transcript/message/41039077#41039077

ws.py
```diff
@@ -35,6 +35,11 @@
 from tld.utils import update_tld_names, TldIOError
 from helpers import log
 
+import os.path
+
+if not os.path.islink(".git/hooks/pre-commit"):
+    os.symlink("hooks/pre-commit", ".git/hooks/pre-commit")
```
j-f1: What about copying `hooks/*` into `.git/hooks/` on each launch? This keeps the scripts up-to-date without the issues surrounding symlinks.

gitmanager.py
```diff
@@ -133,13 +122,6 @@ def add_to_blacklist(cls, **kwargs):
                     blacklist_file.write("\n")
                 blacklist_file.write(item_to_blacklist + "\n")
 
-            # Checkout a new branch (for PRs for non-code-privileged people)
-            branch = "auto-blacklist-{0}".format(str(time.time()))
-            git.checkout("-b", branch)
-
-            # Clear HEAD just in case
-            git.reset("HEAD")
```
teward: This should be kept *somewhere*.  We need to make sure we're actually on `HEAD` and not in a detached or reverted state.  Which is incredibly important.

gitmanager.py
```diff
@@ -148,13 +130,13 @@ def add_to_blacklist(cls, **kwargs):
                        "-m", u"Auto {0} of {1} by {2} --autopull".format(op, item, username))
 
             if code_permissions:
-                git.checkout("master")
-                git.merge(branch)
```
teward: So we're letting GH handle *all* of the commits then?  We had this here because if you have code privs you should already push this but also *locally* merge it.  As you refactor it, this will prevent in-place activation of changed things if one has code perms.

teward: This also was built to push to master directly, hence why we checked *out* master to begin with.  your changes will fubar this process.

gitmanager.py
teward: In your test case in CHQ room, you detailed why this is important - you clobber remote commits that may be just as important.  Therein lies the core problem and why we left this here.

AWegnerGitHub: Discussion: https://chat.stackexchange.com/rooms/11540/conversation/pr-1216

AWegnerGitHub:
Pulling this out of an embedded comment, discussion about this in CHQ: https://chat.stackexchange.com/rooms/11540/conversation/pr-1216

quartata:
As we're currently discussing in chat, I think there might be a bit of a misunderstanding here. Nothing about this will clobber remote commits -- the only window of opportunity for a remote commit to cause a conflict is between the fetch as the push which is a few seconds at max, and if it did occur the push would simply fail (Git never tres to fast forward commits when pushing.)

As for local commits, they're all still there -- they just won't be pushed with the blacklist (by design). Presumably they were created by a human, so a human should resolve and push them.

gitmanager.py
quartata: This is a valid point, however I'm inclined to prefer that we wait for CI.

quartata:
I should add that the `finally` clause makes it so that it always switches back to deploy and deletes any temporary branch ade no natter what, if say the push did fail. I should probbly add in a `stash` just in case deploy has uncommitted changes.

quartata:
I've set up another repository with a clean Git tree with my modifications, and confirmed that blacklisting (at least calling `gitmanager.GitManager.add_to_blacklist` directly) works. Feel free to experiment.

quartata:
Pull requests also work, and no local deploy or master changes were harmed: https://github.com/quartata/smokey-git-test/pull/1

PR #1230, opened 2017-11-15
FP fix for tcl.tk

tripleee:
Exclude the main TCL site from the *.tk answer regexes.

I can find no good way to test this so just throwing a PR in the hopes that someone can validate the fix.


Glorfindel83:
Looks to good to me, except that regex101.com claims you need to escape the slash (in `[/.]`, so `[\/.]`). Does that work differently in Python?

j-f1:
Yes — that only is necessary if the language uses `/` to delimit regular expressions.

angussidney:
!!/approve

Issue #1133, opened 2017-10-05
Report which blacklist pattern(s) triggered the report


j-f1:
Currently, we can only see what was matched. It would be useful to see the regex pattern that matched the URL, allowing us to check if patterns are performing well.

quartata:
This turned out to be way too slow so we're probably going to revert it

tripleee:
!!/bisect now provides this.

PR #1275, opened 2017-11-26
Show which pattern was matched in the why data

quartata:
This adds a new function called `blacklist_factory`, which acts as a wrapper that takes in a list of keywords (either `wb` with word boundaries or `nwb` without word boundaries) and transparently constructs a regex out of them and returns a function that tests a given string against it. Most importantly however, it wraps each keyword in its own named group. Then, when constructing the why data, it uses the last matched named group to retrieve which keyword was matched. A bit of a hack, but it's the best we can do without hooking into the regex engine or maybe using something like Perl's Regexp::Assemble.

Currently the `blacklist_factory` is employed for the `bad_keywords.txt` keywords, `bad_keywords_nwb`, `blacklisted_usernames.txt`, `blacklisted_websites.txt` and `watched_keywords.txt`. It could be used on some of the other fixed regexes (like pattern-matching websites) to enable this effect if desired.

This closes https://github.com/Charcoal-SE/SmokeDetector/issues/1133.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -561,6 +561,39 @@ def mevaqesh_troll(s, *args):
         return False, ""
 
 
+def blacklist_factory(wb=[], nwb=[], **kwargs):
+    pattern = r'(?is){}'.format("|".join(chain(
+        (r"(?P<wb{}>\b{}\b)".format(i, keyword) for i, keyword in enumerate(wb)),
+        (r"(?P<nwb{}>{})".format(i, keyword) for i, keyword in enumerate(nwb))
+    )))
+
+    compiled_pattern = regex.compile(pattern, **kwargs)
+    del pattern
+
+    def f(s, *args):
+        why = []
+
+        for match in compiled_pattern.finditer(s):
+            group = match.lastgroup
+            start, end = match.span()
+
+            if group[:3] == "nwb":
+                matched = nwb[int(group[3:])]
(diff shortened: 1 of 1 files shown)


quartata:
Incidentally, this pattern could also be achieved with a callable object instead of a closure. This might be better if we want to add additional state or maybe say a `fetch()` method to reload the blacklist.

ArtOfCode-:
What's the performance like on this? If you compare runs/sec for the old function and this one, what do you get?

quartata:
I don't have exact numbers but when running strings through the full FindSpam suite there was no noticeable difference. NFA capture groups are fast and the last named group was already tracked, so we're mostly adding one function call and some parsing. Peanuts compared to an NS query

quartata:
I was mostly concered about memory usage, which is why I delete the regex pattern after compilation (something that is not currently done at all - it's recompiled on every check)

quartata:
I would add that if you're concerned about time prformance, findspam is currently a hot mess: most egregiously, tests are run on all fields (title, body, username) before even checking whether those tests should be run on those. If they shouldn't, the result is thrown out. A transition to the `whole_post` model (one of the first things I ever added, and the only option for NG) would make at least method tests cleaner and probably faster.

ArtOfCode-:
!!/approve

PR #1295, opened 2017-11-29
Optimize BlacklistFactory

quartata:
This changes two things:

- replace the closure with a callable object
- instead of using named groups, construct a mapping of regular group indices to line numbers. When I tried this previously, it didn't work out because any groups in the keywords themselves would throw off the count, but now I count the number of groups in each keyword when constructing the mapping.

This went from a 2x slowdown to a little above a 1.25x slowdown.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -561,39 +561,52 @@ def mevaqesh_troll(s, *args):
         return False, ""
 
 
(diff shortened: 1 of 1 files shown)


Undo1:
!!/approve

PR #1343, opened 2017-12-06
Implement fixes for quartata

NobodyNada:
https://chat.stackexchange.com/transcript/message/41585771#41585771


quartata:
!!/approve

PR #1370, opened 2017-12-12
Use asyncio Tasks

quartata:
This swaps out many of our short-lived threads (things like sending stats to Metasmoke) for efficient [asyncio Tasks](https://docs.python.org/3/library/asyncio.html) that run on a single dedicated thread.

The base of this is a new class `Tasks` with three methods:

- `Tasks.do`: takes in a function and arguments and schedules it to be run as soon as the thread is free (first in, first out)
- `Tasks.later`: takes in a function and arguments and schedules to be called after a certain delay in seconds (specified by the keyword argument `after=`)
- `Tasks.periodic`: takes in a function and arguments and schedules it to be called every `interval` seconds

Some examples:

    Tasks.do(metasmoke.Metasmoke.send_deletion_stats_for_post, post_url, True)
    Tasks.later(restart_automatically, after=21600)
    Tasks.periodic(Metasmoke.send_status_ping, interval=60)

I have swapped out most of our one-off thread usage for this, with three exceptions:

- `deletionwatcher.post_message_if_not_deleted` and `deletionwatcher.check_if_report_was_deleted`: these still spawns a thread for each report, since the `websocket` module doesn't offer a way to monitor the websocket for deletion asynchronously. (I may have an alternate way though, still thinking about it.)
- `bodyfetcher post enqueing`: I wasn't entirely certain why this was running on a thread at all, so I chose not to touch it for now.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -15,8 +15,8 @@ from metasmoke import Metasmoke
 from parsing import *
 from spamhandling import handle_spam
 from gitmanager import GitManager
+from tasks import Tasks
 import threading
-from threading import Thread
 import random
 import requests
 import os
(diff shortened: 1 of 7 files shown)


ArtOfCode-:
You've got a codeclimate issue

quartata:
@ArtOfCode- Yeah, since I edited deletionwatcher it wants me to fix it. It's not that complicated so I'll oblige it tho

ArtOfCode-:
!!/approve

PR #1381, opened 2017-12-14
Listen to chat sockets via poll(), eliminate unused CE worker thread

quartata:
This prunes a lot of threads that weren't really doing too much.

1. Previously, ChatExchange would spawn one thread per websocket (that's one thread per room we listen to commands in, essentially). These changes make it so that there's only one thread, which uses an efficient [poll()](https://docs.python.org/3/library/select.html#select.poll) to monitor all the sockets simultaneously. When at least one of them is available to read from, the thread unblocks and dispatches the appropriate callback like normal.

2. This also eliminates the mostly-unused `_worker` thread, which ChatExchange spawns one per chat host in order to do operations like sending or deleting messages asynchronous. In the case of sending messages, we already have our own thread that manages that as well as managing the `_last_messages` for us, and we don't really need to delete messages asynchronously, so it's not really useful.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1189,7 +1189,7 @@ def delete_force(msg):
     """
     # noinspection PyBroadException
     try:
-        msg.delete()
+        msg._client._do_action_despite_throttling(("delete", msg.id, ""))
     except:
         pass  # couldn't delete message
 
@@ -1210,7 +1210,7 @@ def delete(msg):
                "#a-note-on-message-deletion) for more details."
     else:
         try:
-            msg.delete()
+            msg._client._do_action_despite_throttling(("delete", msg.id, ""))
         except:
             pass
 
@@ -1228,7 +1228,7 @@ def postgone(msg):
     if edited is None:
         raise CmdException("That's not a report.")
 
-    msg.edit(edited)
+    msg._client._do_action_despite_throttling(("edit", msg.id, edited))
 
 
 # noinspection PyIncorrectDocstring
@@ -1266,7 +1266,7 @@ def false(feedback, msg, alias_used="false"):
 
     try:
         if int(msg.room.id) != int(GlobalVars.charcoal_hq.id):
-            msg.delete()
+            msg._client._do_action_despite_throttling(("delete", msg.id, ""))
     except:
         pass
 
--- a/chatexchange_extension.py
+++ b/chatexchange_extension.py
@@ -1,8 +1,11 @@
 # coding=utf-8
-from ChatExchange.chatexchange import client, rooms
-import sys
+from ChatExchange.chatexchange import browser, client, rooms
 from datetime import datetime
-from helpers import log
+import json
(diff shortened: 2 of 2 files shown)


chatcommands.py
```diff
@@ -1189,7 +1189,7 @@ def delete_force(msg):
     """
     # noinspection PyBroadException
     try:
-        msg.delete()
+        msg._client._do_action_despite_throttling(("delete", msg.id, ""))
```
j-f1: Could you make a helper for this (i.e. `messages.delete(message)`)?

chatcommands.py
```diff
@@ -1228,7 +1228,7 @@ def postgone(msg):
     if edited is None:
         raise CmdException("That's not a report.")
 
-    msg.edit(edited)
+    msg._client._do_action_despite_throttling(("edit", msg.id, edited))
```
j-f1: Same here — `messages.edit(msg, edited)`

quartata: This is the only instance where we edit a message (and this command doesn't work anyways and hasn't been used in forever), so it's probably fine

chatcommands.py
quartata: Sure, I'll include the try too

chatcommands.py
j-f1: 🔥 the command then?

quartata: I've considered it but I'll leave it in in case someone decides to revamp it. Basically the problem is that the regex used needs to be updated to account for Metasmoke links (yes it's that old)

j-f1: It just doesn’t seem to have any use — the post link is still useful for ♦️s, and AIM will tell you it’s been deleted, even after the two minutes is up.

quartata:
@ArtOfCode- OS X is absolutely supported (I develop on OS X). Windows too, but I believe only for sockets or named pipes (that's all I use though)

quartata:
The only advantage of having one thread per room would be for parallelization, which doesn't happen anyways because of the global interpreter lock. In exchange we waste memory with stacks for each thread and time schedule the!.

ArtOfCode-:
Negative on Windows support. Just tested with both 2.7.13 and 3.4.4 - `module object has no attribute poll`.

ArtOfCode-:
So the question then becomes one of a trade-off: is a marginal waste of memory worth a marginally better socket response time?

ArtOfCode-:
[Chat]: this also gets us better thread safety, which is generally a good thing to have. AFAIC, that tips the trade-off in favour of this approach, but it does still need Windows support.

quartata:
Just wanted to note (as discussed in chat) that this will not make socket response time slower (at least it shouldn't): `poll()` is a system call that blocks until at least one of the registered sockets is available to read from, and isn't really "polling."

ArtOfCode-:
What's the status of Windows support on this?

PR #1388, opened 2017-12-15
Added new rule for titles that end with comma

magisch:
Series of troll posts across Interpersonal.Stackexchange over the last couple days.

Plus, it's fairly accurate with 416 <-> 8 rate
https://metasmoke.erwaysoftware.com/search?autoflagged=&body=&commit=Search&feedback=&feedback_filter=tp&reason=&site=&title=.%2A%5C%2C%24&title_is_regex=1&user_rep_direction=%3E%3D&user_reputation=0&username=&utf8=%E2%9C%93&why=

Requested by Catija (IPS Moderator)

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1142,7 +1142,10 @@ class FindSpam:
         {'method': mostly_dots, 'all': True, 'sites': ['codegolf.stackexchange.com'],
          'reason': 'mostly dots in {}', 'title': True, 'body': True, 'username': False, 'body_summary': False,
          'stripcodeblocks': False, 'max_rep': 50, 'max_score': 0},
-
+        # Title ends with Comma (IPS Troll)
+        {'regex': r".*\,$", 'all': False, 'sites': ["interpersonal.stackexchange.com"],
+         'reason': "title ends with comma", 'title': True, 'body': False, 'username': False, 'stripcodeblocks': False,
+         'body_summary': True, 'max_rep': 50, 'max_score': 0},
         #
         # Category: other
         # Blacklisted usernames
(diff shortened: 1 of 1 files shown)


magisch:
Please review carefully this is the first time I've added a rule all on my own so it may be wrong

tripleee:
No particular need to backslash-escape the comma; it's harmless, but ideally we should adhere to style guidelines for regexes, too. (Sorry, not in a place where I can make a direct edit.)

ArtOfCode-:
At 98% across all sites, this is accurate enough to be added as a network-wide rule, not just IPS. 2 posts on IPS isn't really a big enough sample to judge its accuracy there.

findspam.py
```diff
@@ -1142,7 +1142,10 @@ class FindSpam:
         {'method': mostly_dots, 'all': True, 'sites': ['codegolf.stackexchange.com'],
          'reason': 'mostly dots in {}', 'title': True, 'body': True, 'username': False, 'body_summary': False,
          'stripcodeblocks': False, 'max_rep': 50, 'max_score': 0},
-
+        # Title ends with Comma (IPS Troll)
+        {'regex': r".*\,$", 'all': False, 'sites': ["interpersonal.stackexchange.com"],
+         'reason': "title ends with comma", 'title': True, 'body': False, 'username': False, 'stripcodeblocks': False,
+         'body_summary': True, 'max_rep': 50, 'max_score': 0},
```
ArtOfCode-: `body_summary` is the first 300 or so characters of the body, which we get from the realtime socket. I don't think we want to scan that for this rule, so it can be `False`.

magisch: Done. Also made it for all sites on the network since you appropriately said it's accurate enough for that.

ArtOfCode-:
!!/approve

Undo1:
Turns out metasmoke isn't a representative dataset. This is the clearest case of this issue I've seen. [PR to revert](https://github.com/Charcoal-SE/SmokeDetector/pull/1395) in case I'm missing something.

PR #1395, opened 2017-12-20
Revert rule for titles that end with comma

Undo1:
Looks like we were misled in https://github.com/Charcoal-SE/SmokeDetector/commit/c0b52fe1a35dab9ef3fcdc290754e374983edfb7 by metasmoke's natural tendency toward true positives. [It's not doing well](https://metasmoke.erwaysoftware.com/reason/140) on real data from the network. The detected posts are 3/13, and the true positives have plenty of weight without this reason (which metasmoke is still clamping to 0).


ArtOfCode-:
Agree that it's not doing well, but instead of removing it we should (for now) just restrict it to the originally suggested scope: just IPS.

Undo1:
Done.

Undo1:
!!/approve

PR #1399, opened 2017-12-20
Add spam check "title starts and ends with a forward slash"

paper1111:
There are [quite a number of posts](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title_is_regex=1&title=%5E%5C%2F.*%5C%2F%24&body=&username=&why=&site=&feedback=&autoflagged=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search) that starts and ends with a forward slash.


ArtOfCode-:
Has this been checked against the wider SE dataset? The last rule we added based on MS data didn't go so well...

paper1111:
Well [this query](https://data.stackexchange.com/stackoverflow/query/773026) for SO returns no results if I am doing it correctly, I may now try to run a test instance.

PR #1413, opened 2017-12-24
del-force only for reports

### LINKED ISSUE #1139: Have priority levels for which messages can deleted by Smokey without -force
It would be handy if you could delete *certain* messages with `sd poof`, but not others. For instance: being able to `poof` a !!/alive message would be able to be done without the -force flag. But a report shouldn't be allowed to be just `poof`ed; that'd need a -force flag. As it is right now, everything in Charcoal HQ requires a -force flag.

paper1111:
This PR implements the request for levels of deletion in #1139. Currently, all messages in CHQ have to be deleted with -force, but with this PR only reports need to be deleted with -force.

(Fixes #1139)

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1203,8 +1203,10 @@ def delete(msg):
     :return: None
     """
     if msg.room.id == 11540:
-        return "Messages from SmokeDetector in Charcoal HQ are generally kept "\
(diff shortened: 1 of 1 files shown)


angussidney:
!!/approve

PR #1425, opened 2017-12-30
corrected regex for the last pull request fd4ee51

surajrao:


tripleee:
The dash should not be backslashed, could you please correct that? The backslash is harmless as such but we don't want to perpetrate this hypercorrect antipattern.

surajrao:
@tripleee done

Issue #1442, opened 2018-01-03
Duplicated phone numbers in "bad_keyword.txt"


iBug:
On [Line 233](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L233), there's a pattern `(?<=\+)234\d{8,14}` that covers any phone number that starts with `+234`. Meanwhile, there are 9 other lines about individual phone numbers starting with `234` (no plus sign). I suggest removing the duplicated lines and remove the positive lookbehind for the plus sign on Line 233.

The dupe lines are [425](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L425), [461](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L461), [533](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L533), [549](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L549), [564](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L564), [571](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L571), [661](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L661), [937](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L937), [1007](https://github.com/Charcoal-SE/SmokeDetector/blob/master/bad_keywords.txt#L1007).

Line 233 modification proposal:

```
\+?234\W*\d{3}\W*\d{3}\W*\d{4,10}
```

iBug:
As PR #1446 is merged already, this issue should be considered [status-completed].

PR #233, opened 2016-10-03
Blacklisted few websites

Fortunate-MAN:
https://metasmoke.erwaysoftware.com/post/42261
https://metasmoke.erwaysoftware.com/post/42266
https://metasmoke.erwaysoftware.com/post/42284

PR #1446, opened 2018-01-04
Remove duplicates

iBug:
Remove duplicated phone number pattern according to [Issue #1442](https://github.com/Charcoal-SE/SmokeDetector/issues/1442). Minor merges.


iBug:
Bad maths for me. Please change Line #233 to 

```
\+?234\W*\d{3}\W*\d{3}\W*\d{2,8}
```

tripleee:
Maybe just drop the plus sign actually, or maybe even move the generic Nigerian pattern to the watch list instead.

iBug:
@tripleee PR updated.

tripleee:
Thanks for the update, but now the 234* phone numbers which used to be blacklisted are no longer blacklisted. I'm fine with moving the generic +234 detection to watch but then the individual numbers we have blacklisted should remain blacklisted.

Sorry to be a pest, I appreciate the effort, and sorry I have not been able to communicate my intents properly.

iBug:
@tripleee I understood your request. I've made changes correspondingly.

PR #1483, opened 2018-01-11
Remove hash comments from blacklist --autopull

iBug:
Automatically ignore all lines starting with a hash (treat as comments and ignore) in `blacklist_*.txt` and `bad_keywords.txt`


ArtOfCode-:
!!/approve

PR #1495, opened 2018-01-12
Update findspam.py --autopull

iBug:
Attempt to fix the seemingly missing character at the end of reports like `Title: Position A-B hatever.co` when the actual domain is `whatever.com`.

Also fix `Click Here` not being captured at all.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -907,7 +907,7 @@ class FindSpam:
          'sites': [], 'reason': "blacklisted website in {}", 'title': True, 'body': True, 'username': False,
          'stripcodeblocks': False, 'body_summary': True, 'max_rep': 50, 'max_score': 5},
         # Suspicious sites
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -907,7 +907,7 @@ class FindSpam:
          'sites': [], 'reason': "blacklisted website in {}", 'title': True, 'body': True, 'username': False,
          'stripcodeblocks': False, 'body_summary': True, 'max_rep': 50, 'max_score': 5},
         # Suspicious sites
-        {'regex': r"(?i)({}|({})[\w-]*?\.(co|net|org|in(\W|fo)|us|blogspot|wordpress))(?![^>]*<)".format(
+        {'regex': r"(?i)\b[\w-]*?({}|({})[\w-]*?\.(com?|net|org|in(fo)?|us|blogspot|wordpress))(?![^>]*<)".format(
```
ArtOfCode-: The `\b` is unnecessary - all these regexes get joined together into one giant regex that adds word boundaries automatically.

PR #1512, opened 2018-01-15
Add some juices

iBug:


ArtOfCode-:
We've typically pushed back on new joke commands, since it's another thing to break that doesn't add critical functionality. We haven't had a proposal in a while, so I don't know how others stand on this, but I'd rather not.

AWegnerGitHub:
I agree. No new joke commands - especially ones that are missing tests. That coverage report is really misleading. I'm not sure how coverage increased here.

teward:
No new joke commands please.

teward:
NOTE: Given that AOC, AWegner, and myself are all against 'new' joke commands,  I closed this PR.  Joke commands are 'fun' and all, but as said they introduce more breakage points, and serve no critical purpose.  (I can understand augmenting existing ones for more 'options' like for the coffee command to choose espresso or something, but that was done alongside a much more major refactor - on its own that change wouldn't have stood).

Please don't suggest more 'joke' commands, we don't need them, and they're just another breakage point.

PR #1536, opened 2018-01-21
Improve "no whitespace" detection --autopull

iBug:
The `\S+` detection for "no whitespace" starts to catch valid one-line answers with enough punctuation. [For example](https://metasmoke.erwaysoftware.com/post/102944). Changing to "word" (letters+digits+underscore) should solve this issue (although underscore isn't necessary)


ArtOfCode-:
You've used `\W` here. The regex convention is that the lowercase character class is the regular character class, and the uppercase name is the inverse. So, `\w` is "word character", and `\W` is "non-word character". The (rough) equivalent of `\S` (not-space) is `\w` (word).

PR #1544, opened 2018-01-22
Whitelist common file extensions --autopull

iBug:
Basically I just whitelisted a few common file extensions from being detected as *misleading link*. They often show up as

```
Here is my code: [findspam.py](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py)
```

Which is then reported as

```
Domain *github.com* indicated by possible misleading text *findspam.py*.
```

Hopefully that only shows up on Stack Overflow.

See the metasmoke search [here](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=%5BA-Za-z0-9_%5D*%5C.%28py%7Chtm%7Cjava%29&reason=139&commit=Search). That's a good 48 pieces of craps. (Note: the search isn't 100% accurate as parts of the text that's not the domain are shown in search results, but they won't be detected by the code. `parsed_text.tld` ensures only the domain is checked.)

Currently only 4 file extensions are excluded: `py`, `htm`, `java` and `sh`. They make up the majority of the FPs.


iBug:
Meh. What the heck is that PullApprove error???

findspam.py
```diff
@@ -31,8 +31,8 @@
 COMMON_MALFORMED_PROTOCOLS = [
     ('httl://', 'http://'),
 ]
-# These types of files frequently gets caught as "misleading link"
-SAFE_EXTENSIONS = ['txt', 'js', 'htm', 'html', 'css', 'php', 'py', 'java', 'rb']
+# These types of files frequently get caught as "misleading link"
+SAFE_EXTENSIONS = set(('txt', 'js', 'htm', 'html', 'css', 'php', 'py', 'java', 'rb'))
```
j-f1: I think you can do `{'txt', 'js', 'htm', 'html', 'css', 'php', 'py', 'java', 'rb'}` here.

AWegnerGitHub:
How does this handle things like

    Visit my [homepage.php](http://maliciouslink.com)

?

I think we are going to be hiding things we don't want to hide.

---

Additionally, looking at the example search you've provided, the very first one I [clicked on ](https://metasmoke.erwaysoftware.com/post/102974) isn't related to this. The user put an imgur screenshot and linked to Amazon in the [first revision](https://stackoverflow.com/revisions/48374799/1). That is clearly a misleading link.

iBug:
@AWegnerGitHub While I agree it's a possibility, it's unlikely to happen given the general questioning styles on SO. Furthermore `amazon.com` is not being excluded by this tweak: `com` isn't a "safe extension".

AWegnerGitHub:
I can't approve this until there are test cases. I am hesitant to approve untested whitelists

iBug:
@AWegnerGitHub The search isn't quite accurate. It shouldn't match any part except the domain. My code is exempt from that. You can take the code and feed it with some posts.

findspam.py
quartata: True. I personally prefer the set constructor syntax since the braces one is easy to mistake for a dictionary

iBug:
@AWegnerGitHub I have the test cases ready [here](https://github.com/iBug/iBug.github.io/blob/master/ext/misc/awegner_test.zip). You can verify the code by yourself.

iBug:
Python, HTM(no L) and Java codes make up the majority of the FPs. No need to whitelist the rest as they only generate 1 or 2 FPs each.

findspam.py
```diff
@@ -99,7 +101,7 @@ def malicious_link(s, site, *args):
     try:
         parsed_href = tld.get_tld(href, as_object=True)
         print(parsed_href.domain, SE_SITES_DOMAINS)
-        if parsed_href.tld in SE_SITES_DOMAINS:
+        if site == 'stackoverflow.com' and parsed_href.tld in SE_SITES_DOMAINS:
```
ArtOfCode-: I think you put this in the wrong place... Should it be with SAFE_EXTENSIONS on 117?

iBug: My bad...

PR #1555, opened 2018-01-24
doppelgreener forgot to add the .com

CalvT:


CalvT:
iBug watched the domain yesterday (which I didn't see) so this isn't needed.

PR #1554, opened 2018-01-24
Improve regex

paper1111:
The fixes the error where some answers get reported as a question. As seen is [this message](https://chat.stackoverflow.com/message/40942991) some characters are added by SE. Therefore, this pull request edits the regex to cover that situation.

Fixes #1428

PR #1556, opened 2018-01-24
Fix report fix

paper1111:
Originally in #1554 I tried to fix `!!/report`, but it does not seem to work. Therefore, the new solution is to trim the extra characters added by SE first and then parse the id out.


paper1111:
Ehh I think it STILL doesn't work, not sure why

paper1111:
The second regex should be updated too, should work now

PR #1571, opened 2018-01-26
+FrankerZ in SOBotics

danbopes:
Already privileged in SOCVR and Charcoal HQ. Would like privs here to reduce some spam down in the SOCVR room, as I'm usually hanging around in the SOBotics room often.


jdd-software:
ok with me Franker knows well how SD works

ArtOfCode-:
!!/approve

PR #1573, opened 2018-01-26
Shorten URL links

iBug:
Shorten URL for questions and users (I bet you folks don't know users link can be shortened to `/u/<UserID>`, so I bring it up here).


quartata:
Why the test edits?

iBug:
@quartata @AWegnerGitHub @teward I changed the old tests because they are no loner correct. The change is intended to make Smokey generate shorter links, and leaving old tests will only break the test because it won't generate longer links anymore.

quartata:
You're going to need to change more functions in parsing to accommodate that most likely.

quartata:
Also it would be nice if you would write proper commit messages so we know at a glance what a commit does.

iBug:
@quartata I'll remember that next time. I've gone through the whole file and made sure everything's right

iBug:
Wait, I messed something up while trying to edit commit messages.

iBug:
@quartata @teward @AWegnerGitHub Updated all the changes. Please review.

angussidney:
I’m not 100% sure, but some of the report parsing regexes/some userscripts rely on the message having unshortened links, make sure these aren’t broken by this change.

quartata:
https://github.com/Charcoal-SE/SmokeDetector/blob/master/parsing.py#L182

iBug:
@quartata Did you see the comment is edited in my patch?

ArtOfCode-:
Pretty sure metasmoke relies on links having `/questions/` or `/a/` in them rather than the shortened form...

**`[status-don't-merge]`**

iBug:
@ArtOfCode- Maybe change MS first, to let it recognize both `/questions/` and `/q/`, before merging this one.

ArtOfCode-:
I'm gonna [status-declined] this one. Thanks for your work, @iBug, but given that this is likely to require metasmoke and userscripts changes as well and the gains are minimal, I don't think we win on the benefit/loss scales.

PR #1575, opened 2018-01-29
Migrate "dumbass" from blacklist to offensive

iBug:
bwDraco noticed it's not caught and blacklisted it. IMO this word should better be caught as "offensive keyword" so I'm moving it into the code.


bwDraco:
I'm not too sure if we should include "badass" in this regex. Unlike "dumbass", "badass" is not universally offensive, and can have positive connotations in some contexts.

bwDraco:
!!/approve

Undo1:
!!/approve

PR #1578, opened 2018-01-30
Add new URL shortener, per link below

gparyani:
https://m.erwaysoftware.com/posts/by-url?url=//expatriates.stackexchange.com/a/13069

This post was not caught by the automated filters, because it used a short URL that isn't on the blacklist.


Undo1:
!!/approve

PR #1604, opened 2018-02-09
Update penis/dick length regex to catch more units --autopull

bwDraco:
The "penis/dick x inches" regex at line 814 failed to catch a wingding post which used cm instead of inches (https://metasmoke.erwaysoftware.com/post/98237). This commit changes the regex to handle metric units as well as abbreviations, as well as when the length value and unit are not separated by a space.


Undo1:
!!/approve

PR #1609, opened 2018-02-10
Update p**** regex

NobodyNada:
Catches plurals, and substitutions like [this](https://m.erwaysoftware.com/posts/by-url?url=//stackoverflow.com/a/48725632).


NobodyNada:
Oops, looks like I modified the wrong regex, give me a minute please.

PR #1610, opened 2018-02-11
Add an experimental reason to catch Donald Trump usernames

NobodyNada:
We get [a lot](https://metasmoke.erwaysoftware.com/search?utf8=✓&title=&body=&username_is_regex=1&username=don%28ald%29%3F%28+%2Aj%29%3F+%2Atrump&why=&site=&feedback=&autoflagged=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search) of trolls named after Trump.  The latest of these has posted several answers we have not caught without `!!/report`.

I added an experimental reason to catch users matching `don(ald)?(\W*j)?\W*trump`.  There are 37 `tp`s, 2 `fp`s, and 3 `naa`s in Metasmoke.  On the entire network, there are [38 users](http://data.stackexchange.com/stackoverflow/query/801521/users-named-donald-trump) named Donald Trump or a variant, but only [nine of these](https://gist.github.com/NobodyNada/ac38fa0f4bcf6aaa4b5285b68baed16d) have non-deleted posts.

I added the reason to the end of findspam, with the name "Potentially bad username."  Since the reason contains "username," whitelisting a user will prevent them from being caught by this reason.  I also added this reason to the experimental reason list (although this doesn't cancel its autoflagging weigh, which requires MS console access as far as I can tell.)

### DIFF
--- a/blacklisted_usernames.txt
+++ b/blacklisted_usernames.txt
@@ -65,4 +65,4 @@ emilie\W?larsen
 air\W?purifier\W?reviews
 Cum\W?Juice
 agen judi‭
-donald\W*j?\.?\W*trump
+donald\W<i>j?.?\W</i>trump
--- a/blacklisted_usernames.txt
+++ b/blacklisted_usernames.txt
@@ -65,4 +65,3 @@ emilie\W?larsen
 air\W?purifier\W?reviews
 Cum\W?Juice
 agen judi‭
-donald\W<i>j?.?\W</i>trump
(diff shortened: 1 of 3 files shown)


NobodyNada:
quartata [has added this to the blacklist](https://github.com/Charcoal-SE/SmokeDetector/blob/master/blacklisted_usernames.txt#L68), so we can either close this PR to keep it on the blacklist, or we can revert quartata's commits and merge this PR to make it a "pseudo-watchlist."

angussidney:
I guess we can leave this open and see how the blacklist goes, and revert to this method if necessary.

angussidney:
The current solution of blacklisting seems to be working fine; this PR is no longer needed. I'd suggest keeping the branch just in case we decide to downgrade the check in the future.

PR #1621, opened 2018-02-15
Update bad_keywords.txt

ByteCommander:
Improve catching the number in different formats
https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body=8711&site=120&commit=Search


tripleee:
This was superseded by a similar but different fix which as already published to production.

PR #1622, opened 2018-02-16
Remove bad blacklist (autopull)

iBug:
Word `copula` has a lot of legitimate uses, so I removed it. See the Stack Exchange search [here](https://stackexchange.com/search?q=copula).

3 other fixes (no need for `whatever male enhancement` because `male enhancement` is already caught).


tripleee:
!!/approve

PR #1623, opened 2018-02-16
Fix bad "offensive keyword"

iBug:
Basically just removed `ejaculate` which generated a lot of FPs. Some other tweaks.

Before looking at MS results, please note that most of the TPs are spam, instead of "offensive". They should not count for this keyword.


Undo1:
Looks decent. The true positives are either way above or already way below autoflaggin thresholds, and it's only 22/88 overall.

Undo1:
!!/approve

PR #1624, opened 2018-02-17
Deletionwatcher refactor

quartata:
don't merge this yet

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -261,7 +261,8 @@ def tell_rooms(msg, has, hasnt, notify_site="", report_data=None):
 
         if room.block_time < timestamp and _global_block < timestamp:
             if report_data and "delay" in _room_roles and room_id in _room_roles["delay"]:
-                callback = lambda: _msg_queue.put((room, msg_pings, report_data))
+                def callback():
+                    _msg_queue.put((room, msg_pings, report_data))
 
                 GlobalVars.deletion_watcher.subscribe(report_data[0], callback=callback, timeout=300)
             else:
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -170,7 +170,7 @@ def send_messages():
                         if room.deletion_watcher:
                             callback = room.room._client.get_message(message_id).delete
 
-                            GlobalVars.deletion_watcher.subscribe(report_data[0], callback=callback, max=120)
+                            GlobalVars.deletion_watcher.subscribe(report_data[0], callback=callback, timeout=120)
 
                     _pickle_run.set()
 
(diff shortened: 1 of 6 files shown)


ArtOfCode-:
10 second overview of how a post flows through this?

chatcommunicate.py
```diff
@@ -262,9 +262,9 @@ def tell_rooms(msg, has, hasnt, notify_site="", report_data=None):
 
         if room.block_time < timestamp and _global_block < timestamp:
             if report_data and "delay" in _room_roles and room_id in _room_roles["delay"]:
-                threading.Thread(name="delayed post",
-                                 target=DeletionWatcher.post_message_if_not_deleted,
-                                 args=(msg_pings, room, report_data)).start()
+                task = Tasks.later(_msg_queue.put, (room, msg_pings, report_data), after=300)
```
Undo1: Is `after=300` deliberate here? Seems like it should be derived from `room.block_time`?

deletionwatcher.py
```diff
@@ -17,80 +20,110 @@
...
+                                    if callback and (not max_time or time.time() < max_time):
+                                        callback()
+                        except KeyError:
+                            pass
+
+    def subscribe(self, post_url, callback=None, pickle=True, timeout=None):
+        post_id, post_site, post_type = fetch_post_id_and_site_from_url(post_url)
+
+        if post_site not in GlobalVars.site_id_dict:
+            return
```
Undo1: Should we do something special here, like post to chat or otherwise panic? I'm assuming this isn't meant to happen often.

chatcommunicate.py
```diff
@@ -262,9 +262,9 @@ def tell_rooms(msg, has, hasnt, notify_site="", report_data=None):
...
+
+                GlobalVars.deletion_watcher.subscribe(report_data[0], callback=task.cancel)
```
Undo1: Should this use `subscribe`'s `timeout` param derived from `room.block_time`? Calling `cancel` on the task is useless after the message has been posted.

Undo1:
Here's my understanding of the flow @ArtOfCode- - correct me if I'm wrong @quartata:

1. DeletionWatcher is a singleton held in GlobalVars. On init, it checks for a saved pickle of posts being watched before a reboot or crash and re-subscribes them.
2. When handle_spam is called, it calls `deletion_watcher.subscribe` 
3. When a report is a currently blocked room, `deletion_watcher.subscribe` is called again with the post info and a callback to cancelling a scheduled post task if the post is deleted. A room that removes deleted reports also subscribes each post to the watcher, this time with a callback to remove the message if the post is deleted.
4. The call the `subscribe` subscribes to the appropriate websocket connection, like we do now. It also writes the pickle file in case there's an unexpected restart.
5.  When a message comes down the websocket, it looks up the post and notifies metasmoke, then triggers any callbacks. 

Looks great to me aside from the `if room.block_time` block questions.

chatcommunicate.py
Undo1: Ignore this one too, I can't read.

chatcommunicate.py
Undo1: Nevermind, I can't read.

Undo1:
!!/approve - ready to go @quartata?

quartata:
@Undo1 probably

PR #1625, opened 2018-02-17
Add a list of keywords specially for Stack Overflow (really done)

iBug:
Since Stack Overflow gets special spams, I made an extra list of keywords for it.

RegEx keywords in this list will be checked against Stack Overflow posts only. They will not be checked on other sites. Note that "global" keywords still apply on SO: this list is an exclusive add-on.

The usage is more or less the same as the current keywords. People issue `!!/blacklist-sokeyword` to add keywords to this list.

~~~There's currently one functionality missing: `!!/test` command will not test this list, which may lead to some confusion.~~~ Just implemented it right now. Append `-so` to test commands to test a string as if it's on SO.

There are a few other improvements / code refactory that I've done by the way. Most of them are minor and do not affect how the code works.

---

You can still merge this request. In that case I'll create another for this one final TODO.

**TODO**: Remove keyword from SO blacklist when it is being blacklisted globally. This is unlikely to happen very often, as I can predict.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -201,19 +201,16 @@ def blacklist(_):
                        "and spaces and punctuations with `\\W?`.")
 
 
-def check_blacklist(string_to_test, **is_what):
+def check_blacklist(string_to_test, username=False, watch=False, site=""):
     # Test the string and provide a warning message if it is already caught.
     sample_post = {'title': 'Valid title', 'body': 'Valid body',
                    'owner': {'display_name': 'Valid username', 'reputation': 1, 'link': ''},
-                   'site': '', 'IsAnswer': False, 'score': 0}
+                   'site': site, 'IsAnswer': False, 'score': 0}
 
-    if is_what['username']:
+    if username:
         sample_post['owner']['display_name'] = string_to_test
     else:
         sample_post['body'] = string_to_test
-
-    if is_what['stackoverflow']:
-        sample_post['site'] = 'stackoverflow.com'
 
     sample_post['IsAnswer'] = False
     question = Post(api_response=sample_post)
@@ -228,7 +225,7 @@ def check_blacklist(string_to_test, **is_what):
     reasons = list(set(question_reasons) | set(answer_reasons))
 
     # Filter out watchlist results
-    if not is_what['watchlist']:
+    if not watch:
         reasons = list(filter(lambda reason: "potentially bad keyword" not in reason, reasons))
 
     return reasons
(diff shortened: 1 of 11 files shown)


chatcommands.py
```diff
@@ -194,29 +194,32 @@ def blacklist(_):
...
-                       "Remember to escape dots in URLs using \\.")
+    raise CmdException("The `!!/blacklist` command has been deprecated. "
+                       "Please use `!!/blacklist-website`, `!!/blacklist-username`, "
+                       "`!!/blacklist-keyword`, or perhaps `!!/watch`. "
+                       "Remember to escape dots in URLs using `\\.`, "
+                       "and spaces and punctuations with `\\W?`.")
 
 
-def check_blacklist(string_to_test, is_username, is_watchlist):
+def check_blacklist(string_to_test, **is_what):
```
j-f1: `**is_a`?

iBug: You really care about naming convention? I think the current name `is_what` is just fine. Anyway it's trivial to change that.

quartata:
I just went ahead and made the changes, but for future reference:

1. kwarg splats really only have two legitimate uses: either you don't know how many or what the names of your kwargs are, or you need to pass the entire dictionary off to something else (like to encode in JSON perhaps.) It should definitely not be used for only 3 kwargs cause it's slow.

2. Metasmoke identifies reasons by their name: even if you have two rules that aren't mutually exclusive, if you name them both "bad keyword in {}" they're considered the same. So really you should have the regular blacklists on everywhere and the SO blacklist only on for SO, rather than turning off the regular blacklist for SO and repeating the massive regex.

Undo1:
I don't understand why we need a separate SO keyword list. This is definitely a good capability to have, but the current application doesn't make sense to me. I suspect I'm missing some context here (beyond https://meta.stackoverflow.com/questions/363423/how-is-this-question-spam-or-r-a and https://chat.stackexchange.com/transcript/11540?m=42911327#42911327), but it seems like this would be a fine addition to the global blacklist.

What am I missing?

AWegnerGitHub:
I'm inclined to agree with Undo. This seems...odd, without a but more
context. Especially because when I first saw this it has long path tool in
the SO keywords. That's not SO specific

angussidney:
In the past, when we've had to blacklist items for a single site only, we've just added a separate regex check that feeds into the 'Bad keyword' reason and is limited to the site we want ([example](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L896)).

I really don't see the need for an entirely new blacklist file with related functionality (commands, git integration, etc), especially when the only item on such list would be appropriate on the main blacklist. This change will only unnecessarily increase code complexity and slow down the bot (marginally)

angussidney:
Closing since the only keyword on this list has already been blacklisted, and general consensus is that we don't really need a new blacklist.

PR #1629, opened 2018-02-19
Catch "India Tour"

iBug:
A new list that could be populated just like "company|facility|etc in india".


iBug:
That CircleCI failure is not my fault!

tripleee:
I am unable to repro the test failure locally, but I also don't seem to be authorized to trigger a retest.

tripleee:
!!/approve

(Undo triggered a rebuild, which cleared the CircleCI failure.)

PR #1630, opened 2018-02-19
Stop catching i.stack.imgur.com

iBug:
"Misleading link" linking to `i.stack.imgur.com` is too falsey.

See [the MS search](https://metasmoke.erwaysoftware.com/search?body=href%3D%22https%3A%2F%2Fi.stack.imgur.com&reason=139&commit=Search).

+Another minor tweak.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -35,13 +35,14 @@ COMMON_MALFORMED_PROTOCOLS = [
 SAFE_EXTENSIONS = set(('htm', 'py', 'java', 'sh'))
 SE_SITES_RE = r'(?:{sites})'.format(
     sites='|'.join([
-        r'([a-z]+\.)stackoverflow\.com',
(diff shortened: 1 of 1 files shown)


NobodyNada:
FWIW, the cases that resulted in the original misleading link filter were a user pretending to link to documentation/JSFiddle, but the links actually went to porn on stack.imgur.

I’ve thought about whitelisting imgur if the original link looks like a photo (maybe if the URL contains `photo`, `png, etc?).  We could also try sending a request to the URL in the text and checking if it’s a photo.

iBug:
@NobodyNada We're going to hit *a lot* of documentations and code segments and GitHub repos if we try to determine the URL in link text.

NobodyNada:
@iBug But how many of those 1) link to imgur and 2) already trip the misleading link filter?

iBug:
@NobodyNada A few every day.

PR #1634, opened 2018-02-20
Add !!/checkpost command (mostly complete)

iBug:
So in case Smokey misses a post, any privileged user can force it to scan it.

One can also use this command to scan a potential spam without `!!/report`ing it.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -13,7 +13,7 @@ from datahandling import *
 from blacklists import load_blacklists
 from metasmoke import Metasmoke
 from parsing import *
-from spamhandling import handle_spam
+from spamhandling import check_if_spam, handle_spam
 from gitmanager import GitManager
 from tasks import Tasks
 import threading
@@ -1099,27 +1099,40 @@ def checkpost(msg, url):  # FIXME: Currently does not support batch report
     Force Smokey to scan a post even if it has no recent activity
     :param msg:
     :param url:
-    :return I don't know:
+    :return str:
     """
-
     crn, wait = can_report_now(msg.owner.id, msg._client.host)
     if not crn:
-        raise CmdException("You should totally drop that and try jQuery.")
+        raise CmdException("You can execute the !!/checkpost command again in {} seconds. "
+                           "To avoid one user sending lots of reports in a few commands and "
+                           "slowing SmokeDetector down due to rate-limiting, you have to "
+                           "wait 30 seconds after you've reported multiple posts in "
+                           "one go.".format(wait))
 
     post_data = api_get_post(url)
 
     if post_data is None:
-        raise CmdException("That does not look like a valid post URL")
+        raise CmdException("That does not look like a valid post URL.")
 
     if post_data is False:
         raise CmdException("Cannot find data for this post in the API. "
                            "It may have already been deleted.")
 
-    post = Post(api_response=post_data.as_dict())
+    post = Post(api_response=post_data.as_dict)
+
+    if fetch_post_id_and_site_from_url(url)[2] == "answer":
+        fake_parent = Post(api_response={})
+        fake_parent._post_id = post_data.question_id
+        fake_parent._title = post_data.title
+        fake_parent._answers = []
+
+        post._is_answer = True
+        post._parent = fake_parent
+
     is_spam, reasons, why = check_if_spam(post)
 
     if is_spam:
-        spamhandling.handle_spam(post=post, reasons=reasons, why=why + ["Manually triggered scan"])
+        handle_spam(post=post, reasons=reasons, why=why + "\nManually triggered scan")
         return None
 
     # Temporarily just use the URL provided from the command. That should change someday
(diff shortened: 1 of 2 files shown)


WELZ-gh:
Maybe we should call it `!!/scan` instead. (Since that's all it's doing, and it's also shorter)

paper1111:
@WELZ-gh we can add an alias to the original function so both variants can be used

chatcommands.py
```diff
@@ -376,7 +376,7 @@ def brownie():
     return "Brown!"
 
 
-COFFEES = ['Espresso', 'Macchiato', 'Ristretto', 'Americano', 'Latte', 'Cappuccino', 'Mocha', 'Affogato']
+COFFEES = ['Espresso', 'Macchiato', 'Ristretto', 'Americano', 'Latte', 'Cappuccino', 'Mocha', 'Affogato', 'jQuery']
```
angussidney: Why do you keep adding extra teas and coffees?

chatcommands.py
```diff
@@ -1092,6 +1092,47 @@ def report(msg, urls):
...
+@command(str, whole_msg=True, privileged=True, aliases=['scan', 'test-p'])
+def checkpost(msg, url):  # FIXME: Currently does not support batch report
+    """
+    Force Smokey to scan a post even if it has no recent activity
+    :param msg:
+    :param url:
+    :return I don't know:
+    """
+
+    crn, wait = can_report_now(msg.owner.id, msg._client.host)
```
paper1111: `can_report_now` shouldn't be called as this command doesn't call the API, it doesn't use API quota and won't be rate limited.

iBug:
@ArtOfCode- The queuing code is copied from `ws.py` where the most is processed. We probably don't need to wrap it up in a separate function.

@angussidney Just a fan of jokes and [the jQuery meme](https://meta.stackexchange.com/a/19492/350567).

chatcommands.py
angussidney: > Just a fan of jokes and the jQuery meme.

I'm sorry, but the jokes aren't needed. Especially when such joke involves inserting a programming framework into a list of teas. We've decided in the past that we aren't going to add/modify existing joke commands.

iBug:
I run into trouble trying to make an intercepting API call to fetch post (so the response should be immediate). I can't figure out how to add the post to queue, either. Any help?

iBug:
@ArtOfCode-  There's some differences between the two parts so I cannot simply extract them and wrap them up in one function.

@angussidney Never knew that. Removed the change anyway.

@paper1111  Dropped that for now but I think it should be around some time.

chatcommands.py
quartata: This does call the API, albeit indirectly.

chatcommands.py
```diff
@@ -1092,6 +1092,48 @@ def report(msg, urls):
...
+    post_data = api_get_post(url)
+
+    if post_data is None:
+        raise CmdException("That does not look like a valid post URL")
+
+    if post_data is False:
+        raise CmdException("Cannot find data for this post in the API. "
+                           "It may have already been deleted.")
+
+    is_spam, reasons, why = check_if_spam(title=post_data.title,
```
quartata: `check_if_spam` doesn't work this way. You need a Post (not a PostData either). The best way to handle this is to do `check_if_spam(Post(api_response=post_data.as_dict()))`

However, I'm not sure why you're fetching the post separately, scanning it, then enqueueing it to fetch and scan again. If you want to immediately report the post if it's spam, just call into `spamhandling.handle_spam`

quartata:
Git history on this one has become a bit of a bother

quartata:
@ArtOfCode- most of it already is abstracted into `apigetpost` (although that needs a thorough spring cleaning)

iBug:
@quartata I'm not quite familiar with the post fetching logics so I just copied code from `ws.py`. If you know what the functions are doing then you can just edit the code. Thanks in advance.

chatcommands.py
iBug: Very useful information. I'm doing one last modification today before going to sleep. Will change in a few minutes.

quartata:
@iBug I don't know what you're trying to do though. Do you want it to be scanned immediately or enqueued and scanned later as part of a batch? (You're doing both right now.)

iBug:
@quartata Immediately is preferred, but I don't know the exact logic so I just copy-paste code from other parts which messed it up.

quartata:
I've fixed all of the problems, however this has a critical flaw in that it does not fetch the parent question (and doing so would consume additional API quota).

iBug:
@quartata Thanks for the fix. If it looks good can we give it some trial runs?

@ArtOfCode- It looks complete now.

ArtOfCode-:
@quartata any particular reason other than API quota we can't fetch the parent? We've got quota to burn, so that's not an issue. If we're checking an answer we may as well check the rest of the page as well.

quartata:
@ArtOfCode- no, but it's probably more effort than it's worth

iBug:
@ArtOfCode- Agreeing with @quartata , we aren't getting anything other than the title for an extra quota usage (and coding effort), given how this command is intented to work.

quartata:
actually I just finished writing it locally

quartata:
!!/approve

PR #1635, opened 2018-02-20
Ignore if Community user bumped post [INCOMPLETE]

### LINKED ISSUE #1085: Avoid catching the same post if it hasn't been modified since last report
On several occasions one sees the disambiguation screen on MS because there are multiple reports for a single posts. But many times the contents of the post are exactly the same since the last report. 

SmokeDetector can simply filter our such cases by comparing the last reported time with the last modified time of the post. Even better if it can do an exact string matching from the last report (to avoid change + rollbacks changing the modified time).

One instance of such a case is [here](https://metasmoke.erwaysoftware.com/posts/by-url?url=//dba.stackexchange.com/a/131273). 

The answer has never been edited and also has been marked as fp or naa on all the occasions. 
It was caught just because Community boosted the post.

### LINKED ISSUE #1552: Same post caught multiple times due to Community ♦ bumping
See [this example][1] It has 5 records on the same post. And then see the [revision history][2] of that very post, you'll find how it's caught for 5 times throughout 5 months.

The first time is when it's initially posted, and the second time is after an edit from a moderator. The 3rd, 4th and 5th time it's caught without anything changed is because it's **bumped to front by the Community ♦ user**. In this way a post could be caught repeatedly, generating a lot of FPs (spams are deleted and won't be caught for a 2nd time, so this won't generate extra TP).

So I think this is a bug that should be fixed. We should ignore a post if its latest activity is "bumped by Community ♦" as there's nothing new to be scanned again.

  [1]: https://m.erwaysoftware.com/posts/by-url?url=//raspberrypi.stackexchange.com/questions/71236
  [2]: https://raspberrypi.stackexchange.com/posts/71236/revisions

paper1111:
Implements the functionality requested in #1552. However, as the date where the post is recorded into MS is currently not checked, if a post bumped by Community has spam edited in later, it will slip through the checks, though this chance should be slim.

Fixes #1085

### DIFF
--- a/datahandling.py
+++ b/datahandling.py
@@ -439,8 +439,9 @@ def can_report_now(user_id, chat_host):
 
 
 def has_community_bumped_post(post_url, post_content):
-    ms_posts = Metasmoke.get_post_bodies_from_ms(post_url)
-    for post in ms_posts:
-        if post['body'] == post_content:
-            return True
+    if GlobalVars.metasmoke_key is not None and GlobalVars.metasmoke_host is not None:
+        ms_posts = Metasmoke.get_post_bodies_from_ms(post_url)
+        for post in ms_posts:
+            if post['body'] == post_content:
+                return True
     return False


paper1111:
Can I ignore codeclimate?

ArtOfCode-:
CodeClimate overridden

ArtOfCode-:
Looks good to me so far. What's still incomplete?

paper1111:
@ArtOfCode- "the date where the post is recorded into MS is currently not checked, if a post bumped by Community has spam edited in later, it will slip through the checks, though this chance should be slim."

ArtOfCode-:
Ah, gotcha. Yeah, that shouldn't be hard - filter for `posts.body` and `posts.created_at` instead of just body, and compare dates.

quartata:
I personally think this is something that should be left for the `ignore` command. Just because something was bumped by Community doesn't necessarily mean we should scan it again -- it's entirely possible that we hadn't seen the given revision before until it was bumped by Community,

paper1111:
@quartata regarding "it's entirely possible that we hadn't seen the given revision before until it was bumped by Community"

In this PR the code checks for the existence of the same text with the same url in MS, so if the post has not been scanned by Smokey before and was bumped by Community, it will still be scanned.

paper1111:
Superseded by #2298 because I deleted my old Smokey fork

PR #1648, opened 2018-02-23
Remove sites.google.com

iBug:
Too many FPs. See the MS search [here](https://metasmoke.erwaysoftware.com/search?body=sites.google.com%2Fsite)

[This](https://metasmoke.erwaysoftware.com/post/106388) is the only TP it has generated so far. All other TPs are also caught for other more accurate reasons/keywords.


teward:
!!/approve

PR #1649, opened 2018-02-23
HyperNeutrino: Blacklist slut

SmokeDetector:
[HyperNeutrino](http://chat.stackexchange.com/users/281362) requests the blacklist of the keyword `slut`. See the Metasmoke search [here](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=slut) and the Stack Exchange search [here](https://stackexchange.com/search?q=%22slut%22).

PR #1650, opened 2018-02-23
add "slut" to offensive regex

alexander-liao:
"slut" with any number of "t"s, optionally followed by a "y"

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -442,7 +442,7 @@ def is_offensive_post(s, site, *args):
     if s is None or len(s) == 0:
         return False, ""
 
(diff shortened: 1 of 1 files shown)


teward:
This is effectively a duplicate of #1649

teward:
While this is a good idea, I want to be sure we don't catch any odd cases of false positives.  It's far easier to update the blacklists than it is to adjust the regex here, so for *now* I'm rejecting this suggestion here.

angussidney:
The whole reason why we have an offensive regex is for offensive words. Adding this to the blacklist doesn't make it any easier to edit/remove - either way it still requires someone with git access.

angussidney:
!!/approve

Issue #513, opened 2017-02-07
Manually reported question: Show which rules would have hit


tripleee:
When a post has been manually reported, it's hard manual work to figure out if it was reported because Smoke Detector missed it (which is often a problem which we want to spend time on solving) or just because the reporter was quicker than Smokey (which of course isn't a problem at all).  Could the analysis be run on these questions after they've been reported, and included in the Metasmoke record, so that it's easier to see when we have missed something?

tripleee:
Really a feature request for Metasmoke actually, I guess.  Is there a way to move it?

Undo1:
I'd say it should stay here, actually - SmokeDetector has all the parsing code; this is something that would be implemented here, not in metasmoke.

ByteCommander:
I'd like to give this issue a little *bump*...

superplane39:
What would happen with the weight of the !!/report command with this? Could this theoretically result in manually reported post being autoflagged?

iBug:
https://chat.stackexchange.com/transcript/message/42954284#42954284

Maybe implement a `!!/scan <URL>` command to force a scan on a given post, and nothing should happen if the post is all right.

iBug:
#1634 is a good solution for this and it does all the job I wanted: Scan the post and report it, and if the post is all right then do nothing (it does reply *Post does not look like spam* as a clearer indication). It's rolled out and (hopefulley) good to use now.

PR #1677, opened 2018-03-05
Show reasons that would have been caught for !!/report posts

iBug:
Finally, canonically, ultimately,

A solution to #513


angussidney:
Is there any reason why we can't just merge !!/scan and !!/report into one command with different aliases?

iBug:
@angussidney Yes. They are still very different from each other. `!!/report` reports a post no matter what it is. There are also differences in tests, so we can't just simply merge them into one, at least not by me (maybe someone else with better understanding of the code can).

Undo1:
!!/approve

PR #1698, opened 2018-03-09
Adding Princess Luna to CHQ

magisch:
- Known from the Tavern and SOBotics
- 6,5k rep on SO
- Generally regarded as sane

Requested here: https://chat.stackexchange.com/transcript/message/43315531#43315531

Approve or deny at leisure


Undo1:
!!/approve

PR #1715, opened 2018-03-12
Add charcoal Test to rooms.yml

magisch:
At request of Glorfindel


tripleee:
Strictly speaking, feedback from the test room should not go into the production MS. Perhaps we still want to merge this now, but keep that in mind and attempt to fix it separately...?

magisch:
@tripleee https://github.com/Charcoal-SE/SmokeDetector/issues/1717

PR #1721, opened 2018-03-12
fix crash on targeted blame

micsthepick:
Blame the user of a targeted blame if they provide an invalid user id.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -363,7 +363,7 @@ def blame2(msg, x):
 
     try:
         unlucky_victim = msg._client.get_user(user)
-    except  HTTPError:
+    except HTTPError:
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -361,7 +361,10 @@ def blame2(msg, x):
     for i, char in enumerate(reversed(x)):
         user += (len(base)**i) * base[char]
 
-    unlucky_victim = msg._client.get_user(user)
+    try:
+        unlucky_victim = msg._client.get_user(user)
+    except HTTPError:
+        unlucky_victim = msg._client.get_user(random.choice(msg.room.get_current_user_ids()))
```
ArtOfCode-: How about making it more amusing and picking the person who invoked the command instead?

micsthepick: Sounds like a good idea

Issue #1428, opened 2018-01-01
Incorrect parsing of answer URL as question


Undo1:
Bhargav reported [this post](https://stackoverflow.com/questions/20038353/smooth-animation-when-tab-switches/48046679#48046679) in [this chat message](https://chat.stackoverflow.com/transcript/message/40629714#40629714):

```
 !!/report https://stackoverflow.com/questions/20038353/smooth-animation-when-tab-switches/48046679#48046679
```

The link was somehow parsed as a question link, resulting in this URL in [the report](https://chat.stackoverflow.com/transcript/message/40629715#40629715) ([MS](https://metasmoke.erwaysoftware.com/post/100010)):

```
https://stackoverflow.com/questions/20038353
```

quartata:
This has existed for as long as I've been aware of. You need to use share to get an answer link.

Undo1:
I could have sworn it used to work. If we can't fix it to work work those links (shouldn't be too hard), we should at least reject those format links.

teward:
we DID have it parse answers properly, before the recent refactoring if I remember right.  But it might require some specific debugging tracking to ID the actual culprit in the code.



*Sent from my iPhone.  Please excuse any typos, as they are likely to happen by accident.*

j-f1:
Check out https://github.com/Charcoal-SE/charcoal-se.github.io/blob/site/_assets/js/ms-urls.coffee for the RegExii I use on the ms-urls page to find the question/answer IDs.

Bhargav-Rao:
FWIW, I manually reported another answer, a couple of hours before that [here](https://chat.stackoverflow.com/transcript/message/40629414#40629414), and it worked perfectly. Text of the report: 

    !!/report https://stackoverflow.com/questions/239209/is-it-safe-to-assume-that-the-path-c-windows-system32-always-exists/48046326#48046326

and metasmoke page https://metasmoke.erwaysoftware.com/post/100008

Undo1:
Drat, I was hoping for a quick fix. No such luck:

```
>>> print(parsing.fetch_post_id_and_site_from_url("https://stackoverflow.com/questions/20038353/smooth-animation-when-tab-switches/48046679#48046679"))
('48046679', 'stackoverflow.com', 'answer')
[...]

PR #1723, opened 2018-03-13
Attempt to fix the bad URL parsing in !!/report's

iBug:
#1554 seems to have failed because it's wrong. Smokey receives the actual invisible characters instead of the HTML-encoded ones. See tests starting from [this message](https://chat.stackexchange.com/transcript/message/43369764#43369764).

[Most recent occurrence](https://chat.stackoverflow.com/transcript/message/41585835#41585835) by Bhargav Rao

[The analysis of the message](https://chat.stackexchange.com/transcript/message/43368641#43368641#

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1070,6 +1070,7 @@ def report(msg, urls):
                            "which would slow down reports.")
 
     for index, url in enumerate(urls, start=1):
+        url = rebuild_url(url)
         post_data = api_get_post(url)
 
         if post_data is None:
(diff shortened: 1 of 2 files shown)


tripleee:
Is this related to a known issue? Can you add a pointer? Perhaps a failed `!!/report` in chat if there isn't an issue on GitHub.

tripleee:
So this is related to https://github.com/Charcoal-SE/SmokeDetector/issues/1428

tripleee:
This now looks unintrusive and correct.  I'm still wondering if the code should do the same thing everywhere (look for the sequence `\u200c\u200b` precisely, or replace both characters separately in any context; but not do different things in different places).

!!/approve

PR #1731, opened 2018-03-13
Update blacklisted_websites.txt

WELZ-gh:
removed www (which can make it not catch as blacklisted site)


WELZ-gh:
!!/approve

iBug:
You'd better check the CI build log and see what went wrong.

PR #1732, opened 2018-03-14
Cleanup setup scripts

iBug:


quartata:
The setup.sh script is for when you're first creating an instance. install_dependencies.sh is useful for updating dependencies on existing instances.

quartata:
(although the pip install for the user_requirements is missing from setup.sh....)

PR #1743, opened 2018-03-15
Update README.md

iBug:
We're no longer using ChatExchange as a submodule, right?

### DIFF
--- a/setp.sh
+++ b/setp.sh
@@ -1,0 +1,4 @@
+# This file will not be tracked by git as it's declared in .gitignore
+
+export ChatExchangeU=YourEmailHere
+export ChatExchangeP=YourPasswordHere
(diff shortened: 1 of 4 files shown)


angussidney:
'Credidential' should be 'credidential'

Also bonus marks if you can add a copy of the old setp.sh file to the repo so that we don't need to add our own 👍

iBug:
@angussidney Please don't taunt me for my high typo rate (I code on mobile)... and yeah, I'll add an example file which should not be tracked by Git after modification.

setp.sh
```diff
@@ -0,0 +1,4 @@
+# This file will not be tracked by git as it's declared in .gitignore
```
tripleee: This is obviously not true.

iBug: First, I use GH web editor. Second, there's `git add --force` that force stages a file even if it's in .gitignore.

PR #1748, opened 2018-03-15
fix remove_all_notifications

micsthepick:
in globalvars, notifications have an extra value (True) added to them.
relevant line of code:  `GlobalVars.notifications[i] = notification + (True,)`
I fixed this file to discard that extra value when calling `remove_from_notification_list` as this causes an error (too many arguments).

### DIFF
--- a/datahandling.py
+++ b/datahandling.py
@@ -383,7 +383,7 @@ def remove_all_from_notification_list(user_id):
             my_notifications.append(notification)
 
     for notification in my_notifications:
-        remove_from_notification_list(*notification)
+        remove_from_notification_list(*(notification[:4]))
 
 
 def get_all_notification_sites(user_id, chat_site, room_id):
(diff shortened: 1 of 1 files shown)


iBug:
I think it's better if you modify that function and catch the extra argument, instead of dropping it.

micsthepick:
@iBug should we do anything with it? What is it even for?
also, we would need to either give it a default or modify other calls to it.

iBug:
@micsthepick Just give it a default value and any name you deem reasonable. It won't hurt.

tripleee:
Liking this so far, but waiting for things to settle.

micsthepick:
so how do I tell codeclimate that the function can now accept an optional 5th argument?

tripleee:
Apparently it hardcodes 4 arguments max: https://docs.codeclimate.com/docs/default-analysis-configuration

micsthepick:
should we just go with plan A then?

tripleee:
The previous consensus was apparently to simply ignore codeclimate.

With that, fine by me.

!!/approve

PR #1764, opened 2018-03-17
Don't multi-scan a post

iBug:
Duplicate reports can be easily generated by `!!/scan`ning a post repeatedly. This patch prevents that incident.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1162,6 +1162,18 @@ def checkpost(msg, url, alias_used='scan'):  # FIXME: Currently does not support
     url = to_protocol_relative(post_data.post_url)
     post = Post(api_response=post_data.as_dict)
 
+    if has_already_been_posted(post_data.site, post_data.post_id, post_data.title) \
+        and not is_false_positive((post_data.post_id, post_data.site)):
+        # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
+        # this force scan might be attempting to correct that/fix a mistake/etc.
+
+        response_text = "This post is already recently reported"
+        if GlobalVars.metasmoke_key is not None:
+            ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(url)  # se_link == url
+            return response_text + " [ [MS]({}) ]".format(ms_link)
+        else:
+            return response_text + "."
+
     if fetch_post_id_and_site_from_url(url)[2] == "answer":
         parent = api_get_post("https://{}/q/{}".format(post.post_site, post_data.question_id))
 


tripleee:
!!/approve

chatcommands.py
```diff
@@ -1162,6 +1162,18 @@ def checkpost(msg, url, alias_used='scan'):  # FIXME: Currently does not support
...
 
+    if has_already_been_posted(post_data.site, post_data.post_id, post_data.title) \
+            and not is_false_positive((post_data.post_id, post_data.site)):
+        # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
+        # this force scan might be attempting to correct that/fix a mistake/etc.
+
+        response_text = "This post is already recently reported"
+        if GlobalVars.metasmoke_key is not None:
+            ms_link = "https://m.erwaysoftware.com/posts/by-url?url={}".format(url)  # se_link == url
+            return response_text + " [ [MS]({}) ]".format(ms_link)
```
angussidney: These should be `raise CmdException("text.....")`

chatcommands.py
```diff
@@ -1162,6 +1162,18 @@ def checkpost(msg, url, alias_used='scan'):  # FIXME: Currently does not support
     url = to_protocol_relative(post_data.post_url)
     post = Post(api_response=post_data.as_dict)
 
+    if has_already_been_posted(post_data.site, post_data.post_id, post_data.title) \
+            and not is_false_positive((post_data.post_id, post_data.site)):
+        # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
+        # this force scan might be attempting to correct that/fix a mistake/etc.
+
+        response_text = "This post is already recently reported"
```
angussidney: Since this is the `!!/scan` command rather than the `!!/report` command, this should probably be 'This post has been recently scanned'

iBug: I noticed this and deliberately kept the word *reported*, as it really *is* reported. Scanning a post won't prevent you from scanning again, but a report does. So I'd like to not change it.

chatcommands.py
iBug: I didn't understand that. Why should it be an "exception" like invalid URL or unable to find post?

angussidney: Whenever the user provides an ‘invalid’ input (which includes scanning a recently reported post), or the command throws an error, it’s more semantically correct to raise a command exception. It’s similar to our old `status=false` argument for the old chat system.

quartata: "raising a CmdException will always post the response, even if the command is silent (with a -). So it's best used for errors where the invoker needs to know what they did wrong."

iBug: @angussidney Thanks to quartata and I changed it.

PR #1766, opened 2018-03-18
Cleanup missed fixes from #1723

iBug:
The URL parsing bug should be ultimately fixed. Some old (failed) attempts are still left in the code. This PR cleans them all up.


tripleee:
!!/approve

Issue #1735, opened 2018-03-14
Throttle less important chat messages


Glorfindel83:
Context: [here](https://chat.stackexchange.com/transcript/message/43396899#43396899).
Smokey is busy posting unimportant messages like

> Conflicting feedback across revisions: current, #1
> Conflicting feedback across revisions: current, #1
> [...]
> Conflicting feedback across revisions: current, #1
> Conflicting feedback across revisions: current, #1
> Conflicting feedback across revisions: current, #1

while there's a report waiting, which took more than a minute of waiting to be posted. In the end it wasn't posted at all due to a reboot.

iBug:
I'll take a look at this in a few days and hopefully come up with a fix, after coping with my homeworks and assignments.

PR #1780, opened 2018-03-21
Throttle less important messages (incomplete)

iBug:
This is an implementation of feature request #1735.

I plan to do this by adding another "throttled message queue" (TMQ). Messages in TMQ will be extracted and squashed into one when necessary. Debug messages and MS messages will go to this queue by default. Usually, this should have little to no effect on Smokey's behavior, but when there's a lot of debug messages, it will make a difference.

I'll be working on this FR/PR in the next few days, and drop the *incomplete* notice from the title after I think I've done.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -50,6 +50,8 @@ _global_block = -1
 _rooms = {}
 _last_messages = LastMessages({}, collections.OrderedDict())
 _msg_queue = queue.Queue()
+_th_msg_queue = queue.Queue()  # Throttled message queue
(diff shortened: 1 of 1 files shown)


angussidney:
FYI, markdown doesn't render in multiline chat messages, which may mean that collapsing the messages into one isn't the right solution.

iBug:
@angussidney The what about using things like a slash as a separator instead of newlines?

NobodyNada:
@iBug If we collapse them into a single-line message, we’ll run into the message length limit pretty quickly.

iBug:
I'll restart this if it's later found needed,

PR #1785, opened 2018-03-21
Add welcome command

CalvT:
A command that we can use to point new users to that explains briefly what Smokey is, without pointing to a list of commands.


chatcommands.py
```diff
@@ -546,6 +546,19 @@ def info():
...
+@command()
+def welcome(msg):
+    """
+    Returns the welcome text
+    :return: A string
+    """
+    return "Welcome to " + msg.room.name + "! I'm " + GlobalVars.chatmessage_prefix +\
+           ", a bot that detects spam and offensive posts on the network and"\
+           " posts alerts to chat."\
+           " You can find more about me on the [Charcoal Website](https://charcoal-se.org/)."
```
ArtOfCode-: Website doesn't need to be capitalized

PR #1786, opened 2018-03-21
Add params to welcome, add other user

CalvT:
Fix current error, and add possibility to ping other user with the welcome command

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -548,15 +548,17 @@ def info():
 
 # noinspection PyIncorrectDocstring
 @command()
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -548,15 +548,21 @@ def info():
 
 # noinspection PyIncorrectDocstring
 @command()
```
ArtOfCode-: You're [going to need](https://stackoverflow.com/c/charcoal/a/59/1) `whole_msg=True` in the decorator here.

PR #1787, opened 2018-03-21
Sanitise welcome input

CalvT:
Hopefully sanitise the welcome other user input - so it removes the @, and only accepts the first word.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -555,7 +555,7 @@ def welcome(msg, other_user):
     :param other_user:
     :return: A string
     """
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -555,6 +555,7 @@ def welcome(msg, other_user):
     :param other_user:
     :return: A string
     """
+    other_user = regex.match(r'[^@\s]{3,}', other_user)
```
ArtOfCode-: Run this in a Python terminal, find out what you get. I'm getting `other_user` is always `None`.

PR #1804, opened 2018-03-22
Detect l and | as substitutes for 1

normalhuman:
In phone numbers, these may be used for obfuscation.  Example: https://metasmoke.erwaysoftware.com/post/111204


iBug:
!!/approve

PR #1811, opened 2018-03-23
Remove turkey/turkisk

CalvT:
Troll has been blocked by CM


honnza:
!!/approve

teward:
This is actually supposed to be declined, because this probably isn't over yet...

PR #1813, opened 2018-03-24
Update watched_keywords.txt

K-Davis1:
website wasn't escaped...

This can cause issues if the site is blacklisted (if it hasn't been already)


angussidney:
!!/approve

PR #1816, opened 2018-03-24
remove blacklist if fp from MS (incomplete)

paper1111:
Currently only fp feedback from chat will remove the user from the blacklist. With this PR, if someone marks a post fp in MS (via the API or review), the user will be removed from the blacklist too.


ArtOfCode-:
This doesn't appear to add a response to the new `unblacklist` message type that you're adding in Charcoal-SE/metasmoke#355.

paper1111:
@ArtOfCode- that's why it's "incomplete"

PR #1915, opened 2018-03-26
K.Davis: Watch enhances \W(?:fe)?male\Wfertility

K-Davis1:
Probably should also watch `enhance` as well

Should also catch:

enhance (fe)male fertility


tripleee:
!!/approve

PR #1931, opened 2018-03-29
Add birthday command

CalvT:
Add birthday command - basically to make Smokey wish Art happy birthday, but I'm sure it'll have future applications :)


AWegnerGitHub:
No. We don't need more commands that aren't applicable to Smokey's core mission.

PR #1933, opened 2018-03-30
Fix double credential prompt

iBug:
See [this post](https://stackoverflow.com/c/charcoal/questions/285) in Charcoal (SO Teams).

### DIFF
--- a/nocrash.py
+++ b/nocrash.py
@@ -31,8 +31,6 @@ ChatExchangeP = os.environ.get('ChatExchangeP')
 if ChatExchangeU is None:
     ChatExchangeU = str(input("Username: ")).strip('\r\n')
 
-os.environ['CEU'] = "h"
-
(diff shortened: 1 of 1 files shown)


nocrash.py
```diff
@@ -31,8 +31,6 @@
 if ChatExchangeU is None:
     ChatExchangeU = str(input("Username: ")).strip('\r\n')
 
-os.environ['CEU'] = "h"
-
 if ChatExchangeP is None:
     ChatExchangeP = str(input("Password: ")).strip('\r\n')
```
angussidney: Could you change this so that it uses `getpass` like [this](https://github.com/iBug/SmokeDetector/blob/a8598353fb78b5db98065e182d6e16d825b523a1/ws.py#L72) line? That way the password won't show up in plain text.

PR #1802, opened 2018-03-22
Properly blacklist f**k in usernames

CalvT:
Replaces https://github.com/Charcoal-SE/SmokeDetector/pull/1801

PR #1938, opened 2018-03-30
Update blacklisted_usernames.txt

K-Davis1:
We are getting a few fp's on this line. I think we would have much better success if it was the beginning of the word.


K-Davis1:
250+ tps, 19 fps.

Not too terribly convinced this is needed at this time. Maybe we can revisit this if it becomes more of an issue than a benefit.

NobodyNada:
Note that the problematic case has only been on the blacklist [for 8 days](https://github.com/Charcoal-SE/SmokeDetector/pull/1802)...

PR #1940, opened 2018-03-31
Update blacklisted_usernames.txt

K-Davis1:
Inspired by this [MS post](https://metasmoke.erwaysoftware.com/post/112622) that was not caught.


quartata:
this would match on, say, SmokeDetector-NG

quartata:
since it only needs a word boundary on the left and and the i and r are optional

K-Davis1:
@quartata good point. Should we just blacklist `nggr` specifically? Should we add to watch first?

PR #1960, opened 2018-04-05
!!/standby-except command

iBug:
Standby all instances except matching ones

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -746,15 +746,20 @@ def stopflagging():
 
 
 # noinspection PyIncorrectDocstring,PyProtectedMember
-@command(str, whole_msg=True, privileged=True)
-def standby(msg, location_search):
+@command(str, whole_msg=True, privileged=True, aliases=["standby-except"], give_name=True)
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -746,15 +746,18 @@ def stopflagging():
...
     Forces a system exit with exit code = 7
     :param msg:
     :param location_search:
     :return: None
     """
-    if location_search.lower() in GlobalVars.location.lower():
+    reverse_search = "except" in alias_used
+
+    # Caret `^` is xor
+    if location_search.lower() in GlobalVars.location.lower() ^ reverse_search:
```
ArtOfCode-: What's the logic precedence here? `(location_search.lower() in GlobalVars.location.lower()) ^ reverse_search`, or `location_search.lower() in (GlobalVars.location.lower() ^ reverse_search)`?

iBug: Meh, I forgo that Python chains logical operators arbitrarily, so it's actually `(A in B) and (B in C)`.

angussidney:
!!/approve

PR #1967, opened 2018-04-06
Revert "Restrict new user messages"

NobodyNada:
No longer needed, now that MS allows us to silence these messages.


K-Davis1:
It wasn't about seeing them, it was about rate-limiting Smokey.

ArtOfCode-:
!!/approve

Issue #1965, opened 2018-04-06
Allow users who manually report posts to specify an optional reason to be posted in Smokey's message


gparyani:
I'd like to propose a new feature for SmokeDetector, to allow users who file manual reports to specify an optional reason when doing so. Manual reports are usually filed because the post is still spam, but evaded Smokey's spam heuristics. Occasionally, it may also not be obvious to other humans that it's in fact spam. For example, [this report](https://chat.stackexchange.com/transcript/message/43834039#43834039) for a spam seed, which other users giving feedback in other chat rooms may not know is a spam seed unless they bothered to go into Charcoal HQ and look it up.

Implementing this would allow the reporter to supply a reason indicating that the post is a spam seed, which would be shown in the metasmoke page for the manual report, and in Smokey's messages in all chat rooms.

In my mind, this would look like this:

    !!/report [link] "[reason]" [link] "[reason]"...

An example report would be:

    !!/report https://meta.stackexchange.com/q/3122 "PST repair spam"

teward:
## Summary:

I think this is doable, but the amount of complexity we'll have to have in the processing logic for this will make the code coverage/complexity *much* higher than we would probably like.

Technical details on the implementation are below, *but they do not take into consideration whether MetaSmoke will have any issues with this implementation*.

------

------

## Implementation Data

In `chatcommands.py`, the Report command (currently on line #1057) has a function declaration like this:

    def report(msg, urls):

... from which we then create a list of URLs inside (on line 1073) with:

    urls = list(set(urls.split()))

This set of URLs, then, is iterated over via a `for` loop as:

    for index, url in enumerate(urls, start=1)

... of which each tuple that is returned by `enumerate` is processed via the single index value for the loop and the URL for report processing.

Therefore, the way this is passed in is that we explicitly take *all* the arguments sent via `!!/report` and then use that to populate the `urls` argument that is passed during the call of the `report` command by splitting on space-delimitation (i.e. "foo bar baz" being the arguments run through a `.split()` will return the list of `["foo", "bar", "baz"]` with the way we're doing the current URL list population).

From SmokeDetector's point of view, this is easily doable, at least for splitting these cases down.  Now, there are four specific cases I look at and build upon, and we're assuming that `msgargs` is going to have all the arguments that were passed to the `report` function as part of its declaration, and we'll build 'reports' instead of URLs:

------

**Case 1: Multiple URLs being reported, each with its own defined reason**: *Doable, but not nice*

[...]

PR #1980, opened 2018-04-07
Add custom reporting reasons for manual reports. (#1965)

teward:
Custom reporting syntax would then be:

`!!/report url1 [url2 url3 ...] ["Custom Reason"]`

... where both custom report reasons *and* more than one URL are both optional.  This allows for the following 4 command syntaxes to work:

```
!!/report url
!!/report url "Custom Reason"
!!/report url1 url2 url3
!!/report url1 url2 url3 "Custom Reason"
```
... where for batch reports they *all* get the same 'custom reason' for each report generated by the batch, and also supports the original "Manually reported" reason without subsequent details as well.

The original "request" for this feature is in #1965, but this needs approval from at least Undo or Art, as well as a third set of eyes before we consider including.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1071,15 +1071,16 @@ def report(msg, args):
 
     output = []
 
-    argsraw = args.split(' "')
+    argsraw = args.split(' "', 1)
     urls = argsraw[0].split(' ')
 
     # Handle determining whether a custom report reason was provided.
     try:
-        custom_reason = ' "'.join(argsraw[1:])
         # Custom handle trailing quotation marks at the end of the custom reason, which could happen.
-        if custom_reason[-1] is '"':
-            custom_reason = custom_reason[:-1]
+        if argsraw[1][-1] is '"':
+            custom_reason = argsraw[1][:-1].replace('" "', '; ')
+        else:
+            custom_reason = argsraw[1].replace('" "', '; ')
     except IndexError:
         custom_reason = None
 
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1083,7 +1083,7 @@ def report(msg, args):
     except IndexError:
         custom_reason = None
 
-    urls = list(set(urls.split()))
+    urls = list(set(urls))
 
     if len(urls) > 5:
         raise CmdException("To avoid SmokeDetector reporting posts too slowly, you can "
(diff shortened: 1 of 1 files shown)


gparyani:
@teward This is missing a crucial component of my initial filing: I'd like the reason to be specified in Smokey's messages too, like this:

    [ SmokeDetector | MS ] Manually reported [question/answer] (reason: [reason]): [link]

teward:
@gparyani Lets reserve discussion on the commit and the feature request for the actual request thread, and not the PR.  See my reasoning on the other thread for why I disagree with adjusting the initial chat post messages, versus why adjusting the why reason to display the custom reason is better.

gparyani:
@teward Okay, I see why it's a bad idea. (Do note that it *was* part of the original request: "Implementing this would allow the reporter to supply a reason indicating that the post is a spam seed, which would be shown [...] in Smokey's messages in all chat rooms.")

chatcommands.py
```diff
@@ -1070,7 +1070,20 @@ def report(msg, urls):
...
 
     output = []
-    urls = list(set(urls.split()))
+
+    argsraw = args.split(' "')
+    urls = argsraw[0].split(' ')
+
+    # Handle determining whether a custom report reason was provided.
+    try:
+        custom_reason = ' "'.join(argsraw[1:])
```
ArtOfCode-: Seems like it might break if someone tries to provide multiple reasons.

teward: ArtOfCode: The only other solution is that if there are multiple items beyond 2 total items in the first split set to return an error and a failure case of "Error: that is not proper syntax for an error reason, we can only accept a single report reason".  After I am done with the weekend maintenance today at work and I am sufficiently rested, I'll see if we can "fix" this, but if I may ask in what cases do you think this might fail?  Do you have an example case?

teward: I had to write it this way to include the "tripleee might want to include quotes within the reason" edge case brought ul in chat.  If we are hyper restrictive then we can check the length of the first split and if it sees more than two items (URLs and reason) then throw an error back to Smokey to reply in chat that they broke the syntax.  If we want to make it as close to the original logic I wrote as possible, then we have to disallow any of the 'edge cases', which is what my initial plan was.

ArtOfCode-:
Ready to merge?

PR #1994, opened 2018-04-10
phone number updated to include other numbers beforehand

WELZ-gh:
See https://chat.stackexchange.com/transcript/message/43917775#43917775

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -1241,4 +1241,4 @@ jet\W?prox
 size\W?gain\W?plus
 enhance\W?male
 gluco\W?trim
-833\W?783\W?3300
+\W*833\W?783\W?3300


WELZ-gh:
!!/approve

iBug:
Doesn't look any necessary...

j-f1:
How about just `\d+` at the beginning?

tripleee:
For the record, I changed the `\W` to a `\d`

makyen:
This is a US phone number. The only digit that's valid prior to the phone number is 1. Any other digit is much more likely to be a false positive. Another time, I'd suggest that we don't use `\d+` prior to US phone numbers, as it invites false positives.

Even though US phone numbers are only vary rarely written with a 1 immediately prior to the area code, it's possible. Thus, something like `1?833\W?783\W?3300` would be more reasonable. However, I'd consider both limiting the characters which separate the sequences and allowing more of them. Perhaps: `1?833[\s_)-]+783[\s_-]+3300`, or `1?833\W{0,3}783\W{0,3}3300`.

PR #2024, opened 2018-04-18
Update: fag(g?ot)?   -->   fa+g+(ot)?s?\b

K-Davis1:
This is an attempt to decrease fp's when a legitimate word starts with 'fag', such as in a recent FP: Fagnano's.

See the relevant MS post: https://metasmoke.erwaysoftware.com/post/115362

This may actually find more TP's than originally designed: old method wouldn't catch something like:
faaaag, while simultaneously catching less FP's by utilizing a word boundary on the end. No boundary exists in this update at the start of the word - we can adjust if necessary.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -491,7 +491,7 @@ def is_offensive_post(s, site, *args):
         return False, ""
 
     offensive = regex.compile(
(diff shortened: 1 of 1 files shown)


tripleee:
!!/approve

PR #2025, opened 2018-04-18
Remove duplicate watchlist entry

NobodyNada:
Smokey created two watchlist entries for some reason, [apparently in response to one command](https://chat.stackexchange.com/transcript/message/44073072#44073072).


tripleee:
Root cause submitted as #2026

PR #2027, opened 2018-04-18
Bad ns too

tripleee:
Extend "Bad NS" to cover more name servers which are prevalent in hosting domains seen in spam URL links.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -461,9 +461,14 @@ def bad_pattern_in_url(s, site, *args):
 
 
 def bad_ns_for_url_domain(s, site, *args):
+    invalid_tld_count = 0
     for domain in set([get_domain(link, full=True) for link in post_links(s)]):
         if not tld.get_tld(domain, fix_protocol=True, fail_silently=True):
             log('debug', '{0} has no valid tld; skipping'.format(domain))
+            invalid_tld_count += 1
+            if invalid_tld_count > 3:
+                log('debug', 'too many invalid TLDs; abandoning post')
+                return False, ""
             continue
         try:
             starttime = datetime.now()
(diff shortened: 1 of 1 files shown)


tripleee:
The basic work was done in November, but I have retested and revalidated just now, with 1000+ messages so far scanned and 9 "bad NS" hits where the old code only produced 7. (The additional hits were already blacklisted, so this would merely increase their autoflagging weight.)

Going forward, I hope to see some actual added value in detection, too; and based on my observations so far, at the very least it should increase the average accuracy of this rule.

findspam.py
```diff
@@ -478,10 +483,22 @@ def bad_ns_for_url_domain(s, site, *args):
...
+        for bad_ns in [
+                # Don't forget the trailing dot on the resolved name!
+                # {'dns1.namecheaphosting.com.', 'dns2.namecheaphosting.com.'},
+                # {'dns11.namecheaphosting.com.', 'dns12.namecheaphosting.com.'},
+                'namecheaphosting.com.',
+                {'ns1.md-95.bigrockservers.com.', 'ns2.md-95.bigrockservers.com.'},
+                {'ns1.md-99.bigrockservers.com.', 'ns2.md-99.bigrockservers.com.'},
+                {'apollo.ns.cloudflare.com.', 'liz.ns.cloudflare.com.'},
+                {'chip.ns.cloudflare.com.', 'lola.ns.cloudflare.com.'},
+                {'lloyd.ns.cloudflare.com.', 'reza.ns.cloudflare.com.'}]:
```
ArtOfCode-: This is a bad idea... Charcoal website used lloyd and reza for a while.

ArtOfCode-: _shrug_ Let's see how it does, I guess

PR #2042, opened 2018-04-19
Post title sanitize

ByteCommander:
Sanitize post titles in markdown posts, which were not properly escaped (sometimes escaped twice, sometimes not at all).
See https://chat.stackexchange.com/transcript/11540?m=44097072#44097072 and https://chat.stackexchange.com/transcript/11540?m=44097075#44097075 for examples of the problem.
This pulls the sanitation out into a separate method that can be called from all locations where it is required in the same way.
Also fix some warnings from PyCharm's code analysis.


teward:
Ran this in a test on TCOM, and this didn't break handling spam.  Looks OK to me, but I'd like a second set of eyes.

PR #2108, opened 2018-04-27
API queue 1 for Hindiusm

normalhuman:
Hindiusm is not a large site, but has a large number of troll answers that SD is sometimes slow to detect because of the queue.

Trailing comma per PEP8


ArtOfCode-:
We both did this at the same time. 8c7a56338acf9b30cf1930bdb3aed897a74bc768

PR #2111, opened 2018-04-28
Unwatch "good answer"

NobodyNada:
The regex `(nice|useful|good)\W?(article|answer|blog)` is intended to catch blogspam, but "good answer" causes [a lot of fps](https://metasmoke.erwaysoftware.com/search?utf8=✓&title=&body=good+answer&username=&why=&site=&post_type=&feedback=&autoflagged=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search).  This PR removes "answer" from the regex.


LunarWatcher:
!!/approve

PR #2115, opened 2018-05-01
Repair regexes with invisible characters and improve Makyen regex

makyen:
Remove invisible characters added by chat. It appears that I'm the only one that's added regexes long enough to be affected.
Improve the longer offensive regex


ArtOfCode-:
!!/approve

Issue #2012, opened 2018-04-15
!!/scan-ning an answer seems to scan the question in addition


Glorfindel83:
See [here](https://chat.stackexchange.com/transcript/message/44016288#44016288) where I instructed SmokeDetector to scan an answer, but it responded with

> Potentially bad keyword in title:

(I'd expected it to trigger on blacklisted user, but the troll created another account with the same name).

iBug:
Should be fixed in #2126, closing.

PR #2126, opened 2018-05-03
Don't check title when scanning answers

iBug:
See #2012

Attempt to fix this by changing post title to some placeholder text if the post is an answer.


angussidney:
This feels a bit hackish. I know it's the easiest way to solve the problem, but I'm not sure whether we want to leave it this way or simply fix the actual problem. It could work in the interum though.

PR #2129, opened 2018-05-03
Fixes regex

LunarWatcher:
\u200b and a missing question mark broke the regex. The presence of \u200b (and the removal of them) is impossible to tell from the rendered output, but they can at least be seen in the GitHub text editor.


makyen:
!!/approve

PR #2130, opened 2018-05-05
Makyen avoid some FP

makyen:
Change regex to not detect some FP we've seen in the last couple days (e.g. "kill itself").

PR #2138, opened 2018-05-08
Makyen avoid some FP --autopull

makyen:


tripleee:
Apparently a reissue of #2130

PR #2142, opened 2018-05-08
Fix some FP introduced with previous change.

makyen:
My last change eliminated some FP tightening some things up, but, unfortunately, introduced some FP by attempting to catch some malformed words, which could allow the regex to also match some words which shouldn't be caught. This tightens the malformed words back up to prevent the new FP.


LunarWatcher:
!!/approve

PR #2151, opened 2018-05-10
Remove duplicate watchlist entry

ByteCommander:
This was the cause for some failing CI builds


ByteCommander:
Closing because it's identical to #2150

PR #2160, opened 2018-05-12
Move urlopener\.com from URL blacklist to watchlist.

makyen:
This site does not qualify for inclusion in the blacklist.
This site, urlopener.com ([metasmoke search](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title=&body_is_regex=1&body=%28%5E%7C%24%7C%5CW%29urlopener%5C.com%28%5E%7C%24%7C%5CW%29&username=&why=&site=&post_type=&feedback=&autoflagged=&reason=&user_rep_direction=%3E%3D&user_reputation=0&commit=Search)), has only 2 TP, which are from 2 years ago. It recently picked up an FP, which was reported 3 times. Even discounting the FP, it doesn't have enough TP to justify being on the blacklist. With the FP, it's TP rate is 66.7% (counting all FP as one report). If we count all the FP reports of that one post separately, then urlopener.com's TP rate is only 40%.

### DIFF
--- a/blacklisted_websites.txt
+++ b/blacklisted_websites.txt
@@ -502,7 +502,6 @@ nutribulletrecipes\.org
 wirexapp\.com
 x4up\.org
 decalontop\.com
-urlopener\.com
 mobile57\.com
 learn(spicy|perfact)
 getfitness\.in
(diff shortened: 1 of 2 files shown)


NobodyNada:
!!/approve

PR #2168, opened 2018-05-14
Generate UID links for blacklisted user "why"

iBug:


quartata:
post.post_id is the post that's being scanned, not the post that the user was being blacklisted for

quartata:
the proper fix is to store site, postid instead of the URL really

tripleee:
So is this incorrect now? I like the idea.

ArtOfCode-:
The easy way to do it in this instance would be to just pull the related post ID out of the link - write a regex that takes a link and matches the ID in it.

iBug:
@ArtOfCode- In fact, I'm planning to do it with regex when I have time.

helpers.py
```diff
@@ -103,5 +103,17 @@ def api_parameter_from_link(link):
         return None
 
 
+id_parser_regex = r'https?://[^/]+/[qa]/(\d+)'
```
angussidney: I believe we use `/questions/` URLs in a few places, I'm not sure what format we get here.

iBug: @angussidney I must have forgot that. Changed it now.

PR #2187, opened 2018-05-24
Watchlist: Correct the order of the timestamp and username in one line

makyen:


makyen:
Looking at the file, the separator was converted to spaces. I'm not sure how that happened, as I specifically entered and checked for tabs.

PR #2214, opened 2018-05-31
Account for searching for non-\w characters; remove unintensional non-printing characters

makyen:
Blacklisted usernames: 
* Remove extraneous Unicode Left-to-right Override characters that were at the end of a few usernames. Comparing searches on metasmoke and the listed detections indicate this prevents matching (i.e. these changed blacklisted usernames were not effective).

Blacklisted Keywords:
* Account for searches being bookended by `\b`. When searching for characters from languages other than English (i.e. using non- \w characters) a `\W*` is needed to consume all \W characters up to a `\b` word break.
  Comparing searches on metasmoke and the detections listed for various TP posts indicated in each case this prevents matching (i.e. these changed blacklisted keywords were not effective). There were some few cases for a couple words where these did match correctly on a couple TP, but the significant majority of TP posts where theses existed to be match, they were not detected (after the detection was added to the watchlist/blacklist).

Watchlist:
* Remove some unintentional non-printing characters (mostly Unicode Left-to-right Override characters)
* Again, account for using `\b` bookending while searching for non-English text. [Basically, include the same statements here as were made above for blacklisted keywords wrt. lack of functionality/detection being indicated by comparing search results and listed detections.]

For the blacklisted keywords and watchlist, the drawback for using `\W*` to extend to the next `\b` in both the blacklist keywords and the watchlist is that what's being detected will not be as clear in the "reasons", because what will be shown as the match will be significantly more than the characters/words which we're really looking for.

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -393,9 +393,9 @@ jolique
 clemix
 nuavive
 evanti
-毕业证
+\W*毕业证\W*
 i\Wdidn't\Wfind\Wthe\Wright\Wsolution\Wfrom\Wthe\Winternet
-レーザーポインター
+\W*レーザーポインター\W*
 ultra\Wpure\Wturmeric
 libidogene
 suisse\Wrenewal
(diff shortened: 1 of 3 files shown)


tripleee:
!!/approve

PR #2220, opened 2018-06-01
Demote code-industry\.net to watchlist

makyen:
Per discussion started by @Videonauth [here](https://chat.stackexchange.com/transcript/message/44915752#44915752).
Some research results [here](https://chat.stackexchange.com/transcript/message/44916041#44916041), which show that it's 4/3/1 with two of the TP having conflicting feedback.
There are also [61 hits](https://stackexchange.com/search?page=1&q=url%3acode-industry.net) on SE (tends to show duplicate hits).
I'm not sure that this even deserves a watch at this point, but...


NobodyNada:
!!/approve

PR #2235, opened 2018-06-06
added 'live' and 'errlogs' aliases

WELZ-gh:
because we are lazy and shorter commands are better.


j-f1:
While you’re at it, can you alias `!!/errlogs` too?

ArtOfCode-:
!!/approve

PR #2236, opened 2018-06-06
Add 'errorlog' alias

teward:
We have 'errlog' now, but not 'errorlog' which is a common typo that (at least I) regularly make when trying to access error logs.  So let's add that in there while we're at it.


NobodyNada:
!!/approve

PR #2247, opened 2018-06-12
Improve "body starts with title" detection

iBug:
[This post][1] could've been caught by the reason and thus, autoflagged. Fix by removing white spaces before comparing.

  [1]: https://metasmoke.erwaysoftware.com/post/123116


NobodyNada:
!!/approve

Issue #2251, opened 2018-06-13
Upgrade to tld 0.8


tripleee:
**Smokey runners:** please scroll down to the last comment

As noted in #2250 the latest release of `tld` has backwards-incompatible changes.

https://github.com/barseghyanartur/tld/blob/master/CHANGELOG.rst explains in some details that code should transition to the new method `get_fld()` going forward.

iBug:
For what it's worth, [I have temporarily locked package `tld` to version 0.7.10][1] (which I suppose will be the last release of 0.7.x. If anyone if going to do this upgrade, please remember to release the lock by changing it to `tld>=0.8` after the job is done.

[1]: https://github.com/Charcoal-SE/SmokeDetector/commit/0f9bb66c630a6ce73518e7349dd6ff6950e0ee29

tripleee:
Reopening. Please only close when all the checkboxes are checked.

Check when your instance is updated to have the 0.8.x version of `tld`:
- [ ] teward/Solar Flare
- [ ] teward/Lunar Eclipse
- [ ] teward/Supernova
- [ ] ArtOfCode/EC2
- [ ] Henders/EC2
- [ ] bwDraco/Linode
- [ ] Mithrandir/Tauremornalómë
- [ ] NobodyNada/Raspberry Pi
- [ ] angussidney/EC2
- [ ] Undo's RPI 3

PR #2253, opened 2018-06-13
Update `tld` to 0.8

iBug:
From [`tld`'s changelog][1]:

> 0.8
> 2018-06-13
>
> Note
>
> This release contains backward incompatible changes. You should update your code.

which has caused mysterious CI failure in some time. I'm trying to adapt to the new version of `tld`.

Related: #2251 

[1]: https://github.com/barseghyanartur/tld/blob/master/CHANGELOG.rst#08
[2]: https://chat.stackexchange.com/transcript/message/45127605#45127605


iBug:
@angussidney Good point that I would have missed. Yeah, sure. All instances must have their packages updated by the owners before this is going up.

iBug:
@angussidney I have another concern. Since the 0.8 of `tld` isn't backwards compatible, existing code will start to fail if people update their PyPI package without adding my patch, so it'd require *all* people update their packages and merging this PR *simultaneously*, which looks difficult and impractical.

I can suggest a temporary workaround by adding code to adapt both versions for the time being, and remove adapter code after all instances are updated. Is that a good ~~~solution~~~ workaround?

iBug:
I have pushed the compatibility code. It has passed the CIs for both versions of `tld` (CI for [`0.7`][1] and [`0.8`][2]).

[1]: https://travis-ci.org/Charcoal-SE/SmokeDetector/builds/392199891
[2]: https://travis-ci.org/Charcoal-SE/SmokeDetector/builds/392198988

PR #2262, opened 2018-06-19
Cleanup my recent !!/watch commands.

makyen:
Remove watches missing a `?`. Remove duplicate watch created by SD taking the original version of the message, not the edited version (not sure why that happened).

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -4021,7 +4021,5 @@
 1529333288	WELZ	balajicargopackers\.in
 1529356721	Makyen	evgravingplus\.weebly\.com
 1529366192	Makyen	arsenalloansolutions
-1529369730	Makyen	pageoptimizer\W*(?:dot\W*)pro
 1529369767	Makyen	pageoptimizer\W*(?:dot\W*)?pro
-1529369826	Makyen	bit\W*(?:dot\W*)ly/seoproof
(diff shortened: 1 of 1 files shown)


Undo1:
!!/approve

j-f1:
> Remove duplicate watch created by SD taking the original version of the message, not the edited version (not sure why that happened).

@makyen I think that happens because we grab the original source of the message to avoid accidentally watching `<i>` and stuff, but it probably doesn’t account for edits.

makyen:
@j-f1 That might be it/contributing, but I don't think it is. When we changed to using the raw message content (relatively recently), IIRC, it was a simple change and the raw data already was available from CE, or at least that's what I recall being said.

After the change was made to use the raw message, I've, many times, edited SD command messages, most commonly changing `!!/watch` to `!!/watch-force`. Every time, SD recognized the edited version. I've also seen multiple other people make similar edits and have the edited version recognized by SD. (all based on the responses SD gives to the edit). So, I know SD gets the edited raw message correctly almost all the time (or at least in a the cases I've observed).

This was the first time I've experienced, or seen, SD using the original version of a message upon an edit. There's probably nothing for us to track down at the moment, given the rarity of this happening. It just made it clear that there's the possibility it happens. As a result, SD commands which we don't want the possibility of being repeated (e.g. already applied watchlist & blacklist additions) shouldn't be edited. That's an easy work-around, as long as we know it's needed/desirable.

PR #2270, opened 2018-06-23
Wrap commit messages in code blocks

iBug:
This would be a better solution to the broken chat messages caused by special characters in commit messages than escaping them. However I'm not sure if Smokey's commit message format is important (will break userscripts, etc.).


angussidney:
As far as I know, no userscripts should be affected.

PR #2284, opened 2018-06-28
Grant Shree privileges in SOBotics

rjrudman:
Request has been discussed with room owners, no one raised any objections


iBug:
!!/approve

PR #2286, opened 2018-06-29
What is this regex?

iBug:
```
maleenhancement.{1 7}\..{2 4}
```

I have never learned a syntax like this, and according to my experiences, I suppose it should look like this:

```
maleenhancement.{1,7}\..{2,4}
```

Should it be fixed?

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -1318,7 +1318,7 @@
 1509355242	tripleee	caliber\.com
 1509355872	tripleee	howtotoken\.com
 1509364669	Mithrandir	arabiandesertdubai\.com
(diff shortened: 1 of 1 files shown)


makyen:
Strange. The [`!!/watch`](https://chat.stackexchange.com/transcript/11540?m=40847744#40847744) (has no edit) had the commas in the regex, but [SD's response](https://chat.stackexchange.com/transcript/11540?m=40847745#40847745) clearly indicated that SD didn't see them.

So, the commas were certainly *intended* to be there.

j-f1:
That’s a normal comma, so it’s possible it was just a glitch.

Issue #2289, opened 2018-06-30
!!/report should !!/scan first


longsonr:
It seems that we're supposed to use [!!/scan](https://stackoverflow.com/c/charcoal/questions/101/102#102) now either before or instead of !!/report. 

Rather than having to remind occasional users of smokey to change their habits and have !!/report be a foot gun for them, why not...

1. have !!/report do a !!/scan first which falls back to !!/report if not detected

2. if you feel it necessary introduce a !!/report-no-scan command (or something like that) for the times when the user just wants to report w/o scanning.

that way those people used to using !!/report can continue to do so. 

This idea was originally suggested [here](https://chat.stackexchange.com/transcript/message/45394636#45394636) and [here](https://chat.stackexchange.com/transcript/message/45394662#45394662)

iBug:
Good idea. TBH I've though of this before and I just kept it personal. Now that someone brings this up, I'll make it work.

PR #2290, opened 2018-06-30
Scan multiple URLs with one chat message

iBug:
The `!!/scan` command currently only scans one post per message. It's more convenient to scan more messages in one message. Results are either given as reports (problem found) or squashed into one single chat message.

This is a preparation for implementing #2289. I'll start implementing that after this one is done and merged.

### DIFF
--- a/test/test_chatcommands.py
+++ b/test/test_chatcommands.py
@@ -172,7 +172,7 @@ def test_report(handle_spam):
 
 
 @patch("chatcommands.handle_spam")
-def test_checkpost(handle_spam):
+def test_scan(handle_spam):
     try:
         msg = Fake({
             "owner": {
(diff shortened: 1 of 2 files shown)


test/test_chatcommands.py
```diff
@@ -172,7 +172,7 @@ def test_report(handle_spam):
 
 
 @patch("chatcommands.handle_spam")
-def test_checkpost(handle_spam):
+def test_scan(handle_spam):
```
NobodyNada: Maybe add test cases for scanning multiple posts?

iBug: Nah, not necessary at this point. I'm going to merge `scan` with `report` as said in the last paragraph of the PR description. I'll add tests then.

PR #2291, opened 2018-07-01
+Frenzy Li to CHQ

iBug:
[Request here](https://chat.stackexchange.com/transcript/message/45422994#45422994) and [chat user profile here](https://chat.stackexchange.com/users/40662).

### DIFF
--- a/rooms.yml
+++ b/rooms.yml
@@ -122,6 +122,7 @@ stackexchange.com:
       - 311852   # Filnor
       - 113120   # Gaurang Tandon
       - 153555   # bertieb
+      - 40662    # Frenzy Li
 
   65945:  # Charcoal Test
     commands: true


NobodyNada:
What's the motivation for asking for privileges?  This user seems to have hardly interacted with us and Smokey, only [reporting one post](https://chat.stackexchange.com/transcript/11540?m=41580991#41580991) and [asking for privileges](https://chat.stackexchange.com/transcript/11540?m=41613390#41613390) in December, and has not participated in chat very much since about a year ago.

IMO they haven't yet demonstrated trustworthiness and familiarity with the system, though I'm willing to be persuaded otherwise/overruled.  My suggestion is that they hang around in chat for a bit to get to know the system, and then we add them.

ArtOfCode-:
Default-allow and revoke if necessary

ArtOfCode-:
!!/approve

PR #2292, opened 2018-07-01
Force scan with !!/report, merge !!/scan in

iBug:
#2289 is a good idea. I had thought of it several times before but didn't bring it up.

This PR basically just implements what's proposed in #2289, with the following changes:

- `!!/scan` gets merged into `report()` as an alias
- `!!/report` acts the same as `!!/scan` if the given post is determined as spam
- `!!/report` falls back to original "report" if the post isn't caught, while `!!/scan` will tell you "it doesn't look like spam"
- New command `!!/report-force`, which as intuitive as its name suggests, skips the "scan first" and acts exactly like the original `!!/report`, with scan analysis supplied in the `why` section
- A new `to_metasmoke_link` function that simplifies code and increases maintainability. More uses of this function ~~will come up in another PR or be pushed directly where applicable~~ are dispatched already
- Relevant CI tests for the merged function

One side effect is that `!!/scan` will become a privileged command. This shouldn't cause a lot of trouble as all those who issues this command are privileged users. (Hard to fix)

P.S. I personally don't want `!!/report-force` to be canonically documented because it isn't quite necessary. Users are encouraged to stick to `!!/report`.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -367,7 +367,7 @@ def brownie():
     return "Brown!"
 
 
-COFFEES = ['Espresso', 'Macchiato', 'Ristretto', 'Americano', 'Latte', 'Cappuccino', 'Mocha', 'Affogato']
+COFFEES = ['Espresso', 'Macchiato', 'Ristretto', 'Americano', 'Latte', 'Cappuccino', 'Mocha', 'Affogato', 'jQuery']
 
 
 # noinspection PyIncorrectDocstring
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1198,7 +1198,7 @@ def report(msg, args, alias_used="report"):
 
         # scan_spam == False and alias_used == "scan"
         else:
-            output.append("Post {}: This does not look like spam")
+            output.append("Post {}: This does not look like spam".format(index))
 
     if 1 < len(urls) > len(output):
         add_or_update_multiple_reporter(msg.owner.id, msg._client.host, time.time())
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1112,7 +1112,7 @@ def report(msg, args, alias_used="report"):
                                "for guidance on using custom report reasons.")
 
         if alias_used == "scan":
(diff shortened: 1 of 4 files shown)


j-f1:
Can you add back the tests for `!!/scan` and update them to use the `report` function with `alias_used="scan"`?

iBug:
@j-f1 Done. Should look fine now.

spacemonaut:
> `!!/scan` gets merged into `report()` as an alias

Scan sometimes gets used to scan for a match, *without* any intention of actually reporting. Does that remove this option, such that scan will always report one way or another?

chatcommands.py
```diff
@@ -1141,128 +1143,68 @@ def report(msg, args):
...
-        # Update url to be consistent with other code
-        url = to_protocol_relative(post_data.post_url)
-        post = Post(api_response=post_data.as_dict)
-
-        if has_already_been_posted(post_data.site, post_data.post_id, post_data.title) \
-                and not is_false_positive((post_data.post_id, post_data.site)):
-            # Don't re-report if the post wasn't marked as a false positive. If it was marked as a false positive,
-            # this force scan might be attempting to correct that/fix a mistake/etc.
+        # scan_spam == False
+        if alias_used in {"report", "report-force"}:
```
spacemonaut: Aha, so we only wind up submitting a custom report if we use one of these two aliases, so `!!/scan` never submits a manual report. :)

ArtOfCode-:
!!/approve

ArtOfCode-:
Merge when ready

chatcommands.py
iBug: @spacemonaut I just pushed [a commit][1] that enabled `scan` with custom reasons. The report should look like this now:

```
Post manually scanned by *iBug* in room *Charcoal HQ* with reason: *\_(ツ)_/*.

Body: Position 123-456: blah blah blah
```

[1]: https://github.com/Charcoal-SE/SmokeDetector/commit/50af479ba164aea606191ca4df7078f82585a2d6

PR #2293, opened 2018-07-01
Remove tld 0.7 compatibility code, expect tld 0.8

iBug:
The upgrade of this dependency should be done by now. Anyone can merge this request to finalize this upgrade.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -141,19 +141,6 @@ def contains_tld(s):
     return any([('.' + x) in s for x in TLD_CACHE])
 
 
-# Code maintainer:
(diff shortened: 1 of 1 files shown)


NobodyNada:
Several instances are not checked off the list as having been upgraded; have those been upgraded anyway?

j-f1:
How about adding a temporary commit that pings the instance’s owner to upgrade if that hasn’t happened yet?

iBug:
Based on [Thomas's response][1] I believe we're good to go. Also note that my message was pinned for long enough for every regular to see it.

[1]: https://chat.stackexchange.com/transcript/message/45426622#45426622

PR #2294, opened 2018-07-02
!!/unblacklist command

iBug:
New command `!!/unblacklist`. Removes an item from the three blacklist files. This would be useful when people think a blacklisted item generates too many FPs.

Same as `!!/unwatch`, code-admins only, no PR.

There was originally an idea about something like `!!/demote` that moves a blacklist item to watchlist, but I've decided to suspend it after evaluating the difficulty of implementing it over the current code base.


ArtOfCode-:
!!/approve

iBug:
Tested in production. Looks all good.

PR #2296, opened 2018-07-04
Catch punctuation spam

FYPetro:
This is an improvement on mostly_dots.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -3,6 +3,7 @@
 
 import math
 import regex
+import string
 from difflib import SequenceMatcher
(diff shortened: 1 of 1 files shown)


iBug:
I think this can be further improved to check if the majority of a post is punctuation (not limited to dots) and its usefulness can be further improved. I don't think there are many one-punct-man posts to catch.

FYPetro:
@iBug You know what... I agree with your assessment. A simple remedy would be to calculate the sum and see if it exceeds CHARACTER_USE_RATIO.

PR #2298, opened 2018-07-04
Ignore if Community user bumped post

### LINKED ISSUE #1085: Avoid catching the same post if it hasn't been modified since last report

paper1111:
Implements the functionality requested in #1552 

Fixes #1085


iBug:
Looks very good. Are you sure MS is ignored when it's down? This happens from time to time so it's important that Smokey functions correctly during MS downtime.

paper1111:
@iBug added a timeout so that if MS is down, it will be skipped

Quite strange that CircleCI fails while Travis passes, but anyways I think that's not my fault.

iBug:
You're running CircleCI with an already discontinued version of `tld`. Try overwriting your `user_requirements.txt` with the one in the master branch.

Travis CI is installing and updating dependencies on every run, while CircleCI only updates them if they no longer meets the requirements specified in the repository. This is why Travis is using the correct version but Circle sticks to an old version.

paper1111:
@iBug it **is** checked already

paper1111:
ummm... I edited the file but it still failed with the same issue..?

iBug:
From the build logs, it seems Circle pulls and tests your working tree directly, while Travis pulls the target branch (`Charcoal-SE:master`) and merges your branch in. This means Travis tests the state after the PR should be merged.

`tld` has updated to 0.9, which contains backward-incompatible change that breaks your code (you started this PR from a state in mid-May, which doesn't contain any patches for `tld` 0.8+). Since the master branch is already adapted for `tld` 0.8+, your code passes the tests after being merged with master. This is why Travis and Circle are giving different results.

Anyway, since Circle is fixed now, you can ignore this.

paper1111:
Great, thanks!

PR #2299, opened 2018-07-05
Test blacklist integrity before commit

### LINKED ISSUE #2219: Duplicate watches still happen
It is not at all infrequent for two people to `!!/watch something` more or less at the same time in chat, and have two separate branches be created, tested, and merged without conflict but resulting in a CI failure after the merge.

The run-time checking against duplicates doesn't help at all here, because it only examines the current branch, and arguably couldn't know whether another branch is also being created at some point during the time it is processing the command.

iBug:
Attempt to fix #2219 by checking for duplicates on `master` instead of `deploy`.

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -67,6 +67,9 @@ class GitManager:
             if not status:
                 return (False, message)
 
+            # Checkout master so other pending changes are also there
+            git.checkout("master")
+
             if blacklist_type in [Blacklist.WATCHED_KEYWORDS]:
                 op = 'watch'
                 now = datetime.now().strftime('%s')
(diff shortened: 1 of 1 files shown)


iBug:
@ArtOfCode- I absolutely didn't forget that. [Here][1]'s a `finally` clause that has been there for a lone time `:)`

  [1]: https://github.com/iBug/SmokeDetector/blob/f83d17c4c187303a8c5f49dbe96305a461cceb52/gitmanager.py

ArtOfCode-:
@iBug that `finally` change hasn't made it to this PR; did you commit to the patch branch in your repo or master?

iBug:
@ArtOfCode- That `finally` block has existed in `master` for a long time and I didn't touch it.

iBug:
@ArtOfCode- I'm sure what you reviewed isn't an issue and am going to merge this. Is that all right?

ArtOfCode-:
There's still no finally block included in the changes in this PR

iBug:
@ArtOfCode- There's no need to have another. The one that I presented you with already covers this case and it's in the master branch.

ArtOfCode-:
Oh, I get it now. I was confused about which changes were on which fork. Merge when ready.

PR #2300, opened 2018-07-05
Adapt for MS downtime in (un)blacklist/watchlist commands/replies

teward:
If MS is down we get a connection error problem.  When that happens, we should force code_permissions to be False.  However, to get the proper chat reply that MS is down and a PR made because we can't determine if someone is code privileged or not, we need a metasmoke_down tracker in the 'add to blacklist' function so we can determine which chat reply to send.  This way, we can make PRs if MS is down and we can't determine if someone has code privs.

However, the same can't be said about unblacklist.  It doesn't have PR-branch-creation functionality at all.  So the best we can do is fail with a specific error message.  unblacklist/unwatchlist will need refactored to be able to do that functionality before we can force PRs if someone isn't code-priv'd.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -260,27 +260,22 @@ def do_blacklist(blacklist_type, msg, force=False):
             raise CmdException("That pattern looks like it's already caught by " + format_blacklist_reasons(reasons) +
                                "; append `-force` if you really want to do that.")
 
-    result = None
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
Pedantry: metasmoke, not MetaSmoke

ArtOfCode-:
MWR

PR #2302, opened 2018-07-06
Ignore experimental reasons if the post is manually reported

iBug:
See https://chat.stackexchange.com/transcript/message/45508505#45508505 and https://chat.stackoverflow.com/transcript/message/43164457#43164457

This is a workaround by detecting if the `why` text starts with `Post manually`, which should always be the case if the post is manually scanned/reported.


iBug:
Tested in SOBotics, it's working.

Issue #2217, opened 2018-05-31
"misleading link text" should not trigger on IDNA


tripleee:
Live example: https://chat.stackexchange.com/transcript/message/44901038#44901038

> Domain h%c3%a5nd.no indicated by possible misleading text hånd.no.

PR #2315, opened 2018-07-08
Don't catch IDNA domains as "misleading links"

iBug:
See #2217

I was quite resistant about doing this before I noticed that Python has `str.encode('idna')`. So I just happily did this. URL-decoding is also involved in the post mentioned in 2217 so I added support for it, too.


ArtOfCode-:
CI

PR #2319, opened 2018-07-10
Escape markdown link name

FYPetro:
To fix [messages with \\, \[ and \] in the link name](https://chat.stackexchange.com/transcript/message/45561684#45561684).


helpers.py
```diff
@@ -29,6 +29,10 @@ def escape_format(s):
     return s.replace("{", "{{").replace("}", "}}")
 
 
+def escape_markdown_linkname(s):
```
iBug: I'm sure there's already a function named `escape_markdown` located somewhere.

PR #2321, opened 2018-07-10
Remove `cloud-init`

iBug:
This file looks like it's already caught for: Too old, outdated, and hardly used.

Shall we improve its usefulness or delete it from the working tree? I think it's better deleted so the work tree is cleaner.

Summoning @Undo1


Undo1:
If it doesn't work anymore, definitely delete it. Easy enough to dig out of the history of we want it back.

Undo1:
!!/approve

PR #2322, opened 2018-07-11
Allow the watchlist and blacklist to match the start and end of strings

makyen:
The Python `\b` assertion matches `^\w` and `\w$`, but it doesn't match `^\W` and `\W$`. If we want to have blacklist and/or watchlist entries which are able to match the beginning or end of the body (and/or titles and usernames when looking for `\W` characters), then we need to bookend the matches with `(?:^|\b)` at the start and `(?:\b|$)` at the end, rather than just with `\b` on both sides.

Documentation: [6.2.1. Regular Expression Syntax](https://docs.python.org/3/library/re.html#regular-expression-syntax)

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1054,11 +1054,11 @@ class FindSpam:
         #
         # Category: Bad keywords
         # The big list of bad keywords, for titles and posts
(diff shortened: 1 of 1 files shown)


iBug:
!!/approve

PR #2329, opened 2018-07-14
Fetch message source for blacklist/watch commands

iBug:
From time to time people edit blacklist commands and it's known that Smokey takes the source of the original (unedited) message, which has caused some troubles. Manually fetching the latest message source is a solution, and since it's applied only on blacklist/watch commands, it won't have a mattering performance impact.

For my analysis details [read down from this message](https://chat.stackexchange.com/transcript/message/45646416#45646416).

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 # noinspection PyUnresolvedReferences
 from chatcommunicate import add_room, block_room, CmdException, command, get_report_data, is_privileged, message, \
-    tell_rooms
+    tell_rooms, fetch_source
 # noinspection PyUnresolvedReferences
 from globalvars import GlobalVars
 from findspam import FindSpam
@@ -241,6 +241,7 @@ def do_blacklist(blacklist_type, msg, force=False):
     :return: A string
     """
 
+    msg.content_source = fetch_source(msg)
     chat_user_profile_link = "https://chat.{host}/users/{id}".format(host=msg._client.host,
                                                                      id=msg.owner.id)
 
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -141,6 +141,18 @@ def pickle_last_messages():
             pickle.dump(_last_messages, pickle_file)
 
 
+def fetch_source(msg):
+    headers = {"Content-type": "text/plain"}
+    try:
+        response = requests.get(r"https://chat.{}/message/{}?plain=true".format(msg._client.host, msg.id), headers=headers)
+    except:  # Something wrong, not handling them for now, use placeholder
+        return msg.content_source
+
+    if response.status_code == 200:
+        return response.content.decode(response.encoding)
+    return msg.content_source
+
+
 def send_messages():
     while True:
         room, msg, report_data = _msg_queue.get()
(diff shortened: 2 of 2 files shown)


quartata:
"Smokey takes the source of the original message" this is not true. The only time these things happen is when a cached version of the original message is returned instead of the edited message

quartata:
I haven't looked at the !!/blacklist command in a while but it looks like people have made it use content_source now, I'm going to probably need to fix that

makyen:
@quartata I'm unsure what you mean by "fix that" in this context. One possible interpretation is that you intend to just revert back to the old functionality.

I don't know Python, so I'm not qualified to review the code, but from your comments, I figured you could use some background/context information.

The `!!/watch` and `!!/blacklist` commands were changed a couple months ago to use the raw message (I understand that to mean the `content_source`). This was done so that people did not have to remember to `\` quote all of the characters in their regexes which SE was interpreting as Markdown. The way that SD had been doing it prior to that change resulted in many added watches and blacklisted regexes that needed to be repaired (primarily due to pairs of `*` being changed to `<i>`/`</i>`. My impression is that people would like the functionality of not having to worry about SE's Markdow interpretation of their regexes. I know that I certainly prefer not needing to do so.

The way that change was implemented has been working for a couple of months. There's been a minor issue of, from time to time, when a user edits their command to SD that SD takes the original version. From a user's perspective, the negative effects of this (potentially duplicated `!!/watch`/`!!/blacklist` entries), happened irregularly.

To fix that, @iBug has:

* (in a different PR, already merged) made changes such that SD won't apply duplicate watch or blacklist items. This also addresses issue #2219 with PR #2299 (and a subsequent commit).
* This current PR: @iBug [noted](https://chat.stackexchange.com/transcript/message/45646416#45646416) that the issue of not taking the edited version was that SD did take the edited version of the command, but not the raw arguments used by `!!/watch` and `!!/blacklist` for the regex. This is because the `content_source` is never updated to the edited version. Thus, `!!/watch` and `!!/blacklist` always see the original version of the message for their argument, the regex.

  This PR is intended to address that issue by always fetching the current `plain` version directly from SE chat via `/message/[id]?plain=true`.

iBug:
@quartata I checked the source code of ChatExchange and it uses BS4 to parse the HTML of `/message/{id}/history` page. On the `content_source` part, it takes the last "message block" which contains the markdown source of the first revision of the message. That's why `content_source` is always the MD source of the original message - it's CE's fault. So I fetch the source again to avoid this issue.

quartata:
Oh, then this is a separate issue. Still, CE is the one that should be patched

quartata:
One of the reasons CE should be patched instead is that you'll introduce strange behavior by overwriting the content_source property, since message objects are reused. (It will cause at least one big with autocomments, that I know of)

iBug:
@quartata It does no harm to update `content_source` to the correct one. My function returns the original value if the new value failed to fetch, so there shouldn't be anything wrong.

iBug:
@quartata uhm, my previous analysis is a bit wrong. The `get_history` function in `browser.py` ***is*** parsing the correct message block, so if we used `get_history` on a message, we would've got the edited message MD source.

What's actually wrong is the message object updater (`events.py` line 90). Upon an `event.MessageEdited`, `content_source` is not updated (it's initialized, however), so a message object containing the old MD source in `content_source` is passed to Smokey. This is, still, something that needs to be patched.

---

After more investigation and testing, I found that the edited source is not dispatched via websocket, which means we have to manually grab them whenever we want the latest message source.

quartata:
@iBug OK I think I get what's going on now. It's sounding like the problem might be that the LazyFrom for content_source is run once (when it's retrieved at MessagePosted) and then since the same object is used for the MessageEdited it doesn't run again.

iBug:
Already done by quartata in c0bf0004141e2bb33651e32d33b000e4d68fd4f4. Good approach then.

PR #2348, opened 2018-07-27
Show why the post is not spam in !!/scan

iBug:
When people scan a post, there are only two outcomes: a report, or "This does not look like spam". Sometimes when it's very obvious the post looks like spam, the "does not look like" response can be confusing because the post is actually not reported for another reason (ignored post, whitelisted user, etc.).

This PR makes specific "reasons for not reporting" available, and also corrects `!!/report`-ed posts that would have been caught but ignored and thus considered "would not have been caught".


AWegnerGitHub:
!!/approve

PR #2355, opened 2018-07-31
Drop requirement of dulwich

iBug:
`dulwich.repo` is only used once for obtaining Git commit information, which can well be replaced with a subprocess call to CLI `git-log`.

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -16,7 +16,8 @@ from flovis import Flovis
 
 
 def git_commit_info():
-    data = sp.Popen(['git log -1 --pretty="%h%n%H%n%an%n%s"'], shell=True, cwd=os.getcwd(), stdout=sp.PIPE, stderr=sp.PIPE).communicate()
+    data = sp.Popen(['git log -1 --pretty="%h%n%H%n%an%n%s"'],
+                    shell=True, cwd=os.getcwd(), stdout=sp.PIPE, stderr=sp.PIPE).communicate()
     if data[1]:
         raise OSError("Git error:\n" + data[1].decode('utf-8'))
     short_id, full_id, author, message = data[0].decode('utf-8').strip().split("\n")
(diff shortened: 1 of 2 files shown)


globalvars.py
```diff
@@ -11,17 +11,17 @@
...
 from flovis import Flovis
 
 
 def git_commit_info():
-    git = Repo('.')
-    commit = git.get_object(git.head())
-    return {'id': commit.id.decode("utf-8")[0:7], 'id_full': commit.id.decode("utf-8"),
-            'author': regex.findall("(.*?) <(.*?)>", commit.author.decode("utf-8"))[0],
-            'message': commit.message.decode("utf-8").strip('\r\n').split('\n')[0]}
+    data = sp.Popen(['git log -1 --pretty="%h%n%H%n%an%n%s"'],
```
tripleee: I have a number of issues with this. I like the idea to get rid of Dulwich but ... avoid `subprocess.Popen()` in favor of `subprocess.run()`, avoid `shell=True` (and while you *had* `shell=True`, don't put the command line in a list -- harmless but mispeading); and don't needlessly force a `cwd=` which sacrifices a system call to not change the current working directory.

    data = sp.run(['git', 'log', '-1', '--pretty="%h%n%H%n%an%n%s"'],
         shell=False, stdout=sp.PIPE, stderr=sp.PIPE, universal_newlines=True)

The result from `sp.run()` is slightly different so the following code lines will need to be adapted accordingly.

Also, are we sure this mode of `git` is completely robust and that it always generates predictable output?

iBug: `sp.run` is available as of Python 3.5, so I intentionally avoided it because Hender's EC2 is known to run Python 3.4.x.

The other question about using `git log --pretty`, I picked this approach because it's an answer to virtually any question about "programatically get information about a commit / latest commit", so I believe it's a method safe enough to employ here. I chose the newline as field separator because it's not a valid character in all fields.

PR #2358, opened 2018-07-31
Load ChatExchange credentials from config file

iBug:
In addition to loading CE creds from environment variables, this enables the loading from local `config` file. If both are present, the ones defined in environment takes precedence, so this does essentially nothing to running instances.

As a downside, I removed asking for input via `input` and `getpass` as it causes confusion.

Previously CE cred is asked in both `nocrash.py` and `ws.py` if it's not found. It can be confusing if one is present in `config` but still asked from input, because `nocrash.py` does not load config. There were two options available:

1. Load `config` in `nocrash.py` and suppress terminal input if cred is found
2. Totally drop terminal input for cred

To keep things simple, I picked 2, because `nocrash.py` is a replacement for `nocrash.sh` and does not depend on any other files, and it's important (at least to me) to keep this isolation.

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -133,10 +133,17 @@ class GlobalVars:
     metasmoke_ws = None
 
     try:
+        chatexchange_u = config.get("Config", "ChatExchangeU")
+        chatexchange_p = config.get("Config", "ChatExchangeP")
+    except NoOptionError:
+        chatexchange_u = None
+        chatexchange_p = None
+
+    try:
         metasmoke_host = config.get("Config", "metasmoke_host")
     except NoOptionError:
         metasmoke_host = None
-        log('info', "metasmoke host not found. Set it as metasmoke_host in the config file."
+        log('info', "metasmoke host not found. Set it as metasmoke_host in the config file. "
             "See https://github.com/Charcoal-SE/metasmoke.")
 
     try:
--- a/ws.py
+++ b/ws.py
@@ -78,15 +78,17 @@ except TldIOError as ioerr:
             raise ioerr
 
 if "ChatExchangeU" in os.environ:
+    log('debug', "ChatExchange username loaded from environment")
     username = os.environ["ChatExchangeU"]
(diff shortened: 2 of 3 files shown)


ws.py
```diff
@@ -80,12 +79,21 @@
...
+elif GlobalVars.chatexchange_u:
+    log('info', "ChatExchange username loaded from config")
+    username = GlobalVars.chatexchange_u
 else:
-    username = input("Username: ")
+    log('error', "No ChatExchange username provided. Set it in config or provide it via environment variable")
+    os._exit(6)
+
 if "ChatExchangeP" in os.environ:
     password = os.environ["ChatExchangeP"]
```
AWegnerGitHub: Can a similar `info` log be added here so that we know they were loaded from environment variables?

globalvars.py
```diff
@@ -132,11 +132,18 @@ class GlobalVars:
...
+    except NoOptionError:
+        chatexchange_u = None
+        chatexchange_p = None
+
     try:
         metasmoke_host = config.get("Config", "metasmoke_host")
     except NoOptionError:
         metasmoke_host = None
-        log('info', "metasmoke host not found. Set it as metasmoke_host in the config file."
+        log('info', "metasmoke host not found. Set it as metasmoke_host in the config file. "
```
ArtOfCode-: Random space?

ws.py
```diff
@@ -80,12 +79,21 @@
...
+elif GlobalVars.chatexchange_p:
+    log('info', "ChatExchange password loaded from config")
```
ArtOfCode-: Can this be debug not info? It's only really going to be relevant if you're debugging an issue - in most cases, you're going to know where it was loaded from.

globalvars.py
j-f1: Currently, it looks like there’s no space between this sentence and the next one. Adding the space here fixes that.

ArtOfCode-:
LGTM

PR #2363, opened 2018-08-03
Watch without force.

tripleee:
- Phone number watches no longer trigger Mostly non-Latin

- Potentially bad NS hits do not block watching

- Refactor blacklist checks to properly check both username and
  regular blacklists (bug fix)

- Small test case for !!/watch

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -182,7 +182,7 @@ def blacklist(_):
                        "Remember to escape dots in URLs using \\.")
 
 
-def check_blacklist(string_to_test, is_username, is_watchlist):
+def check_blacklist(string_to_test, is_username, is_watchlist, is_phone):
     # Test the string and provide a warning message if it is already caught.
     if is_username:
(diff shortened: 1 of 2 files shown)


tripleee:
This augments an earlier commit of mine which went directly to `master` but turned out not to work.

This became slightly confounding so I'm going via a PR for added visibility and review.

chatcommands.py
```diff
@@ -207,15 +207,16 @@ def check_blacklist(string_to_test, is_username, is_watchlist):
     reasons = list(set(question_reasons) | set(answer_reasons))
 
     # Filter out watchlist results
+    filter_out = ["potentially bad NS"]
```
iBug: I doubt this would actually get filtered out because `x not in reason.lower()` would make the capitalized letters `NS` never hit, no?

tripleee: Heh, thanks :-) I'll try to fix tomorrow.

PR #2391, opened 2018-08-30
[Pending Test] Detect if MS is down and disable it

iBug:
MS gets down from time to time and it throttles Smokey a lot because communication with MS runs on the main thread.

This approach does these:

- [X] Wrap-up `get` and `post` so they count failures
- [X] Wrap-up `get` and `post` so they return None during MS down
- [X] Mark flag `GlobalVars.metasmoke_down` when failurecount exceeds threshold, send warning to chat rooms when that happens
- [x] Replace existing code so they used the wrapped-up `get` and `post`
- [x] Integrate MS websocket into this counting
- [x] Quick command that enables/disables MS
- [ ] Prevent automatic reboot from resetting the flag

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -556,6 +556,23 @@ def errorlogs(count):
     return fetch_lines_from_error_log(count or 50)
 
 
+@command(privileged=True, aliases=["ms-down", "ms-up"], give_name=True)
+def metasmoke():
+    if alias_used == "metasmoke":
+        status_text = [
+            "metasmoke is up. Currect failure count (consecutive): {}".format(GlobalVars.metasmoke_failures),
+            "metasmoke is down. Currect failure count (consecutive): {}".format(GlobalVars.metasmoke_failures),
+        ]
+        return status_text[GlobalVars.metasmoke_down]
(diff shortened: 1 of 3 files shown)


iBug:
The last one, since there's `!!/ms-down` and `!!/ms-up` commands that allows manually setting MS state to down or up, as well as that we want to avoid rebooting during MS downtime and that MS downtime usually don't last for more than a few hours, it's safe to ignore the last checklist item.

Pending merge now.

chatcommands.py
```diff
@@ -557,6 +557,23 @@ def errorlogs(count):
     return fetch_lines_from_error_log(count or 50)
 
 
+@command(privileged=True, aliases=["ms-status", "ms-down", "ms-up"], give_name=True)
+def metasmoke(alias_used):
+    if alias_used in {"metasmoke", "ms-status"}:
```
j-f1: Could we make these aliases non-privileged so anybody can check the status?

iBug: Done in 7cb88502682d883862e42b17a670fd577287893a

PR #2419, opened 2018-09-15
Move "illuminati" and "vampire" pattern-matching email into findspam.py.

makyen:
illuminati is 85/0 (tp/fp)
vampire is 26/0 (tp/fp)

They've both been on the watchlist for 3 months. However, the `vampire` is being changed to `vampire?` in order to catch more recent examples.


iBug:
!!/approve

PR #2447, opened 2018-10-01
Use git porcelain for branch checking

### LINKED ISSUE #2399: We shouldn't parse output of git-status
Currently in `globalvars.py`, we parse the output of `git status` and check for `HEAD detached`, to determine whether we're on a branch or not.

This is a bad practice because the output of `git status` is meant to be human-readable, not to be parsed by programs. We should probably use a program-friendly interface to determine whether HEAD is detached from branch.

MartijnBraam:
This fixed #2399

This makes the branch check use `git symbolic-ref -q HEAD` instead of `git status`, if in a non-detached state the result of symbolic-ref is `refs/heads/master`, if the head is in a detached state it exists with exit code 1.

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -24,11 +24,14 @@ def git_commit_info():
     return {'id': full_id[:7], 'id_full': full_id, 'author': author, 'message': message}
 
 
-def git_status():
+def git_ref():
     try:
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Thanks for contributing!

PR #2452, opened 2018-10-03
Fixed regex causing FP on white spaces

kishan-1234:
Added check in regex to ignore whitespaces

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1412,7 +1412,7 @@ class FindSpam:
          'body': True, 'username': False, 'stripcodeblocks': True, 'body_summary': False, 'questions': False,
          'max_rep': 1, 'max_score': 0},
         # Link text without Latin characters
-        {'regex': u">[^0-9A-Za-z<'\"]{3,}</a>", 'all': True,
+        {'regex': u">\s*([^\s0-9A-Za-z<'\"]\s*){3,}</a>", 'all': True,
          'sites': ["ja.stackoverflow.com", "ru.stackoverflow.com", "rus.stackexchange.com", "islam.stackexchange.com",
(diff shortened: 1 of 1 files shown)


iBug:
Looks good. You may want to escape the backslashes in the string so Python parses them as blackslashes instead of parsing `\s` as an escape sequence.

kishan-1234:
@iBug I didn't get your point regarding escaping the backslashes. Can you please explain further, it would be very helpful if you could give an example where regex fails in this regard.

iBug:
There's nothing wrong with the regex, but that it's a Python string:

```python
>>> print("\s")
\s
>>> print(u"\s\n")
\s

>>> print(u"\\s\\n")
\s\n
```

While it doesn't matter here, it's better to escape the backslashes.

kishan-1234:
Thanks for explaining @iBug  !!

If the regex is correct, escaping the backslash in the input string might require code changes in some different module/location right?. Please correct me if am wrong

ArtOfCode-:
The easy way to do what @iBug is suggesting is to change `u"..."` (unicode string) to `r"..."` ('raw' string). All strings in Python 3 are automatically unicode strings, so you don't need the `u`, and adding the `r` means you don't need to escape backslashes (which is what's done elsewhere in the file, if you want some examples of that).

iBug:
@ArtOfCode- 

```
u"...['\"]+\s*"
       ^^
```

ArtOfCode-:
@iBug Not sure I understand what you're getting at there.

kishan-1234:
Thanks a lot @ArtOfCode- !!

Done the change.

iBug:
Looks good enough now.

PR #2466, opened 2018-10-12
ASN watch

tripleee:
This implements watching for Autonomous System Numbers via the Team Cymru public DNS mapper.

The ASN watchlist is populated with ASNs from a few recent fake tech support spam posts.

Halflife has been logging ASNs for several months now in the Charcoal Test room; I have been vaguely planning to look into the results for some time now, and so, here this is now.

I also had bare IP address watching implemented but it seems less useful at this point so I backed it out for now.

The code has the potential to make SmokeDetector slow but on my laptop, the additional overhead for the added DNS queries was insignificant. I have run a test over 600 recent posts (couple of hours of API feed I think; did it really take that long?) and it added on the order of 0.1 seconds per link I think.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -682,7 +682,9 @@ def watched_asn_for_url_hostname(s, site):
         s, site,
         # Watched ASN list
         [
+            '3842',    # RAMNODE - RamNode LLC, US
             '8100',    # ASN-QUADRANET-GLOBAL - QuadraNet Enterprises LLC, US
+            '34788',   # NMM-AS D - 02742 Friedersdorf Hauptstrasse 68, DE
             '40676',   # AS40676 - Psychz Networks, US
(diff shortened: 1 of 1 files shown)


tripleee:
Scanned another 400 or so with the latest tweaks (I had a silly parsing bug which would prevent the rule from triggering; discovered by experimenting with a much larger set of ASNs which I expected to get many FPs but nothing happened at first ...)

Here are some (US-only) ASNs as reported by Halflife, to give you an idea:

https://chat.stackexchange.com/search?q=us%2Fus&room=65945

tripleee:
First FP, not a horribly alarming one but still an FP.

https://chat.stackexchange.com/transcript/message/47142541#47142541

tripleee:
Another 1,000 scanned with 1 TP, no FPs.  I'll push a couple of minor additional ASNs which were included in the latest run.

PR #2476, opened 2018-10-17
Updated

Ayushraj1:
please merge the modifications


iBug:
I probably won't merge this until it's clear to me what it's intended for. Can you please explain your changes?

makyen:
Based on this user's profile, this looks like someone that is shotgunning PRs in order to qualify for Hacktoberfest. However, the quality level of some of their PRs is above that which I've previously seen. Other PRs, not so much.

The phone number they added has no hits in MS in titles or bodies.

It's unclear why several sites should be added with explicit queue depths of 1.

angussidney:
Additionally labelling as [invalid] until an explanation is given

PR #2484, opened 2018-10-23
Fix string formatting error

42B:
https://github.com/Charcoal-SE/SmokeDetector/blob/8b5d6e5cc1e728d8d2e78c053e4def548a9d07e5/test/test_findspam.py#L108

```python
>>> "Expected {1} on {}".format('body', 'expected_spam')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
ValueError: cannot switch from manual field specification to automatic field numbering

>>> "Expected {1} on {0}".format('body', 'expected_spam')
'Expected expected_spam on body'
```


ArtOfCode-:
Thanks!

PR #2485, opened 2018-10-24
Use requests' params keyword argument instead

42B:
* http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls


ArtOfCode-:
This is really useful, thanks!

42B:
@ArtOfCode- 

No problem!

Issue #2256, opened 2018-06-14
Canonical procedure when one of our dependency packages has rolled out a backward-incompatible update?


iBug:
From time to time, a PyPI package of Smokey's dependency gets a new, backwards-incompatible version published. This creates some hard time for us like it just happened some time ago (see #2251 and #2253).

On one hand, we can't update the code directly to adapt to the new version as adapted code will not run correctly on instances with the old version of the package.  
On the other hand, server owners can't update the package directly before adaption code is deployed.

This requires a developer and a server owner to have time to do the work *simultaneously*. While some of us are able to handle this, it's much too complicated in a general case.

So, is there a "canonical", or at least, better solution when this happens again?

j-f1:
Is there a way to use some sort of dependency lockfile to avoid accidental upgrades, then install from that whenever the instance pulls? From my experience with Yarn, a package manager for JS that works like that, installs where nothing needs to change are pretty quick.

tripleee:
`pip freeze` does that I believe. But blocking accidental upgrades isn't the beef here; managing upgrades we want without breaking anything is what this ticket is about.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

j-f1:
@tripleee Those are the same thing — if you control the exact version of each dependency on each instance, and can update it with a commit, you can make the backward-incompatible changes and upgrade the dependency in one commit, which will cause the update to be installed on pull.

PR #2491, opened 2018-10-25
Provide a mechanism to initiate `pip` updates from within Smokey

teward:
Currently, whenever we have a requirements file change, we ping all maintainers and require them to update the Smokey instances manually, likely due to system libraries or system `pip` or such being in use.

This pull request adjusts `nocrash.py` to utilize an exit code (`20`) that does two things:  (1) Pulls the `deploy` branch to update it, and then (2) calls a `pip` update within userspace of both the `requirements.txt` and the `user_requirements.txt` files.

This will permit us to properly get new requirements upgrades done automatically without any major requirements or mass-pinging or mass manual updating of instances each time.

**This will require all running SmokeDetectors to be manually restarted, however, because of the changes to `nocrash.py`**

We also implement the `pip_update` and `pip-update` alias for Chat Commands to call `os._exit(20)` which will be caught by `nocrash.py` and initiate the `pip` upgrade process.  If for any reason `pip` calls fail, the Smokey instance will not automatically restart because we could be in a dangerous 'half-upgraded requirements' state which could explode Smokey.  However, provided `pip` works as intended and updates go smoothly, SmokeDetector should automatically restart.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -686,6 +686,7 @@ def pip_update():
     else:
         os._exit(20)
 
+
 @command(privileged=True, aliases=['pip-force'])
 def pip_force():
     """
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -674,7 +674,7 @@ def master():
     os._exit(8)
 
 
-@command(privileged=True, aliases=['pip-update'])
+@command(privileged=True)
 def pip_update():
     """
     Initiate `pip` requirement updates in userspace, or warn if we're in a venv.
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -25,7 +25,7 @@ from html import unescape
 from ast import literal_eval
 # noinspection PyCompatibility
 import regex
-from helpers import only_blacklists_changed, only_modules_changed, log, expand_shorthand_link, reload_modules
+from helpers import only_blacklists_changed, only_modules_changed, log, expand_shorthand_link, reload_modules, is_venv
 from classes import Post
 from classes.feedback import *
 
@@ -677,10 +677,21 @@ def master():
 @command(privileged=True, aliases=['pip-update'])
 def pip_update():
     """
-    Initiate `pip` requirement updates in userspace.
+    Initiate `pip` requirement updates in userspace, or warn if we're in a venv.
     :return: None
     """
-    # Exit code 20 triggers the pip calls in nocrash.py
+    if not is_venv():
+        raise CmdException("Smokey is not running in a Python virtualenv.  Updating may cause other unrelated things "
+                           "to break.  Use pip-force to force the pip update anyways.")
+    else:
+        os._exit(20)
+
+@command(privileged=True, aliases=['pip-force'])
+def pip_force():
+    """
+    Force `pip` upgrades regardless of env.
+    :return: None
+    """
     os._exit(20)
 
 
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -683,8 +683,8 @@ def pip_update(alias_used="pip-update"):
     if alias_used == 'pip-update-force' or is_venv():
         os._exit(20)
     else:
-        raise CmdException("Smokey is not running in a Python virtualenv.  Updating may cause other unrelated things "
-                           "to break.  Use pip-force to force the pip update anyways.")
+        raise CmdException("SmokeDetector  is not running in a Python virtualenv. Updating may cause other unrelated "
+                           "things to break.  Use `pip-update-force` to force the update anyways.")
 
 
 # noinspection PyIncorrectDocstring,PyProtectedMember
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -674,6 +674,16 @@ def master():
     os._exit(8)
 
 
+@command(privileged=True, aliases=['pip-update'])
+def pip_update():
+    """
+    Initiate `pip` requirement updates in userspace.
+    :return: None
+    """
+    # Exit code 20 triggers the pip calls in nocrash.py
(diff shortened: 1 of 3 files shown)


AWegnerGitHub:
I was encouraged to raise this issue here instead of the chat room to further the discussion. 

---

I disagree with this proposal, unless there is a mechanism that warns the user we will be clobbering their development environment and allows them to cancel the upgrade so they can address their other applications.

If a user is not running Smokey in a virtual environment, I don't think it's safe for us to assume we can update their packages. Smokey is not guaranteed to be the only application running on the system. It's also not guaranteed to be the only application using a specific library. We have several popular libraries in Smoke Detector:

 - requests
 - pytest
 - beautifulsoup4
 - coverage
 - flake8

If we unexpectedly update these without giving the user both a warning and the option to cancel the upgrade and check their other applications, we are potentially making a big problem for them.

ArtOfCode-:
If we can detect whether or not Smokey is running in a venv, we can add a config option - `warn_before_upgrade` - to allow runners to choose whether or not they get warned before package upgrades. This allows for:

 - Smokey being run in a venv - we can just upgrade packages without warning anyone
 - Smokey outside a venv, but runner is okay with unattended upgrades - no need to warn anyone, `warn_before_upgrade` is set to False
 - Smokey outside a venv, but runner wants to okay upgrades - `warn_before_upgrade` set to True, and the runner gets a warning about any requested upgrades.

teward:
We could also di this in a different way.

test if we're in a venv (see https://github.com/Charcoal-SE/SmokeDetector/pull/2491/commits/1d052b35d7ed497081e5598067bd0025a9adabfc as an example), and in the ChatCOmmand call, set it so that if we're not in a venv it doesn't work unless we provide a 'force' argument (`pip-force`).  Similar to how we have watchlist/blacklist forces.

The venv checking code is identical, and i'm working on the 'force' bits now.

teward:
https://github.com/Charcoal-SE/SmokeDetector/pull/2491/commits/7a476c6fef1227f262a78b4f823323bef41c0e01 has an attempt at this, though it may not behave entirely as expected in some circumstances.

nocrash.py
```diff
@@ -143,6 +148,38 @@ def error(message):
...
+            warn('Not pulling updates; we are on Windows')
+
+        count = 0
+        crashcount = 0
+
+        log('Initiating automated `pip` upgrade calls within userspace.')
+        if is_venv():
+            command = "python3 -m pip install --upgrade -r "
+        else:
+            command = "python3 -m pip install --upgrade -user -r "
```
ArtOfCode-: You missed a hyphen

```suggestion
            command = "python3 -m pip install --upgrade --user -r "
```

chatcommands.py
```diff
@@ -674,6 +674,16 @@ def master():
     os._exit(8)
 
 
+@command(privileged=True, aliases=['pip-update'])
```
ArtOfCode-: You don't need the `pip-update` alias - that's the name of the command, so it's already there.

```suggestion
@command(privileged=True)
```

nocrash.py
teward: This was fixed in 7a476c6 and 5d69e9b because I caught it in my other code changes.

chatcommands.py
```diff
@@ -674,6 +674,28 @@ def master():
...
+    :return: None
+    """
+    if not is_venv():
+        raise CmdException("Smokey is not running in a Python virtualenv.  Updating may cause other unrelated things "
+                           "to break.  Use pip-force to force the pip update anyways.")
+    else:
+        os._exit(20)
+
+
+@command(privileged=True, aliases=['pip-force'])
```
ArtOfCode-: Suggestion: add a `pip-force` alias to the `pip_update` command def, and check venv based on `alias_used` - our blacklist commands do this, if you want an example of how to use it.

teward: Working on this...

teward:
@ArtOfCode- 16373e5 and 2fd72ad should address this.  They both should address the concerns you had.

teward:
@AWegnerGitHub any concerns leftover with this feature change and the commits as is?  I think we addressed the 'needs a warning' bit, and the venv/not-venv concern with the changes...

AWegnerGitHub:
I am still uncomfortable with this change. This is different than updating Smokey, itself, automatically. This is updating the user's entire development environment. The warning is a good step, but the problem here is that the warning really isn't enough.

> SmokeDetector  is not running in a Python virtualenv. Updating may cause other unrelated things to break.  Use `pip-update-force` to force the update anyways.

By using `pip-update-force` we are allowing anyone with permissions to change the instance runner's environment. No one other than the instance runner knows what their environment is. Only the runner can reasonably say "Yup, this isn't going to cause a problem." Everyone else just wants Smokey to run correctly, so if `pip-update` fails, they'll do the `-force` and happily move on. SmokeDetector should work then, but the underlying environment was just changed and it's reasonable to assume the instance runner doesn't even notice (because they weren't in CHQ at the time) and now their environment has changed.

iBug:
@AWegnerGitHub I'm having the same idea. What about my proposal above about manually enabling this, so instance owners who run Smokey in a venv can safely enable this?

teward:
Because of the unending controversy over this, I'm closing this.  However, we need to keep in mind if we're going to make **major** requirements changes requiring manual intervention, we need a better way that 'pinging the runners in CHQ' to get the "You need to update this" message out.  (cc @Undo1 @ArtOfCode- )

iBug:
The undergoing discussion is essentially a duplicate of #2256 , by the way.

PR #2494, opened 2018-10-26
Resolve MS post link when blacklisting users

### LINKED ISSUE #1450: Resolve Metasmoke URL in 'blacklisted for {}'
When a user is blacklisted, the Metasmoke report references the post which caused the user to be blacklisted in the first place.

Example from https://metasmoke.erwaysoftware.com/post/100562:

    Blacklisted user - blacklisted for //stackoverflow.com/a/48091715 (https://m.erwaysoftware.com/posts/by-url?url=//stackoverflow.com/a/48091715) by the metasmoke API

There are a few usability problems and a performance problem which would be nice to work around.

* The display is very wide. On my screen right now, there is a fugly horizontal scroll bar and the visible text ends soon after the opening parenthesis.
* The protocol-less non-URL often points to a deleted post, so copy/pasting that into the Location par in your browser and adding a https: prefix often doesn't work anyway.
* Copy/pasting the long URL reveals that it requires Metasmoke to resolve from the original post's URL to a Metasmoke URL and you don't know where it will end.

Resolving the Metasmoke URL when generating this notice would reduce the perceived latency by reducing the number of interactive API calls (though it will probably end up spending more resources in noninteractive use -- it's a fairly safe prediction that nobody visits most of these URLs more than once). Especially when Metasmoke is already under some load, the interactive use is a bit sluggish as it is, whereas slightly slower batch processing probably will not bother anyone in the grand scheme of things, even if it is objectively speaking a bit wasteful.

By way of example, this:

    Blacklisted user - blacklisted for https://metasmoke.erwaysoftware.com/post/100453 (//stackoverflow.com/a/48091715) by the metasmoke API

is both shorter and friendlier for copy-pasting, and has the additional minor feature that you can see roughly how old the blacklisting incident is by comparing the Metasmoke ID in the current URL to the one in the link (in this case, 100562 vs 100453, so probably not more than a couple of days old, as we currently get on the order of 100 new Metasmoke reports per day).

iBug:
This should resolve #1450 

I added `timeout=5.000` hoping that MS downness will not throttle Smokey too badly.


ArtOfCode-:
What does this do for us? Looks like it changes the SE post link to an MS post link, which... I don't know if we want? It'd be cool to have both, if we could.

iBug:
@ArtOfCode- Oops, I didn't realize that. I guess it needs some expansion.

iBug:
@ArtOfCode- Ah yes. Can you add an option to MS so when a post has multiple records, it generates a redirect to the latest occurrence instead of showing a "disambiguation list"? The option could be a GET parameter.

iBug:
@ArtOfCode- How does it look now?

PR #2496, opened 2018-10-26
Effectively blacklist blacklisting disk-partition.com

tripleee:
disk-partition.com has been in and out of the blacklist repeatedly. It should not be blacklisted.

Add disk-partition.com to GlobalVars.valid_content so as to trigger the "too broad" check if someone attempts to blacklist it again.

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -151,7 +151,7 @@ class GlobalVars:
     # Miscellaneous
     log_time_format = config.get("log_time_format", "%H:%M:%S")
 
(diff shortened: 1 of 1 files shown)


iBug:
Just a note here: That `valid_content` was intended for accidental blacklisting of common phrases (like a single dot - that means everything) since Rapid Reload bypasses CI and can cause potential mess. If there's need to prevent more stuff from being blacklisted, we should probably find a new approach (IMO it's a decent solution pro-tempore).

PR #2504, opened 2018-10-29
Identify "boneheaded" nameserver setups: NS1 and NS2 are same IP as host itself

tripleee:
These have been discussed informally in chat from time to time, and we know of a few cases where spammers are using this nameserver setup specifically.

Being your own nameserver is technically a bad decision, though I have seen amateurs set up similar configurations simply because they did not understand why you want to have multiple independent name servers. But this has come up repeatedly with spam domains now, and in my informal testing, I have not seen any false positives so far (though this is after only about 500 posts scanned).

The PR also involves a refactoring effort which removes some code duplication in `findspam.py` between the ASN and NS lookup paths, and adds a simple cache for reusing recent results. (The cache is set up with completely ad-hoc parameters; if we were serious, we could run this for a few days and take measurements to establish how many entries it would actually make sense to cache and how often to purge the cache.)

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -24,7 +24,9 @@ from helpers import log
 from globalvars import GlobalVars
 import blacklists
 
+
 TLD_CACHE = []
+DNS_CACHE = dict()
 LEVEN_DOMAIN_DISTANCE = 3
 SIMILAR_THRESHOLD = 0.95
 SIMILAR_ANSWER_THRESHOLD = 0.7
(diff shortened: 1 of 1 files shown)


tripleee:
For what it's worth, I have now scanned another 500+ posts with the exact code in this PR and no false positives still. It feels like the DNS cache actually reduces the overhead considerably compared to the current code base, but perhaps I'm subject to some serious confirmation bias.

I don't have a rig to leave it run unattended on, but I'll resume testing tomorrow if this is not already merged by then.

PR #2507, opened 2018-10-30
Update ws.py

naveenuchiha:


iBug:
Hi! Welcome to SmokeDetector. While we appreciate your contribution, would you mind taking a look at our continuous integration and at least pass the checks?

Also, can you please describe why you're removing those blank lines? It seems to violate our code conventions.

Before we understand the intention behind this PR, we have to label it as [invalid]. Thank you for your understanding.

PR #2512, opened 2018-10-31
Smokey Pickle Transfer Protocol

### LINKED ISSUE #2511: Smokey pickle transfer protocol via chat (SPTP)
Due to an earlier development failure, Smokey dumped the whole output of `git log` in message 47386004 and 3 other messages (all moved to Charcoal Test). It resulted in a message more than 3 MiB in length.

It indicates that there's no length limit or at least "very high" of multiline messages. I think this can be utilized (abused) to transfer pickles or other synchronized data like blacklisted users, ignored posts, etc. A chat command will dump the pickle directory or selected ones after some simple packaging and encoding, for example `tar zcf - | base64`, and post the resulting string to Charcoal Test. Another chat command will load the pickle from a given chat message and restore pickled data, and probably merge with its local data as well.

Of course, `tar` isn't the only solution. We can even pack the data up in Python and make it into a single file before base64-ing, but that's technical implementations and the base is the same (via chat).

What's your idea about this approach?

iBug:
Implements and closes #2511, the Smokey Pickle Transfer Protocol

### Data

It should pack selected data into a `dict` and pickle that dict, then compress the pickled data to reduce size, and base64-encode the compressed data so it's suitable to be posted as a chat message. A header line and an ending line similar to PGP armored data are also added. A loader function does the reverse thing, with some data validation in place.

Example data (try to break the CI test and it'll show up in stdout):

```
-----BEGIN SMOKEY DATA BLOCK-----

eNpVjz1OAzEQhfOz/BOh3MAdKVDEbgjJFhSIAkU+AUUUeXfH6xEbG63HiShAFCiV
O8zVOAo1m6RipCnme6On9z6i77t2azdvYeBPFksgUQgSYXtGhEsIj/c/v18Xq893
f1yZXBAaHfzTg9GE2hln2UwTlPVOGLKZZK/GXdbALADqkpFCe8XyGgQBEyw3WmLJ
JFbA1kiKWaGbX3JSMtQMKfhuDavgByJNiySB0TQdJ9Miza7jOE7GtyIdT+BGTmEi
s1FWxJPgjyrQJSm7C93PKpE/V2gJioWzUNvAO76/Vkjwn7Z8D0tt6oa8GEt7og2h
xH3NLXFOnTa2vM07vLsJPOIH/HATnDqbB3XebG8e3PAPT7x00A==

-----END SMOKEY DATA BLOCK-----
```

Further improvement may include a merging process so local data isn't overwritten directly.

Currently included data:

- blacklisted / whitelisted users
- ignored posts
- notification configuration (this is what user complains the most about not being synced)

This list can be easily expanded by adding entries to `SmokeyTransfer.ITEMS` with the format shown in the comment.

### Transmission

Two new chat commands. One of them tells Smokey to post the generated data as a chat message somewhere ([Charcoal Test](https://chat.stackexchange.com/rooms/65945) or [Socket Science](https://chat.stackexchange.com/rooms/84821), current implementation uses the latter), and the other tells Smokey to load data from a particular chat message. To keep a baseline of security, Smokey will reject to load from messages that are not posted by itself, though this can be extended to trusted users if needed (hand-crafting those messages? well do it if you can `:)`)

`!!/dump-data` dumps the data and posts a chat message to all rooms with role `dump`, and `!!/load-data <message ID>` loads data from message with the given ID. Proper check is in place before loading.

### More ideas

- Provide a way to merge data instead of overwriting existing data (will do this later)
  - Potentially we'll have trouble really deleting something (like removing a user from blacklist) if the merge method can't resolve deletion
- Automate the dumping and loading (very difficult)
- CI tests for these functions, where applicable
- Create a CLI tool for manually analyzing pickles and SPTP data
  - Extend the message owner check to allow loading from messages posted by authorized users

More TBA?

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1591,12 +1591,22 @@ def feedback(msg, post_url, feedback):
 
 @command(privileged=True, aliases=['dump-data'])
 def dump_data():
-    return NotImplemented
+    s = SmokeyTransfer.dump()
+    tell_rooms_with('dump', s)
+    return None
 
 
 @command(int, privileged=True, aliased=['load-data'])
 def load_data(msg_id):
-    return NotImplemented
+    msg = get_message(msg_id)
+    if msg.owner.id != 120914:  # TODO: implement an is_self() in chatcommunicate, don't use magic numbers
+        raise CmdException("Message owner is not SmokeDetector, refusing to load")
+    try:
+        s = msg.content_source
+        SmokeyTransfer.load(s)
+    except ValueError as e:
+        raise CmdException(str(e)) from None
+    return None
 
 
 #
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1589,6 +1589,16 @@ def feedback(msg, post_url, feedback):
     raise CmdException("No such feedback.")
 
 
+@command(privileged=True, aliases=['dump-data'])
+def dump_data():
+    return NotImplemented
+
+
+@command(int, privileged=True, aliased=['load-data'])
+def load_data(msg_id):
+    return NotImplemented
+
+
 #
 #
 # Subcommands go below here
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1596,7 +1596,7 @@ def dump_data():
     return None
 
 
-@command(int, privileged=True, aliased=['load-data'])
+@command(int, privileged=True, aliases=['load-data'])
 def load_data(msg_id):
     msg = get_message(msg_id)
     if msg.owner.id != 120914:  # TODO: implement an is_self() in chatcommunicate, don't use magic numbers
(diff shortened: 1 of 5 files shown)


iBug:
Should be good to merge now. The "merge data" functionality, as said in the paragraph "further improvement", is postponed (by me).

ArtOfCode-:
Should we have some way of doing this automatically? Perhaps when an instance starts for the first time, it should drop a request in chat for the active to send its pickles over; the active could also dump its pickles before dying. That way, we don't have to rely on manual sync of pickles, where missing it once puts us permanently behind.

iBug:
@ArtOfCode- Good idea, but I'm not clever enough to figure out any approach `:(`

PR #2514, opened 2018-10-31
Bad word

krishyadav007:


iBug:
Hi! Welcome to our project, SmokeDetector. Thanks for your contribution first, but can you give some explanation why you're adding this as a "bad keyword"?

It has 4 hits in out database, and [3 hits][1] on the Stack Exchange search, giving a (visible) ratio of 57%, which makes it not qualify as a bad keyword, per our [blacklisting guidelines][3].

Before we what you're doing here, we have to label this PR as [invalid]. Thank you for your understanding.

  [1]: https://stackexchange.com/search?q=%22pagal%22
  [2]: https://metasmoke.erwaysoftware.com/search?body=pagal
  [3]: https://github.com/Charcoal-SE/SmokeDetector/wiki/Guidance-for-Blacklisting-and-Watching

PR #2521, opened 2018-11-05
Added K.Davis to SOCVR permissions

K-Davis1:
See https://chat.stackexchange.com/transcript/message/47490030#47490030


makyen:
As an RO in SOCVR, adding K.Dᴀᴠɪs is approved.

PR #2533, opened 2018-11-09
Move Smokey back to RPG General Chat

spacemonaut:
During a concentrated June/July spam wave, Smokey was moved to a separate overflow chat from the main RPG General Chat room. Now that the spam wave has died down we can move Smokey back to RPG General Chat.

(I'm one of the RPG.SE diamond moderators.)


tripleee:
!!/approve

PR #2534, opened 2018-11-11
autoflagged command update filter & use first report

makyen:
Update the filter used for the /posts/urls endpoint when fetching data for the `autoflaged` command.
Adjust the logic to use the oldest report for the URL, which is the only one that will be autoflagged.


ArtOfCode-:
!!/approve

PR #2537, opened 2018-11-14
Add IP blacklisting and watch

tripleee:
The code in this PR looks up the IP address for every host in a post, then checks if the IP address we get back is one we have on a list, similarly to how the ASN and NS checks already work.

This is a rough proof of concept, but I'm hoping it can be accepted as a baseline for future work in this direction.

There is some fugly code duplication between the IP and ASN lookups in particular; the code should arguably be refactored and modularized to have a single "loop over URL hosts" function with different options and parameters depending on what you are looking up.

I have blacklisted two IP addresses and added a bunch of IP addresses to watch based on completely ad-hoc exploration of recent Halflife reports.  Briefly speaking, [search for an AS and take note of recurring IP addresses](https://chat.stackexchange.com/search?q=26496&Room=65945&page=2&pagesize=50&sort=newest) then maybe search for each of those in turn. I focused on big providers which we cannot just plonk in the ASN watch list, though this could obviously be repeated for many other scenarios.

As a brief test, I have scanned some 600 posts during the last hour or two (one hit on a spam which was reported for other reasons too; watched IP match on `goldencondor.org`; no false positives).


tripleee:
Having the lists statically in the code is obviously a wart. Maybe with the SQLite (or some variation of that initiative) we could make these easy to manage from chat, and perhaps via some sort of API.  I'm thinking SmokeDetector (or perhaps Halflife) could do the bulk of the routine work which we currently perform manually.

ArtOfCode-:
What sort of hit rate is this getting?

tripleee:
The hit rate so far is low, and I don't expect it ever to be very high; but the watchlist and the blacklist added in this PR are still very small, and should certainly be expanded if we get this into production.

This is mainly a fallback for when you can't watch an entire AS but still have something reasonably specific within that AS that you would like to get at.

I'm thinking down the line the watch functionality could be automated (add sites at a certain TP/FP threshold) sort of how the "blacklisted user" works now to help us find new incidents involving the same resource.

tripleee:
Rebased to change some commit messages (misleadingly said "Blacklisted" when the IPs were actually "Watched")

tripleee:
Erm, uh, looks like I botched the rebase :-/

tripleee:
Synced with {{master}} and force pushed with the history fixed up properly.

tripleee:
There's a lot of stuff to merge. I'll probably try to rebase this to branch off the current {{master}}. Please ping me if you don't see signs of life here.

RobCoding:
+1 for this PR, 👍 - Regarding this conversation: https://chat.stackexchange.com/transcript/message/48454626#48454626

tripleee:
Weird wacky whitespace fallout from merge + rebase ... Hang on, one more rebase.

PR #2552, opened 2018-11-25
Added EJoshuaS to CHQ Privileges

K-Davis1:
He already has privileges in SOCVR and has experience with the commands.


Undo1:
!!/approve

PR #2553, opened 2018-11-25
Update rooms.yml

superplane39:


rooms.yml
```diff
@@ -481,6 +481,7 @@ meta.stackexchange.com:
       - 310369   # quartata
       - 350567   # iDebug (iBug on SO)
       - 377214   # Sonic the Anonymous Hedgehog (gparyani elsewhere)
+      - 163529   # Shree
```
ferrybig: It looks like you added lots of spaces after your name, is there any reason for that?

iBug:
@ferrybig If someone is already privileged anywhere, or better yet, has been active since, we are often happy to grant them in other places (except SOCVR and SOBotics), so no need for extra reviews.

PR #2558, opened 2018-11-30
Refactor findspam.py

iBug:
Since we have a lot of similar objects in `findspam.py` (the rules, which are currently individual dictionaries), they're best wrapped into a class.

Some time earlier I already wrapped chat commands into a class (4629191ad3ca9af6156e740f1ede3a484861138f), but this time it's going to provide more maintainability than the previous one.

- Rules can have default values for attributes and a more unified handling
- Any function can be wrapped into a rule with a decorator, so adding a function-based rule only requires adding one block of code, not two. Currently you need to define a function and then add a dict into `FindSpam.rules`
- Default values for rule properties, like default "max rep", similar to default arguments for the chat command decorator. This keeps things clean and minimal
- Extending rule creation is possible (not implemented yet), like creating a new rule based on an existing rule, or templating
  - Idea: Maybe we can now separate post-checking logic and the actual rules? Like moving all those `create_rule`s to a separate file, `rules.py`
- More waiting to be discovered

The new way to create a spam check is very simple and easy

For regex check, simply call `create_rule` at the end of `findspam.py` like this:

```python
create_rule("reason", "regex")
```

All other settings like `sites` or `max_rep` can be supplied as keyword arguments to `create_rule`. For example, this will create a rule with max rep being 30 and max score being 1, and it won't check answers.

```python
create_rule("reason", "regex", max_rep=30, max_score=1, answer=False)
```

Note that `questions` and `answers` have been renamed to their singular forms (`question` and `answer`).

To create a functional check, use `create_rule` as a decorator, similar to chat commands:

```python
@create_rule("reason", question=False)
def new_check(s, site):
    return True, "all answers are spam"
```

The default settings for `create_rule` is:

```python
all = True
sites = []
title = True
body = True
username = False
body_summary = False
stripcodeblocks = False
max_rep = 1
max_score = 0
question = True
answer = True
disabled = False  # Create rule but don't add to rule list
```

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -141,7 +141,7 @@ class PostFilter:
         if (post.is_answer and not self.answer) or (not post.is_answer and not self.question):
             # Wrong post type
             return False
-        elif self.all_sites != post.post_site in self.sites:
+        elif self.all_sites == (post.post_site in self.sites):
             # Post is on wrong site
             return False
         elif (post.owner_rep > self.max_rep) or (post.post_score > self.max_score):
--- a/findspam.py
+++ b/findspam.py
@@ -267,7 +267,7 @@ class Rule:
 # what if a function does more than one job?
 def create_rule(reason=None, regex=None, *, all=True, sites=[],
                 title=True, body=True, body_summary=False, username=False,
-                max_score=1, max_rep=1, question=True, answer=True, stripcodeblocks=False):
+                max_score=0, max_rep=1, question=True, answer=True, stripcodeblocks=False):
     if not isinstance(reason, str):
         raise ValueError("reason must be a string")
 
--- a/findspam.py
+++ b/findspam.py
@@ -349,6 +349,7 @@ class FindSpam:
 def create_rule(reason=None, regex=None, func=None, *, all=True, sites=[],
                 title=True, body=True, body_summary=False, username=False,
                 max_score=0, max_rep=1, question=True, answer=True, stripcodeblocks=False,
+                whole_post=False,  # For some functions
                 disabled=False):  # yeah, disabled=True is intuitive
     if not isinstance(reason, str):
         raise ValueError("reason must be a string")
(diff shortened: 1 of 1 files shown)


j-f1:
How do you feel about using a [`namedtuple`](https://devdocs.io/python~3.7/library/collections#collections.namedtuple) instead?

iBug:
@j-f1 Looks like a better alternative. I'll try it tomorrow.

iBug:
@j-f1 Not quite suitable here. Although it's going to be useful to some extexts, it's not pure data collection (bundled with methods) so namedtuple doesn't provide extra benefits. I'll stick to classes.

findspam.py
```diff
@@ -1118,7 +1441,12 @@ def toxic_check(post):
     return False, False, False, ""
 
 
-# noinspection PyUnusedLocal,PyMissingTypeHints
+if GlobalVars.perspective_key:  # don't bother if we don't have a key, since it's expensive
+    toxic_check = create_rule("toxic {} detected", func=toxic_check, whole_post=True, max_rep=101, max_score=2)
```
ArtOfCode-: Is the assignment to `toxic_check` used anywhere?

findspam.py
```diff
@@ -121,6 +121,272 @@
...
+        return result, why
+
+    @staticmethod
+    def match_info(match):
+        start, end = match.span()
+        group = match.group().replace("\n", "")
+        return "Position {}-{}: {}".format(start + 1, end, group)
+
+    @staticmethod
+    def match_infos(matches):
```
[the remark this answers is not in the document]
ArtOfCode-: Hm. Not sure I can think of a better name either... Not critical, can be changed later if someone does come up with one.

findspam.py
ArtOfCode-: Flake will complain at you for having an unused variable, though

iBug: @ArtOfCode- No it does NOT at module level because module variables can be imported. Local variables are useless outside the function so flake8 complains about those.

ArtOfCode-: Huh, TIL

ArtOfCode-:
Ready to merge?

iBug:
@ArtOfCode- Yes, please.

PR #2592, opened 2019-01-17
Rescan certains posts again after a short period

iBug:
See Charcoal-SE/metasmoke#565 for more info.

Rescans a post instead of handling scan results directly if it meets certain criteria:

- is a question
- is on WordPress Development
- caught for "bad keyword in body"
- has an aggregate reason weight of no more than 100 for its initial version
- during spam hour (extended to 2 AM to 2 PM UTC - 2 hours before and after typical spam hour)


iBug:
Closing stubs

PR #2597, opened 2019-01-22
Run network-based spam checks in an alternate thread

iBug:
We have a few network-based checks like "bad NS", "bad ASN", "problematic NS config" and Google Perspective. Network is usually slow compared to calculation tasks like RegEx matching. As we're currently applying all checks on a post using a single `for` loop, the overall time taken to scan a post would likely be affect a lot by network instability. Moving all these network-based checks to an alternate thread should give the overall time taken a boost, assuming all these network request functions releases the GIL when making the requests.

I'll pull this on my backup instance and run it for a few days to gather some statistics and see how it actually performs. The code in this PR should be always good to merge once CI passes.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -301,21 +301,24 @@ class FindSpam:
         result = []
         why_title, why_username, why_body = [], [], []
 
-        def alternate_job():
+        def alternate_job(rule):
             nonlocal result, why_title, why_username, why_body
-            for rule in FindSpam.rules_alt:
-                title, username, body = rule.match(post)
-                if title[0]:
-                    result.append(title[1])
-                    why_title.append(title[2])
-                if username[0]:
-                    result.append(username[1])
-                    why_username.append(username[2])
-                if body[0]:
-                    result.append(body[1])
-                    why_body.append(body[2])
-        alt_thread = Thread(target=alternate_job)
-        alt_thread.start()
+            title, username, body = rule.match(post)
+            if title[0]:
+                result.append(title[1])
+                why_title.append(title[2])
+            if username[0]:
+                result.append(username[1])
+                why_username.append(username[2])
+            if body[0]:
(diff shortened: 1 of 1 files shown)


iBug:
Aha, doesn't really need actually running - just change some CI tests and see how it improves.

iBug:
Results on CI tests seems consistent and promising.

The initial version that creates one extra thread for all "alternate" checks (NS, ASN, NS config, no Perspective on CI) made a differece of 2 secs on CircleCI (12s vs 14s) and less than 2 secs on Travis CI (22.5s vs 24s).

The enhanced version further decreased time by 1 sec on CircleCI (11s vs 14s), but instead slowed a bit on Travis CI (23s vs 24s).

I noticed that the time taken on Travis CI has a greater variation (less stable), and that may account for a part of the difference of the enhancement.

In real cases, since real SE posts often contains more links than these CI tests, the time decrease may be further visible, depending on host environment (CPU and network latency).

Final poll: One thread for all alt checks or one thread for each?

angussidney:
Not sure how easy this would be to do in practise, but could we consider trying both sets of code simultaneously on two test instances, and see which has the higher scan rate when applied to real-world data? I think we already have most of the monitoring tools we would need to try such an experiment

tripleee:
The network checks take a lot of time so running them serially is going to be slow no matter what. One thread each would probably shave off much more than what you are currently seeing. On the other hand, the really proper solution to this would be to use `async` and poll for the results as they come back. (We discussed this in chat and tentatively concluded that it might be too hard.)

iBug:
Closing this for now. I'll learn `async` and re-do this.

iBug:
@triplee Just a small comment:

> (We discussed this in chat and tentatively concluded that it might be too hard.)

Fact has proven it's not that hard :P

PR #2619, opened 2019-01-31
Run network-based spam checks with asyncio

iBug:
This is an alternative approach to #2597 using Python 3.5 `async` keyword, an alternate event loop, as well as some other async-related libraries.

Well, true asyncio this time, not creating an extra thread for every call.

Now that this PR is complete (ready to merge), reviews and suggestions are welcome.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1038,7 +1038,7 @@ async def ns_for_url_domain(s, site, nslist):
     for domain in set(domains):
         ns = await dns_query(domain, 'ns')
         if ns is not None:
-            nameservers = set([server.target.to_text() for server in ns])
+            nameservers = set([result.host + "." for result in ns])
             for ns_candidate in nslist:
                 if (type(ns_candidate) is set and nameservers == ns_candidate) \
                     or any(ns.endswith('.{0}'.format(ns_candidate))
--- a/findspam.py
+++ b/findspam.py
@@ -975,7 +975,6 @@ def purge_cache(cachevar, limit):
             del cachevar[old]
 
 
-# TODO: get asyncdns working here
 async def dns_query(label, qtype):
     global DNS_CACHE
     qtype = qtype.upper()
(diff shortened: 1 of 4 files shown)


iBug:
FYI: I'm running a backup instance with a *terrible* network connectivity, to collect some timing data around this PR. I'll come up with those data later.

findspam.py
```diff
@@ -12,21 +12,30 @@
...
 import chatcommunicate
 
 from helpers import log
 from globalvars import GlobalVars
+from tasks import new_async_loop
 import blacklists
 
 
+# Cheap trick to create compatible code :)
+class dns:  # noqa: N801
```
ArtOfCode-: Why does this not have a PascalCase name?

tripleee:
By quick look, this seems promising. Eagerly expecting timing results and other observations.

iBug:
@tripleee I got the [raw timing stats][1] from my server (China, bad network) and the result looks as promising. But I had some difficulty trying to visualize those 180k numbers (read from [here](https://chat.stackexchange.com/transcript/message/48824545#48824545), note: one bad number manually removed before visualization). That could provide some good insights after all.

  [1]: https://gist.github.com/iBug/e8624d68dd60498eb1a67a776eb834cb
  [2]: https://chat.stackexchange.com/transcript/message/48824545#48824545

tripleee:
So the average slightly *increased* for the middle 60%? But shaving off more than 25% from the overall average is still nice. I guess it means it improves performance for the slowest messages, which isn't at all a bad thing. But are the savings substantial enough to defend the additional complexity?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

iBug:
@tripleee Your last sentence is the exact reason why I'm hesitating and not merging this directly. It's also important to know that the data is taken on a server located in China where international network connection is severely bad. Given that our production instances are all in US or Europe where there isn't ridiculously strict internet censorship (and therefore better connectivity), it's unlikely this will improve a lot, and thus not worth the extra complexity and overheads.

I myself think this is a good start and probably something good to be archived (not merged), though.

PR #2664, opened 2019-03-06
Add whitelist for ASN checks

bertiebaggio:
This creates a whitelist - ASN_WHITELISTED_WEBSITES - and modifies
the asn_for_url_host() check to shortcut if `hostname' is on the
whitelist.

There are a few sites which were causing fp's because they
happened to come under a watched ASN. Some will be used on
particular sites more than others (eg [thefreedictionary on
ELL/ELU](https://metasmoke.erwaysoftware.com/post/160036), docs.unity3d.com on Game Dev).

Mentioned in chat, Makyen suggested a simple static list:
https://chat.stackexchange.com/transcript/11540?m=49080625

Top 20 hosts watched with ~178 MS records between them at
time of creation.

ASN fp search (MS):
https://metasmoke.erwaysoftware.com/data/sql/queries/164-bad-asn-in-false-positives

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -60,10 +60,11 @@ WHITELISTED_WEBSITES_REGEX = regex.compile(r"(?i)upload|\b(?:{})\b".format("|".j
     r"jsfiddle\.net", r"codepen\.io", "pastebin"
 ] + [se_dom.replace(".", r"\.") for se_dom in SE_SITES_DOMAINS])))
 ASN_WHITELISTED_WEBSITES = ["docs.unity3d.com", "trac.ffmpeg.org", "bitcoincore.org", "ffmpeg.org",
-    "latex.codecogs.com", "www.advancedcustomfields.com", "name.com", "businessbloomer.com",
-    "wkhtmltopdf.org", "www.thefreedictionary.com", "ruby-doc.org", "www.site.com.br",
-    "www.ffmpeg.org", "unity3d.com", "test.ooo-pnu.ru", "swift.org", "site2.com", "rxweb.io",
-    "www.tenforums.com", "www.rhydolabz.com"]
+                            "latex.codecogs.com", "www.advancedcustomfields.com", "name.com",
+                            "businessbloomer.com", "wkhtmltopdf.org", "www.thefreedictionary.com",
+                            "ruby-doc.org", "www.site.com.br", "www.ffmpeg.org", "unity3d.com",
+                            "test.ooo-pnu.ru", "swift.org", "site2.com", "rxweb.io", "www.tenforums.com",
+                            "www.rhydolabz.com"]
 # ^^ top 20 FP hosts that get reported due to 'bad ASN', collated by regex parsing
 # https://metasmoke.erwaysoftware.com/data/sql/queries/164-bad-asn-in-false-positives
 COUNTRY = [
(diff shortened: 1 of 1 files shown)


bertiebaggio:
Running this on a test Smokey instance overnight - which was pleasantly simple to set up! - resulted in 7 skips. Not a huge amount but with a growing review queue reducing the numbers might help a little! I also realised I had omitted a couple of entries, which I rectified before creating this PR.

findspam.py
```diff
@@ -59,6 +59,13 @@
     "microsoft", "newegg", "cnet", "regex101", r"(?<!plus\.)google", "localhost", "ubuntu", "getbootstrap",
     r"jsfiddle\.net", r"codepen\.io", "pastebin"
 ] + [se_dom.replace(".", r"\.") for se_dom in SE_SITES_DOMAINS])))
+ASN_WHITELISTED_WEBSITES = ["docs.unity3d.com", "trac.ffmpeg.org", "bitcoincore.org", "ffmpeg.org",
+    "latex.codecogs.com", "www.advancedcustomfields.com", "name.com", "businessbloomer.com",
```
angussidney: Flake8 is suggesting that these four lines should be indented to be level with the start of the square brackets

findspam.py
```diff
@@ -1157,6 +1164,7 @@ def bad_ip_for_url_hostname(s, site):
 
 def asn_for_url_host(s, site, asn_list):
     for hostname in post_hosts(s, check_tld=True):
+        if hostname in ASN_WHITELISTED_WEBSITES: continue
```
angussidney: Flake8 is also suggesting that you separate this into two lines

findspam.py
bertiebaggio: Thanks for weighing in on that one- I wasn't sure about that as the whitelist above had a different indent level :)

bertiebaggio:
> It looks like you have a few code linting errors - they shouldn't be too hard to fix

Sorted now. Also added a low-output debug-level log statement as the actual check part has one. May not be necessary but low impact as in practice it didn't spit out much when I was running a test instance.

findspam.py
```diff
@@ -1157,6 +1165,10 @@ def bad_ip_for_url_hostname(s, site):
 
 def asn_for_url_host(s, site, asn_list):
     for hostname in post_hosts(s, check_tld=True):
+        if hostname in ASN_WHITELISTED_WEBSITES:
```
tripleee: As indicated in a separate comment, this should be something like

    if any(hostname == x or hostname.endswith("."+x) for x in ASN_WHITELISTED_WEBSITES):

tripleee:
Thanks for this submission, I have been meaning to do something like this.

tripleee:
Forked your query: https://metasmoke.erwaysoftware.com/data/sql/queries/165-bad-asn-in-false-positives

findspam.py
```diff
@@ -59,6 +59,14 @@
     "microsoft", "newegg", "cnet", "regex101", r"(?<!plus\.)google", "localhost", "ubuntu", "getbootstrap",
     r"jsfiddle\.net", r"codepen\.io", "pastebin"
 ] + [se_dom.replace(".", r"\.") for se_dom in SE_SITES_DOMAINS])))
+ASN_WHITELISTED_WEBSITES = ["docs.unity3d.com", "trac.ffmpeg.org", "bitcoincore.org", "ffmpeg.org",
```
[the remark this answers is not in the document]
bertiebaggio: That makes sense. 

As an initial pass I thought that 'if x in y' was easier to understand at a glance; it also made the ASN whitelist conservative / explicit- though the only situation where we would want to whitelist foo.example.com but still have bar.example.com go through ASN checks would be subdomain hosters like wordpress.com, Wix, etc. (Not the case proposed here)

I agree that subdomain checks are worthwhile though. That would also permit stripping those that have 'www' but no other duplicates back to plain hostname as well.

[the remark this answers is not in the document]
bertiebaggio: I have tested these against 2 new checks in `test_findspam.py`:

 - skip ASN subdomain not explicitly on whitelist (noexist.ffmpeg.org)
 - catch bad ASN (www.siliconflip.com)

should I include these in the branch for merging?

tripleee:
Sorry for taking so long to get back to you, and thanks for the contribution!

bertiebaggio:
That's great, thanks- this was my first PR, and it's nice to be part of the process ☺

PR #2879, opened 2019-04-19
Fixes a couple mistakes

LunarWatcher:
Bumps the tp rate on both, still 0 hits on SE (aside one current tp, but that one isn't worth counting)


makyen:
!!/approve

PR #2983, opened 2019-05-16
Add Stephen Kennedy to Charcoal HQ

tripleee:
Creating a PR for visibility.

SOCVR room owner, over 1000 feedbacks in Metasmoke (1.13% invalid but none in the last month).


tripleee:
Also upgraded Stephen's privilege level to Core.

PR #3076, opened 2019-06-03
Add new city-list  as per discussion on charcoal chat. https://chat.s…

sri-shree:
Add new city as per discussion on charcoal chat.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1899,7 +1899,8 @@ city_list = [
     "Neemrana", "Raipur", "Faridabad", "Kanhangad", "Kanhanjad",
     "Trichy", "Tiruchirappalli", "Tiruchi", "Trichinopoly",
     "Kerala", "Mangalore", "Mangaluru", "Thane", "Bhiwandi", "Ghatkopar",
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -1899,7 +1899,7 @@ def religion_troll(s, site):
     "Neemrana", "Raipur", "Faridabad", "Kanhangad", "Kanhanjad",
     "Trichy", "Tiruchirappalli", "Tiruchi", "Trichinopoly",
     "Kerala", "Mangalore", "Mangaluru", "Thane", "Bhiwandi", "Ghatkopar",
-    "Mulund",
+    "Mulund","neemrana","kota","alwar","chittorgarh","bhilwara","banswarabhiwadi",
```
tripleee: Please add spacing and capitalization as in the surrounding code. The lack of spaces will break the flake8 test but sticking to the conventions of the surrounding code more broadly is always a good idea. (You'll probably need to break this onto two lines in order to stick within 80 columns, though I guess we tolerate longer lines.)

PR #3332, opened 2019-09-25
Update feedback.py with comments

AzharMithani:


makyen:
1. Please fix the [CI failures](https://circleci.com/gh/Charcoal-SE/SmokeDetector/55805?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). This PR won't be merged with CI failures.
2. The spacing you're using for these comments appears inconsistent. It would be better to use consistent spacing. Part of what CI is complaining about is not using the required formatting, but there's a bit more of general formatting beyond what CI would actually complain about.
3. These comments don't really appear to provide much additional information beyond the already existing code. While adding comments to code can be very helpful, these specific comments don't really appear to provide much beyond the plain and obvious meaning of the following line of code (or a couple/few lines of code).
4. It's not clear to me what is intended to be meant by "correct feedbacks" and "Incorrect feedbacks". If those comments are going to exist, then it would be better to make it clear what is meant by "correct" and "incorrect". In other words, the comments bring to mind the question of what was correct/incorrect? Was it that the feedback was invalid? Was it that the feedback was indicating something was correct/incorrect? Basically, if there are comments there, then they could be much more clear.

makyen:
@AzharMithani Given that you've submitted several PRs to various projects over a relatively short period of time, I suspect you're motivated by [Hacktoberfest](https://hacktoberfest.digitalocean.com/), which is great. However, just so you know, a PR only counts for Hacktoberfest if the PR is made between 2019-10-1 and 2019-10-31 (inclusive), which is still several days from now. On the other hand, good contributions, are, of course, valuable anytime, and we welcome people participating in this project at any time.

makyen:
You're welcome to update this and we'll consider reopening it.

Issue #2398, opened 2018-09-07
!!/report messages should be added as a comment on MS


angussidney:
### Scroll down for up-for-grabs summary and guidance.

The `!!/report` command can take an optional 'custom reason' argument. [The current implementation](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L1217) simply adds some extra text to the 'why' info for the report.

It would be ideal if this also added a comment to the MS post on behalf of the command user, so that the message is prominently shown on the MS post itself, along with the other comments.

Note: when Smokey reports a post, it doesn't know the the MS ID and thus can't send a request to `/comments/post` to add a comment. There are two potential workarounds for this:

 - Wait for a few seconds before sending the posts-by-url followed by add comment request (could we use the [tasks](https://github.com/Charcoal-SE/SmokeDetector/blob/master/tasks.py) system for this?)
 - Add a new optional field to the MS post creation process, which will post the comment once the post is created. (note that this will require some MS dev work).

angussidney:
If we can make a decision on which approach we wish to take, and do any MS dev if necessary, this should be a good [hacktoberfest] issue.

iBug:
![contrib-guidance](https://user-images.githubusercontent.com/10406565/45100551-949aa500-b121-11e8-93e9-0bda7186d113.png)

# Summary
**Difficulty:**
![1/5 stars](https://user-images.githubusercontent.com/10406565/45103061-edb90780-b126-11e8-920a-1f25b8943efa.png)  
**Scope:**
![Narrow scope](https://user-images.githubusercontent.com/10406565/45105247-bbf66f80-b12b-11e8-9a46-79972ec1f7e3.png)  
**Issue Owner:** TBC

This issue involves sending data to metasmoke with delaying and retrying. It's relatively easy given that there are working code around that you can adopt from.

# Guidance
The function you need to modify is around [line 1180 in `chatcommands.py`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L1180). You can create a new function that tries to post a comment to metasmoke, by calling [`Metasmoke.post_auto_comment()`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/metasmoke.py#398) with proper arguments. Then you can use `Tasks.later()` to start this function. It should exit upon success.

[...]

PR #3352, opened 2019-10-04
added async comment call on report post call

dineshkrishnareddy:
fix https://github.com/Charcoal-SE/SmokeDetector/issues/2398


ArtOfCode-:
Thanks @dineshkrishnareddy!

Issue #3321, opened 2019-09-13
Don't post the title in chat when it's been identified as offensive


ArtOfCode-:
Related: #811

When Smokey has detected a post using the reason "potentially offensive keyword in title", we should avoid posting that title in chat.

This should only apply to non-CHQ rooms - in CHQ it's an occupational hazard, but in other rooms it could be avoided. We can always substitute it for something like "(potentially offensive - see MS for details)".

ArtOfCode-:
![Contributor Guidance](https://charcoal-se.org/files/hacktoberfest-header.png)

# Summary
**Difficulty:**  
![2/5 stars](https://charcoal-se.org/files/issue-diff-2.png)

SmokeDetector sometimes picks up potentially offensive posts. When this happens, we'd like to avoid posting the titles of these posts in chat, as we would usually do - except for our main room, Charcoal HQ, where it's acceptable.

# Guidance
You'll need to modify the `handle_spam` function in `spamhandling.py`. Just before the end of the function where the messages get sent to chat (that's the `chatcommunicate.tell_rooms` calls), you'll need to add a new check to make sure that the `reasons` list doesn't contain any items that mention "offensive". If it does, you'll need to alter the post's title to a summary - something along the lines of "(potentially offensive - see MS for details)".

# How to get help

For additional support in completing this issue, either:

 - comment on this thread
 - if you have a Stack Overflow account, drop into [our chatroom](https://chat.stackexchange.com/rooms/11540) (you'll need 20 reputation - if you don't have that, let us know your username or ID here and someone will grant access manually)

PR #3353, opened 2019-10-04
handle spam title

dineshkrishnareddy:
fix https://github.com/Charcoal-SE/SmokeDetector/issues/3321


ArtOfCode-:
Right idea, but not quite the right place to fix it - it's in `message` that the title should be redacted, not in the call to `Metasmoke.send_stats_on_post`. The latter can have the full unredacted title passed to it; `message` should be changed to remove the title.

dineshkrishnareddy:
@ArtOfCode- thanks for helping, fixed it. I thought if we send `stats on post` it will eventually send message on chat

makyen:
Are we wanting to replace the title anytime a reason contains "offensive"?

I was under the impression we wanted to replace the title only if the offensive material was detected in the title, not anywhere in the post. The [list of reasons can be found on metasmoke](https://metasmoke.erwaysoftware.com/reasons). The specific reason for the title is "Offensive title detected". [Note: this reason text is different than is stated in the issue for this PR. The reason text mentioned in the issue doesn't appear to exist.]

ArtOfCode-:
Apologies for not being clearer here - what we're looking to do is to change _just_ the title portion of `message` when there's an offensive title. `message`, at the point you're modifying it, is something like `[ SmokeDetector | MS ] Offensive keyword detected in title: "YOU'RE ALL TERRIBLE I HATE YOU" on somesite.se`. The title portion of that is just the "YOU'RE ALL TERRIBLE I HATE YOU" - the rest of the message can stay intact. Take a look at how the `message` is constructed from various other variables, and you'll want to replicate that just with a modified title to remove the potentially offensive content.

dineshkrishnareddy:
@ArtOfCode- @makyen Hey from what I understood, if there is any reason with `Offensive title detected` we should change the title part and keep the remaining part of the message same, So I made a fix accordingly. 
let me know if this fix the issue

makyen:
@dineshkrishnareddy As you've probably noticed, another person also worked on this Issue. Their code ended up being merged. I'm sorry for the lack of clear communication that another person also submitted a PR (they did so without any prior communication or indication they were working on this). I understand it's disappointing not to have code merged. I'm sorry that this code won't end up being used.

As it turned out, the code which was merged didn't fully resolve Issue #3321, or at least not as originally desired. The portion which was not implemented was having two separate messages sent to the chat rooms. One message with the original post title sent to Charcoal HQ and another message with the masked title sent to all other rooms. Currently, there's one message, with the masked title, which is sent to all appropriate chat rooms. If you would like to work on changes such that there are those two messages, that would be appreciated. If you do, it would probably be best to close this PR and branch from the current master.

Thank you again for the effort you put in here.

ArtOfCode-:
On the bright side, as far as I know, not being merged doesn't affect whether a PR qualifies for Hacktoberfest this year, so this should still count towards your progress if you're participating.

dineshkrishnareddy:
@makyen @ArtOfCode- thanks for your concern but can you please let me know how to send message to Charcoal HQ, I checked the code base but couldn't find any

makyen:
Well, given that room configurations are controlled by rooms.yml, what's probably needed is an additional configuration option in rooms.yml under `msg_types:` which is something like:

    - dont-mask-offensive-titles

That config would then be placed under all rooms in which offensive titles shouldn't be masked (i.e. Charcoal HQ).

Then `chatcommunicate.tell_rooms` would need to be able to be called with appropriate arguments indicating 3 different cases: 

1. The message should be sent ignoring that config option (i.e. just obeying the criteria as they currently exist). In other words, basically what happens now (i.e. all current calls to `chatcommunicate.tell_rooms` should continue to work).
2. The message should be sent to rooms qualifying by the other indicated criteria, but also requiring having the `dont-mask-offensive-titles` config option.
3. The message should be sent to rooms qualifying by the other indicated criteria, but also requiring *not* having the `dont-mask-offensive-titles` config option.

This could probably be handled by adding that config option to Charcoal HQ in rooms.yml and using it as a prop in the `has` and `hasnt` arguments.

There might be a better way to do it, but that's what I see off the top of my head. I've also only taken a relatively cursory look at the code in this area, so might be missing something.

makyen:
@dineshkrishnareddy So that you know, ArtOfCode has added code which implement the change I detailed above. It ended up being more involved than we would normally expect for a Hacktoberfest project. As a result of the problem being resolved, work on this issue is no longer needed. Thank you for the effort which you put into this. I'm sorry that it didn't result in a merged PR for you.

PR #3366, opened 2019-10-06
Don't post the title in chat when it's been identified as offensive

sgh304:
This adds handling for #3321.

If the 'offensive title detected' reason is detected when handling a spam post, the built message will not feature the post's title, but instead `(Potentially offensive title -- see MS for details)`. A test has also been added to verify this behavior.

One question: I took a quick look, but are there any other reasons that should cause the post title to be left out of the message?

### DIFF
--- a/spamhandling.py
+++ b/spamhandling.py
@@ -136,8 +136,12 @@ def handle_spam(post, reasons, why):
         # in a proper way in chat messages.
         sanitized_title = parsing.sanitize_title(post.title if not post.is_answer else post.parent.title)
         sanitized_title = escape_format(sanitized_title).strip()
+
         # Remove title if it is potentially offensive
-        message_title = sanitized_title if 'offensive title detected' not in reasons else '(Potentially offensive title -- see MS for details)'
+        if 'offensive title detected' not in reasons:
+            message_title = sanitized_title
+        else:
+            message_title = '(Potentially offensive title -- see MS for details)'
 
         prefix = u"[ [SmokeDetector](//git.io/vyDZv) ]"
         if GlobalVars.metasmoke_key:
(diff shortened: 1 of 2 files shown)


makyen:
As I understood it, the desire was to have the message which is send to Charcoal HQ still contain the original title, while messages sent to any other rooms contain the substituted title. That means that two different chat messages need to be constructed and sent, with separately limited distribution on each message. I believe that what this currently does is change the title for all rooms.

In addition to having a test which checks for the title being substituted, it would be good to have a test that checks that the title isn't substituted when the detection reason isn't matched.

sgh304:
@makyen good idea on the test, thank you.

as for the first part, my assumption from the desired error message (`(Potentially offensive title -- see MS for details)`) was that we didn't want to include the title in any channel messages, however we still wanted to log the title in MS itself (which I believe the `send_stats_on_post` call here https://github.com/Charcoal-SE/SmokeDetector/blob/6eec69001e1d6c5d5da939f8ec6aa155010fddb0/spamhandling.py#L165 does). am i making sense?

makyen:
@sgh304 The [issue](https://github.com/Charcoal-SE/SmokeDetector/issues/3321) says (emphasis mine):

> This should **only apply to non-CHQ rooms** - in CHQ it's an occupational hazard, but in other rooms it could be avoided.

However, the initial comment in the issue is missing an explanation as to what CHQ means. CHQ is an abbreviation for the [Charcoal HQ chat room](https://chat.stackexchange.com/rooms/11540/charcoal-hq).

I'm not saying that the change will be rejected without implementing different messages in CHQ and other rooms, but it's preferred, at least as far as I'm aware.

ArtOfCode-:
Let's see how this goes for now. If we decide we do want the titles back in CHQ, we can always add that in later on.

Issue #3467, opened 2019-11-10
CI tests have been silently failing for the last month


makyen:
Our CI tests have been silently failing for a month.

The last build on the master branch that *actually* passed CI testing was build number 56777 [on 2019-10-08](https://circleci.com/gh/Charcoal-SE/SmokeDetector/56777) (commit [297e0f9c](https://github.com/Charcoal-SE/SmokeDetector/commit/297e0f9cc45d97f92f99e25e642dd6acda2a4b74)). The first build that failed was [just after that](https://circleci.com/gh/Charcoal-SE/SmokeDetector/56779) ([commit 61cf8381](https://github.com/Charcoal-SE/SmokeDetector/commit/61cf8381d3d8b9dd7ad7f59d0b1d5e28fd18e5e4)).

#### A good run:
The output for a good `pytest` looks like (note the existence of a summary of the test results) (from [this CircleCI run](https://circleci.com/gh/Charcoal-SE/SmokeDetector/56777)):
[![passing pytest output][1]][1]

#### `pytest` exiting early without completing the tests, but not indicating failure:
The output of the crashing/failing `pytest` runs looks like (note lack of summary) (from [this CircleCI run](https://circleci.com/gh/Charcoal-SE/SmokeDetector/56779)):
[![crashing pytest output][2]][2]
The testing output looks so close to the correctly functioning one because the tests which were causing the issue were in the last few of the entire sequence. That's just where they happened to be located. Changing the order of the tests just results in the tests following the ones with the issue not being run.

I've submitted PR #3468, which resolves the immediate issue of the tests crashing. However, that PR won't resolve the issue that `pytest` *can* crash and silently fail. Until that's resolved, we need to be periodically manually checking that `pytest` actually completes.

<sub>Note: the title of this is tending towards click-bait. That's intended. The fact that `pytest` can silently fail is something which our contributors should be aware of, so they check for it from time-to-time. Once a bit of time has passed, I expect to change the title to something less like click-bait.</sub>

  [1]: https://i.stack.imgur.com/zpXQr.png
  [2]: https://i.stack.imgur.com/k6vcO.png

ArtOfCode-:
That's a warning, not a test failure - currently status-bydesign. If we want to fail on warnings too, I imagine we can do that somehow.

makyen:
[...]

PR #3468, opened 2019-11-10
Fix issues with offensive title tests

makyen:
Unfortunately, there's been an issue with the test cases for the offensive title tests which resulted in the `pytest` test silently failing on every CI build over the last month.

This resolves the issues which were causing the failure, but doesn't resolve the problem that `pytest` can silently fail.

In addition, this adapts the offensive title tests to the updated code which implements sending masked and unmasked messages to appropriate rooms. It just checks that `chatcommunicate.tell_rooms()` is called appropriately. It does not check for the messages actually being sent to appropriate rooms.

This is related to Issue #3467, but doesn't solve the problem that `pytest` *can* fail silently.

### DIFF
--- a/test/test_spamhandling.py
+++ b/test/test_spamhandling.py
@@ -7,7 +7,8 @@ import pytest
 import os
 import json
 from classes import Post
-from unittest.mock import MagicMock, ANY
+from unittest import TestCase
+from unittest.mock import MagicMock, call, ANY, DEFAULT
 import chatcommunicate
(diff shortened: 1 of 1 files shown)


makyen:
I was planning for this to be reviewed and/or to push another commit with a general solution. However, I need to do a bit more testing on the general solution and it looks like it's more important to just get the CI testing actually working. Thus, I'm going to merge this.

PR #3545, opened 2019-12-03
Add .ae to pattern-matching TLD rule

tripleee:
I propose adding this TLD to the list of domains we always flag.

https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=%5C.ae%5Cb&post_type=answer indicates 87/95 where 3 of the 4 TPs are more than a year old.


ArtOfCode-:
Works for me.

PR #3568, opened 2019-12-06
IP check examines nameserver IP addresses, too

### LINKED ISSUE #3437: IP watch / blacklist should examine NS address, too

132.148.29.42 is both the web site address and the name server address for techindiainfotech.com. In this particular case, we probably don't want two rules to trigger (if we can avoid it) but more generally, when an IP address is watched or blacklisted, we also want it to trigger if it matches on the address of a name server for a domain.

tripleee:
Closes #3437


tripleee:
Sorry for the noise, I seem to have gotten a bit rusty /-:

PR #3580, opened 2019-12-11
findspam.py: Move IP, NS, and ASN lists to external YAML files

tripleee:
Rather than keep on  committing brittle changes to `findspam.py` for routine updates, move these resources to external files, just like we already have for bad keywords, domain names, and phone numbers.

I grappled with the design but eventually landed on <del>JSON</del><ins>YAML</ins> for this. I don't think adding more ad-hoc TSV variants is a good way forward, and envision moving all of our external lists to a unified format eventually (YAML or something similarly flexible and structured).

We can't make assumptions about the order of the YAML entries in each file, but PyYAML always seems to sort the keys (which is why the `Schema` keys are proper case). The results look good so far, but it probably requires additional testing. I have only done brief sanity checks after converting from JSON (which is what the original version of this pull request had) to YAML. Unlike JSON, YAML should be reasonably diffable and human-editable.

Down the line, I also hope to implement some sort of schema validation. I started on my own ad-hoc validation rules but decided to keep them really simple and maybe look at some proper YAML schema validation module going forward.

I'm hoping to migrate the IP blacklist to proper CIDR netblocks in a future update; this already contains some movement in that direction (though for the time being mainly in the naming).


iBug:
What about using YAML like we're already doing for user privileges and chat rooms?

tripleee:
We are using JSON in a lot of places too; but I guess I should look at YAML as well. It could in fact be somewhat easier to coerce into the line per entry organization of the old CSV files (at least approximately).

tripleee:
@iBug Thanks for the suggestion; please don't merge this yet -- I'll refactor it to use YAML instead.

tripleee:
Thanks for the suggestion. This is now refactored to use YAML instead of JSON and rebased + force pushed.

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+    SCHEMA_VARIANT = 'yaml_cidr'
+    SCHEMA_PRIKEY = 'ip'
+
+    def __init__(self, filename):
+        super().__init__(filename)
+
+    def _parse(self, keep_disabled=False):
+        with open(self._filename, 'r', encoding='utf-8') as f:
+            y = yaml.safe_load(f)
+        assert y['Schema'] == self.__class__.SCHEMA_VARIANT
```
iBug: Repeated appearances of `__class__` seems redundant to me. Just use `self.AAA` instead of `self.__class__.AAA` and it'll work correctly (`self.attr` falls back to `self.__class__.attr` if `attr` not found for object).

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+        assert y['Schema_version'] <= self.__class__.SCHEMA_VERSION
+        for item in y['items']:
+            if not keep_disabled and 'disable' in item and item['disable']:
```
iBug: Maybe `item.get('disable')` which defaults to `None` if key `disable` isn't present?

blacklisted_cidrs.yaml
```diff
@@ -0,0 +1,24 @@
+Schema: yaml_cidr
```
iBug: Unsure if this matters much, but I've rarely seen uppercase in YAML keys (unless they're treated as strings instead of identifiers). Also, I'd suggest a file extension of `yml` rather than `yaml`, though this is purely preferential. Think about it like `jpg` vs `jpeg`.

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+            r'(?(DEFINE)(?P<octet>'
+            '' r'1[0-9]{0,2}|2(?:[0-4][0-9])??|25[0-5]?|2[6-9]|[3-9][0-9]?))'
+            r'^(?&octet)(?:\.(?&octet)){3}$'
+        )
+        if 'ip' in item:
+            if not ip_regex.match(item['ip']):
+                raise ValueError('Field "ip" is not a valid IP address: {0}'.format(
+                    item['ip']))
+            '''
+            if 'cidr' in item:
```
iBug: What about unioning `ip` and `cidr` so that a CIDR range can appear as the value to an `ip` key, while treating pure IPs as if they had a `/32` suffix. Many places I see supports mixing them.

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+                key=lambda x: x[self.__class__.SCHEMA_PRIKEY])
+        }
+        callback(d)
+        with open(self._filename, 'w', encoding='utf-8') as f:
+            yaml.dump(d, f)
+
+    def _validate(self, item):
+        ip_regex = regex.compile(
+            r'(?(DEFINE)(?P<octet>'
+            '' r'1[0-9]{0,2}|2(?:[0-4][0-9])??|25[0-5]?|2[6-9]|[3-9][0-9]?))'
```
iBug: I'm unsure what the purpose of the leading pair of single quotes here is...

blacklisted_cidrs.yaml
tripleee: Yeah, I prefer `jpeg` (-:

iBug: But we already have `rools.yml` and `users.yml`, so consistency becomes another point for me :-)

tripleee: Like I allude to in the updated PR description, I canged the Schema keys to capitalized to keep them at the top of the file. It seems that PyYAML forces alphabetical order and this is the simplest workaround, though I begrudge it.

blacklists.py
tripleee: I plan on something like that going forward, but for the time being this simply moves the data we have now. My proposal would be to use `cidr: '12.34.56.0/24'` but allow 'ip: '12.34.56.78'` as a convenience shorthand for `cidr: '12.34.56.78/32'`

blacklists.py
tripleee: It's a hack to indent the line visually even though `flake8` boneheadedly wants to enforce its hamfisted convention. Maybe I'll have to add a comment.

iBug: Perhaps [`(?x)`](https://docs.python.org/3/library/re.html#re.X) is a better solution than this hack.

blacklists.py
iBug: Sorry if I didn't made myself clear. I mean we accept both `ip: 1.1.1.1` and `ip: 1.1.1.1/24` forms and discard the `cidr` key, just as you'd write in a firewall rule (you don't put raw IPs and CIDR blocks into two inputs, right?).

blacklists.py
iBug:
```suggestion
            if not keep_disabled and item.get('disable'):
```

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+                self._parse(keep_disabled=True),
+                key=lambda x: x[self.__class__.SCHEMA_PRIKEY])
+        }
+        callback(d)
+        with open(self._filename, 'w', encoding='utf-8') as f:
+            yaml.dump(d, f)
+
+    def _validate(self, item):
+        ip_regex = regex.compile(
+            r'(?(DEFINE)(?P<octet>'
```
iBug:
```suggestion:-1+3
        ip_regex = regex.compile(r'''
            (?(DEFINE)(?P<octet>
              1[0-9]{0,2}|2(?:[0-4][0-9])??|25[0-5]?|2[6-9]|[3-9][0-9]?))
            ^(?&octet)(?:\.(?&octet)){3}$''', regex.X
```

blacklists.py
iBug:
```python
        ip_regex = regex.compile(r'''
            (?(DEFINE)(?P<octet>
              1[0-9]{0,2}|2(?:[0-4][0-9])??|25[0-5]?|2[6-9]|[3-9][0-9]?))
            ^(?&octet)(?:\.(?&octet)){3}$''', regex.X
        )
```

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+                if compare[prikey] == item[prikey]:
+                    break
+            else:
+                raise ValueError('No {0} found in list {1}'.format(
+                    item[prikey], d['items']))
+            del d['items'][i]
+
+        self._write(remove_callback)
+
+    # FIXME: enumerate gets YAML item array index, not line number
```
iBug: I'm afraid this isn't easily solvable. We could just stick with YAML array index and rely on Ctrl-F (or Vim `/`) for quick lookup.

blacklisted_cidrs.yaml
tripleee: The consistency argument convinced me, though I note that this flies in the face of some official YAML recommendation.

blacklists.py
tripleee: Hmm, but a CIDR isn't an IP. But that's for the future anyway.

blacklists.py
tripleee: I found a Stack Overflow recipe for this but let's tackle that separately if that seems important. I guess the `!!/bisect` functionality is where we need to display the line number; perhaps that could be refactored to not always assume the result is a line number.

blacklists.py
tripleee: Updated and rebased with this change; thanks for the suggestion!

blacklists.py
tripleee: Again, thanks for this tip.

blacklisted_cidrs.yml
```diff
@@ -0,0 +1,24 @@
+Schema: yaml_cidr
```
ArtOfCode-: Do these need to be capitalized? I'd prefer to standardize casing if not, i.e. `schema:` and `schema_version:`

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+    SCHEMA_VARIANT = 'yaml_cidr'
+    SCHEMA_PRIKEY = 'ip'
+
+    def __init__(self, filename):
+        super().__init__(filename)
+
+    def _parse(self, keep_disabled=False):
+        with open(self._filename, 'r', encoding='utf-8') as f:
+            y = yaml.safe_load(f)
+        assert y['Schema'] == self.SCHEMA_VARIANT
```
ArtOfCode-: Rather than asserts, can these be `if` checks with a `log('error', '...')` call if they fail? Logging then exiting is preferable to just crashing.

blacklists.py
```diff
@@ -140,17 +147,174 @@ def exists(self, item: Union[str, dict]):
...
+    """
+    YAML parser for name server blacklists.
+    """
+    SCHEMA_VARIANT = 'yaml_ns'
+    SCHEMA_PRIKEY = 'ns'
+
+    def _validate(self, item):
+        host_regex = regex.compile(r'^([a-z0-9][-a-z0-9]*\.){2,}$')
+        if 'ns' not in item:
+            raise ValueError('Item must have member field "ns": {0!r}'.format(item))
```
ArtOfCode-: Are these ValueErrors caught and logged, or do they crash the instance? If it's the latter, it'd be preferable to change it to the former, even if we have to exit after logging it.

blacklisted_cidrs.yml
tripleee: This was brought up by iBug as well. I wanted them to be lower case but that caused PyYAML to put them at the end of the file, which wasn't great either. Apparently it is hardcoded to alphabetize the keys on the main level, with no easy workaround. So I grudgingly capitalized them.

blacklists.py
tripleee: I guess that should be changed; I see the TSV parser has a `try`/`except` around similar constructs. I suppose I could use that here too. I was thinking perhaps there should be a stricter validation in the unit tests but I haven't worked on that yet.

blacklists.py
tripleee: Yeah, same as above.

blacklists.py
tripleee: I updated to replace the asserts, too, with raise ValueError. I'm still thinking about how to robustly trap the errors, and whether in fact it makes more sense to trap and continue. These are integrity errors which in some sense _should_ fail, though of course it's nice if it doesn't ruin the other parts of Smokey even when there is a problem.

tripleee: I added a commit which allows Smokey to start up even if the blacklists are hosed. It will still fail in interesting ways if you attempt to add or remove something invalid; but doing that programmatically is going to be in the scope of a separate PR, so I'm not touching the `gitmanager.py` code yet.

blacklists.py
tripleee: Good idea, thanks!

blacklisted_cidrs.yml
ArtOfCode-: In that case, can we standardize `Schema_version` to `SchemaVersion`?

blacklists.py
makyen: I need to look at the new commits to see what the update was, but just based on the text description:

I would expect SD failing to start up when the blacklists are hosed would be a feature. I'm not sure we want to enable it to properly start when the blacklists are messed up, unless that's behind an option of some sort. Blacklist integrity is something we want to maintain and we want SD to be able to use the data in the external lists.

As I understand it, currently, if any of the blacklists fail integrity checks, then SD will fail to start. If SD fails to start three times in a row, then nocrash.py will move the instance to the next oldest commit. This process repeats until it finds a commit which runs (e.g. where the blacklists are not messed up). Are we sure we want to do something other than that? Why would we want to run, in production, on a commit where the blacklists are messed up?

blacklisted_cidrs.yml
tripleee: Is that really an improvement though? Ugh.

blacklists.py
tripleee: So I will just drop that commit for now then.

blacklists.py
tripleee: Here's a Stack Overflow question with a couple of suggested answers which look at least vaguely promising: https://stackoverflow.com/questions/13319067/parsing-yaml-return-with-line-number

PR #3616, opened 2019-12-19
I translated your readme

kwarkmc:
I translated README.md in Korean! It will be helpful for many Korean users or contributors!

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -1,0 +1,166 @@
+SmokeDetector
+=============
+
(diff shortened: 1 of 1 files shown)


tripleee:
Thank you, it looks like you went to a lot of effort here. Unfortunately, we're not currently looking to put out information in languages other than English; we will not be able to maintain or develop non-English resources or communicate to people in their own language anyway at the moment. It's apprecated that you took the time to do this, however. But at this point, I don't think we can accept this contribution.

PR #3625, opened 2019-12-21
Improve link unshortener function

iBug:
This patch includes the following fixes / improvements for the link unshortener helper

- Use GET requests

  Some link shortener servers reject HEAD requests or otherwise disregards the W3 specs for HEAD requests by returning something different (e.g. 200 OK, try `curl -I https://t.cn/blahblah`)

- Limit maximum redirect chain length

  Previous code would block if any url(s) in the chain loops and this function will unfortunately run forever. Now raises an exception instead. Default depth is 10 which should cover 99.9% of them. For reference, most browsers stop after 20 or so (see [this Stack Overflow answer](https://stackoverflow.com/a/36041063/5958455)).

- Allow arbitrary HTTP method (body is always discarded)

  Now that the response body is discarded (credits to [this answer](https://stackoverflow.com/a/23718527/5958455)), it's safe to use an arbitrary HTTP request method, though it's still potentially vulnerable if a server decides to send a lot of headers. ~~This remains unfixed however.~~ `requests` seems to limit this to 100 header lines and a total of 64 KiB for all headers by default, so we're not vulnerable.

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -209,7 +209,7 @@ def unshorten_link(url, request_type='GET', depth=10, explicitly_ignore_security
     for tries in range(depth):
         if response_code not in [301, 302, 303, 307, 308]:
             break
-        res = requests.request(request_type, url, headers=headers, stream=True)
+        res = requests.request(request_type, url, headers=headers, stream=True, allow_redirects=False)
         res.connection.close()  # Discard response body for GET requests
         response_code = res.status_code
         if 'Location' not in res.headers:
--- a/helpers.py
+++ b/helpers.py
@@ -203,6 +203,7 @@ def reload_modules():
 
 
 def unshorten_link(url, request_type='GET', depth=10, explicitly_ignore_security_warning=False):
+    orig_url = url
     response_code = 301
     headers = {'User-Agent': 'SmokeDetector/git (+https://github.com/Charcoal-SE/SmokeDetector)'}
     for tries in range(depth):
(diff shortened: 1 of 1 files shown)


iBug:
@ArtOfCode- ?

ArtOfCode-:
_le sigh_... and I was _just_ saying how good I am at getting sidetracked. Sorry. Incoming shortly...

helpers.py
```diff
@@ -202,28 +202,22 @@ def reload_modules():
...
-        'GET': requests.get,
-        'HEAD': requests.head
-    }
-    if request_type not in requesters:
-        raise KeyError('Unavailable request_type {}'.format(request_type))
-    if request_type == 'GET' and not explicitly_ignore_security_warning:
-        raise SecurityError('Potentially unsafe request type GET not acknowledged')
-
-    requester = requesters[request_type]
+def unshorten_link(url, request_type='GET', depth=10, explicitly_ignore_security_warning=False):
```
ArtOfCode-: You can drop the `explicitly_ignore_security_warning` parameter here now.

helpers.py
```diff
@@ -202,28 +202,22 @@ def reload_modules():
...
+    orig_url = url
     response_code = 301
     headers = {'User-Agent': 'SmokeDetector/git (+https://github.com/Charcoal-SE/SmokeDetector)'}
-    while response_code in {301, 302, 303, 307, 308}:
-        res = requester(url, headers=headers)
+    for tries in range(depth):
+        if response_code not in [301, 302, 303, 307, 308]:
```
ArtOfCode-: This can (and technically probably should) still be a set - `{301, 302, 303, 307, 308}`.

iBug:
@ArtOfCode- I'm going to fix the ones you've reviewed, and would you handle this spammer? Thanks.

ArtOfCode-:
Spammer handled and reported to GH support. I can't delete the reviews, but I guess support might be able to.

Issue #3768, opened 2020-03-21
Permit rooms.yml to specify receiving experimental reports on a per-site basis


makyen:
On numerous occasions, there has been either confusion over why some reports for a site have not appeared in a room focused on the site (e.g. reports for stackoverflow.com in SOCVR), or a desire for one room or another to receive "experimental" reports only for specific sites.

Currently, if a post is caught only by [reasons considered "experimental"](https://github.com/Charcoal-SE/SmokeDetector/blob/master/globalvars.py#L79), then it is sent only to rooms which have `experimental` in their `msg_types` in rooms.yml. This is currently an all-or-nothing setting. The room either gets experimental reports for all sites, or it gets them for no sites.

### Proposal
In rooms.yml under `msg_types`:
* Change `experimental` to `experimental-all-sites`, which indicates that reports with only experimental reasons will be sent to the room for all sites.
* Add `experimental-site-<site domain>` (e.g. `experimental-site-stackoverflow.com`). This is similar to the format of the `site-<site domain>` entries which already exist.
* Change `all` to `all-sites`, so it's clearer the setting indicates the room accepts reports for all sites. "All sites" is the current actual meaning of `all`. Unfortunately, `all` within `msg_types` could be interpreted in various ways. Personally, when I see just `all`, I assume it means that the room should receive every single chat message type, not that it means just "all sites".

PR #3769, opened 2020-03-21
Per-site "experimental-site-<site domain>" setting for rooms.yml

makyen:
In rooms.yml under `msg_types`:
* Change `experimental` to `experimental-all-sites`, which indicates that reports with only experimental reasons will be sent to the room for all sites.
* Add `experimental-site-<site domain>` (e.g. `experimental-site-stackoverflow.com`). This is similar to the format of the `site-<site domain>` entries which already exist.
* Change `all` to `all-sites`, so it's clearer the setting indicates the room accepts reports for all sites. "All sites" is the current actual meaning of `all`. Unfortunately, `all` within `msg_types` could be interpreted in various ways. Personally, when I see just `all`, I assume it means that the room should receive every single chat message type, not that it means just "all sites".

Implements #3768

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -326,16 +326,22 @@ def tell_rooms(msg, has, hasnt, notify_site="", report_data=None):
     msg = msg.rstrip()
     target_rooms = set()
 
+    # Go through the list of properties in "has" and add all rooms which have any of those properties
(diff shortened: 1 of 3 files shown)


tripleee:
Looks very straightforward to me. Thanks in particular for the added comments.

PR #3790, opened 2020-04-02
YAML validation

tripleee:
This adds a test case to check the YAML lists for some simple problems.

Testing revealed a number of issues which are also solved with this PR.

- Embarrassing typos in a regex
- Embarrassing typos in a couple of watched NS entries
- Comment out a number of stale NSes which no longer resolve

The commented-out entries should probably be removed eventually, but some of the issues may be transient, so I want to keep them around and revisit them in a few weeks.

### DIFF
--- a/watched_nses.yml
+++ b/watched_nses.yml
@@ -101,7 +101,7 @@ items:
   - lee.ns.cloudflare.com.
 - ns:
   - barbara.ns.cloudflare.com.
-  - logan.ns.cloudflarem.com.
+  - logan.ns.cloudflare.com.
 - ns:
   - bart.ns.cloudflare.com.
   - lady.ns.cloudflare.com.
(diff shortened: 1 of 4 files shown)


tripleee:
The test currently adds about 45 seconds to the duration of the test suite. Maybe it should be hacked to run only at the end.

tripleee:
The CI test failure looks spurious; I sometimes get this on my laptop, too.

Will rebase and re-push and hope it works this time.

watched_nses.yml
```diff
@@ -516,7 +513,9 @@ items:
   ns: mihanwebhost.com.
 - ns: offshoreracks.com.
 - ns: sathyats.net.
-- ns: shared-host.org.
+- comment: no longert responding 2020-04-02
```
ArtOfCode-, line 516: You've substituted typos in NSes to typos in comments :)

tripleee: Thanks for noticing, I'll fix and merge on Monday.

tripleee:
See also now #3793.

PR #3806, opened 2020-04-09
Reenable DNS tests, catch on dns.resolver.NoNameservers (fix #3793)

### LINKED ISSUE #3793: Spurious failures in CI

Two of my recent PRs had Circle CI failures in code which I had not touched. One of them failed, then succeeded after I rebased my commits -- the other exhibited the opposite behavior: test passed, then I rebased the commits and force pushed, and the same code failed.

For the record; #3789 #3790

The test should not fail spuriously.

I think I had this on my laptop occasionally too, so I don't think it's specific to Circle CI.

The failing code has some comments which vaguely hint at what might be wrong, but then why would they fail only some of the time?

teward:
This should fix #3793 as it now handles and catches upon the unhandled `dns.resolver.NoNameservers` error that was breaking CI when DNS SERVFAILs are encountered.

### DIFF
--- a/blacklists.py
+++ b/blacklists.py
@@ -293,6 +293,10 @@ class YAMLParserNS(YAMLParserCIDR):
                     soa = dns.resolver.query(ns, 'soa')
                     log('debug', '{0} has no A record; SOA is {1}'.format(
                         ns, ';'.join(s.to_text() for s in soa)))
+            except dns.resolver.NoNameservers:
+                if not item.get('pass', None):
+                    log('debug', '{0} has no available servers to service DNS '
+                                 'request.'.format(ns))
             return True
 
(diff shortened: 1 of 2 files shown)


blacklists.py
```diff
@@ -293,6 +293,10 @@ def item_check(ns):
                     soa = dns.resolver.query(ns, 'soa')
                     log('debug', '{0} has no A record; SOA is {1}'.format(
                         ns, ';'.join(s.to_text() for s in soa)))
+            except dns.resolver.NoNameservers:
+                if not item.get('pass', None):
+                    log('debug', '{0} has no available servers to service DNS '
```
ArtOfCode-, line 298: Should be WARN loglevel

teward: Pushed an update.

Issue #3720, opened 2020-01-14
Exclude goo.gl/forms and goo.gl/maps from the short URL check


gparyani:
Yes, `goo.gl` is a URL shortener, and yes, I'm aware that it's been discontinued. I don't disagree that it should remain on the watchlist since old links still continue to work.

However, since (and before) its discontinuation, it has been repurposed to provide internal Google link shortcuts, e.g. `goo.gl/maps/...` and `goo.gl/forms/...`.

Can these, as well as any other legitimate Google-internal short URL schemes, please be removed from the short URL check?

tripleee:
[goo.gl/forms](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body=goo.gl%2Fforms) is 34/37 and many of the recent TPs trigger no other rule. (Well, none of them are strictly speaking recent, but anyway.) Do we really want to disable this just to add a separate watch for forms?

[goo.gl/maps](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body=goo.gl%2Fmaps) is a more reasonable 7/24 where only one of those 7 TPs was caught solely because of this rule.

gparyani:
Fixed in #3813

PR #3813, opened 2020-04-14
Exclude goo.gl/maps/ from URL shortener (#3720)

tripleee:
Also refactors URL shorteners so they are only defined in one place.

#3720 asks for /forms/ but I am not including that for now; see comment in the ticket.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -63,11 +63,12 @@ WHITELISTED_WEBSITES_REGEX = regex.compile(r"(?i)upload|\b(?:{})\b".format("|".j
 URL_SHORTENER = r"(?:{})".format('|'.join(regex.escape(site) for site in (
     '9nl.me', 'adf.ly', 'adfoc.us', 'adyou.co', 'alturl.com', 'amzn.to',
     'bfy.tw', 'bit.do', 'bit.ly', 'bluenik.com', 'buff.ly',
-    'cl.ly', 'clkmein.com', 'dyo.gs', 'fb.me', 'goo.gl',
+    'cl.ly', 'clkmein.com', 'dyo.gs', 'fb.me', 'goo.gl',  # doctored; see below
     'is.gd', 'j.mp', 'ow.ly', 'post.ly', 'rurl.us', 'surl.cn.com'
     't.co', 'tiny.cc', 'tinyurl.com', 'tr.im', 'tgig.ir',
     'wp.me',
-    )))
+)))
+URL_SHORTENER = URL_SHORTENER.replace(r'goo\.gl', r'goo\.gl(?![?&/]maps/)')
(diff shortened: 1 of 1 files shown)


findspam.py
```diff
@@ -60,6 +60,14 @@
...
     r"jsfiddle\.net", r"codepen\.io", "pastebin", r"nltk\.org", r"xahlee\.info", r"ergoemacs\.org"
 ] + [se_dom.replace(".", r"\.") for se_dom in SE_SITES_DOMAINS])))
+URL_SHORTENER = r"(?:{})".format('|'.join(regex.escape(site) for site in (
+    '9nl.me', 'adf.ly', 'adfoc.us', 'adyou.co', 'alturl.com', 'amzn.to',
+    'bfy.tw', 'bit.do', 'bit.ly', 'bluenik.com', 'buff.ly',
+    'cl.ly', 'clkmein.com', 'dyo.gs', 'fb.me', 'goo.gl',
+    'is.gd', 'j.mp', 'ow.ly', 'post.ly', 'rurl.us', 'surl.cn.com'
+    't.co', 'tiny.cc', 'tinyurl.com', 'tr.im', 'tgig.ir',
+    'wp.me',
+    )))
```
makyen: All of these strings need the `.` changed to `\\.`, or to be raw strings an be `\.`.

I'd suggest a variable name that explicitly states that it's a regex. So, something like `URL_SHORTENER_DOMAINS_REGEX`


While it didn't previously exist, if we're going to be detecting these domains generically, they need some type of bookending to prevent false positives. What I've been using in the blacklisted websites list is:

    \b(?<![^\W_]-)gfix\.in(?![.-][^\W_])\b

So, line 63 above could be changed to:

    URL_SHORTENER_DOMAINS_REGEX = r"(?:\b(?<![^\W_]-)(?:{})(?![.-][^\W_])\b)".format('|'.join(regex.escape(site) for site in (

If this is only going to be used in the two locations indicated in this commit, then bookending here isn't needed, because those two uses have their own, more restrictive, bookending. My concern is that if we have a variable with a general list, then it's likely to be used somewhere else sometime in the future and we'd need to remember that it needs to be bookended.

tripleee: The regex already contains the backslash; we run the whole thing through `regex.escape` (which is why the goo.gl entry has to be doctored separately afterwards).

I'll rename the variable and think about the other comments. Thanks for the feedback.

makyen: Ahh.. sorry. I completely missed the `regex.escape()`, even though I copy & pasted it. I just didn't look for it / see it.

tripleee:
@makyen Is the latest change sufficient for the time being? If so I'll rebase and merge.

PR #3823, opened 2020-04-23
Update tld package to 0.12

makyen:
### Summary:
* Changes user_requirements.txt to `tld~=0.12`
* Removes SD's separate parsing of the TLD file, compensating with additional checks using the tld package
* Maintains compatibility with tld 0.90, so SD instances runners don't *have* to run

        pip3 install --user -r user_requirements.txt --upgrade
  after this merges. SD instance runners *should* do so, but things won't break if they don't.
* Adds additional "misleading link" tests to CI
* Adds an additional exclusion to the "misleading link" detection, so link text which is at the end of the URL is not detected (e.g. not detect detect URLs like "https://example.com/foo.txt" for link text "foo.txt")
* Fixes Travis CI so that it tests the code in PRs, rather than always testing the current master branch.

### Longer description:
As a result of a change in where the TLD list is stored on the net, the [tld package](https://pypi.org/project/tld/) has been updated to version 0.12 to fetch the list from the new URL. This is the [same issue that brought our SD instances down a few days ago](https://chat.stackexchange.com/transcript/11540?m=54134908#54134908). Merging this PR and then all SD instance runners executing:

    pip3 install --user -r user_requirements.txt --upgrade
will get us back to a bit better off than we were prior to the URL change.

Moving to `tld~=0.12` resulted in some incompatibilities with our code, so additional changes to SD were needed.

Specifically, the `tld.defaults.NAMES_LOCAL_PATH` property was removed from the package. That property was undocumented and not supported by the tld package. We were using the property to determine where the tld package stores the file containing the list of TLDs in order for SD to parse the file itself and verify that the link text we were testing contained all of a TLD preceded by a `.`. Effectively, this resulted in verifying that the link text contained a first level domain (FLD) in addition to a TLD. The only place this was being used was in the "misleading link" detection. The code for our parsing the TLD file was removed and an additional check was added to perform the same overall task, but not requiring us to parse the TLD file ourselves.

Compatibility was maintained with tld version 0.9.0, which was the version we previously specified as the minimum. This should allow this to be merged and run on existing SD instances without *requiring* that they update to tld 0.12 prior to running on this commit.

We're no longer using an undocumented feature of tld, so the package is allowed to use future releases based on the 0.* version, rather than locking it to 0.12.*.

(Not actually part of updating to tld 0.12): An additional exclusion was added to reduce FP on the misleading link detection. Link text which parses to a domain will no longer be detected, if the tail of the `href` URL matches the link text. This means that it won't detect URLs like "https://example.com/foo.txt" for link text "foo.txt".

Some additional tests were added for misleading links.

The issue with Travis always testing the master branch was identified after this PR was opened. There's more detail in my later comments.


makyen:
I'm not sure what's going on with Travis CI. The [output from Travis CI](https://travis-ci.org/github/Charcoal-SE/SmokeDetector/builds/678709296) indicates it's testing the old code, *not* the code as of the last commit in this PR. The output shows the ` test/test_findspam.py` tests without the ones which were added in [the PR's 4th commit (8057ddb)](https://github.com/Charcoal-SE/SmokeDetector/pull/3823/commits/8057ddba97964c6879907bdb74d1596e392bbcdf), and shows log output of "malicious.com: not an SE domain", which was a log output that was removed in [the PR's 2nd commit 
(872730d)](https://github.com/Charcoal-SE/SmokeDetector/pull/3823/commits/872730dffd8e539158316a0d29432b185525c876).

Basically, the Travis CI output is at least very similar to what I recall when testing the code with only the PR's 1st commit.

makyen:
So, for the CI tests, Travis only gets "collected 299 items", but Circle CI gets "collected 304 items".

I've tried running this on Travis in my fork. When the code is in a branch other than `master`, [it fails](https://travis-ci.org/github/makyen/Charcoal-SE-SmokeDetector/builds/678747850), but if I run the same thing in the `master` branch, [it passes](https://travis-ci.org/github/makyen/Charcoal-SE-SmokeDetector/builds/678753744).

makyen:
OK. Found the problem. The .travis.yml file includes the lines:
```
before_script:
  - sh test/spinup_git.sh
```

And test/spinup_git.sh includes:
```
git checkout --track origin/deploy
git checkout master
```

The history for those two files indicates those two git commands have been run since 2018-10. So, at least since that time, Travis CI hasn't been actually running on any PR. It's always been running against the master branch. It's only causing a problem here because the master branch fails with `tld` package version 0.12 and the `pip3 install -r user_requirements.txt --upgrade` is executed after Travis checks out the branch that is supposed to be tested and before test/spinup_git.sh is run.

I'll push a fix.

makyen:
@iBug @teward Just a heads-up that after this PR is merged, in order to update the tld package to 0.12, it will be desired that you `pull` and execute:

    pip3 install -r user_requirements.txt --upgrade
However, it is not *required* that you do so. Your SD instances will run with this code merged and tld versions between 0.9.0 and 0.9.8, which were the permitted versions with the old user_requirements.txt entry. Until the tld version is updated on your instance, the instance will continue to be unable to update the TLD list, as is currently the case.

teward:
> @iBug @teward Just a heads-up that after this PR is merged, in order to update the tld package to 0.12, it will be desired that you `pull` and execute:
> 
> ```
> pip3 install -r user_requirements.txt --upgrade
> ```
> 
> However, it is not _required_ that you do so. Your SD instances will run with this code merged and tld versions between 0.9.0 and 0.9.8, which were the permitted versions with the old user_requirements.txt entry. Until the tld version is updated on your instance, the instance will continue to be unable to update the TLD list, as is currently the case.

Uhhhhh... Osiris is actually running a fork of 0.11 with the changes that were included in 0.12 early.  So there's actually no need to pull until we update anything else heh.

makyen:
@teward 
> Uhhhhh... Osiris is actually running a fork of 0.11 with the changes that were included in 0.12 early. So there's actually no need to pull until we update anything else heh.

With effectively tld 0.12, there's a minor issue which causes some undesired operation in the "misleading link" code. It's enough to [cause CI to fail](https://circleci.com/gh/makyen/Charcoal-SE-SmokeDetector/603), but the effects shouldn't be all that bad in the real world. It will be resolved when this PR is merged.

makyen:
I went ahead and merged this. Having it merged should get the "misleading link" detection functional again.

PR #3840, opened 2020-05-04
findspam.py: normalize NS to lower case

tripleee:
blacklists.py: normalize NS to lower case

_normalize method added to facilitate this

test/test_blacklists.py: validation code for the above


tripleee:
https://metasmoke.erwaysoftware.com/post/239770 should have tripped the watch on `tenten.vn` but didn't because the name server query returns the result in upper case.

PR #3850, opened 2020-05-16
Proposed potential bug fixes

user12986714:
Changed those code initiating websocket into a function init_websocket; monitored whether metasmoke_ws_t is alive; changed recovery using the newly created init_websocket

Warning: this is written on a phone, and also that I am new to python, hence there can be potential issues.

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -205,14 +205,20 @@ def setup_websocket(attempt, max_attempts):
         return None
 
 
+def init_websocket(max_tries):
+    for tries in range(1, 1 + max_tries, 1):
+        ws = setup_websocket(tries, max_tries)
+        if ws:
+            break
+    else:
+        log('error', 'Max retries exceeded. Exiting, maybe a restart will kick things.')
+        exit_mode("reboot")
+
+    return ws
+
+
 max_tries = 5
-for tries in range(1, 1 + max_tries, 1):
-    ws = setup_websocket(tries, max_tries)
-    if ws:
-        break
-else:
-    log('error', 'Max retries exceeded. Exiting, maybe a restart will kick things.')
-    exit_mode("reboot")
+ws = init_websocket(max_tries)
 
 GlobalVars.deletion_watcher = DeletionWatcher()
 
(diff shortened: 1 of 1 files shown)


user12986714:
err... CircleCI is not happy with this change. Have no idea where I messed up.

makyen:
Thank you for the contribution. I'd like to accept your contribution, but you've put me in a bit of a quandary.

First off, you're really addressing two things in your PR. PRs are often better if they are focused on one thing. Just as important, if not more so, is to have commits that are laser focused on only one thing. Doing that makes reviewing significantly easier, particularly with good commit/PR comments.

I admit that there have been times when I've failed to keep my commits and PRs as narrowly focused as would be desirable, so it's something we all probably need to work on. In this case, you have two separate things in the same commit and PR. The first part is refactoring the SE WebSocket initialization code. Your implementation is as good as what I'd said [here](https://chat.stackexchange.com/transcript/message/54390035#54390035) that I'd already done (I got a bit pedantic in my change, so I probably even prefer most of yours). If that change was a separate PR, or even a separate commit, I'd probably merge it, and just make the other changes I've done on top of it.

However, you've also got code in the same commit and PR which is testing `metasmoke_ws_t.is_alive()` and logging an error when it's not alive, and does so every time a message is received from the SE WebSocket, which can be multiple times a second, but is usually less than 1/s. There are two problems with that. 

A) There's an intended, and used, operational mode where it's expected that the `metasmoke_ws_t` thread dies while the SE WebSocket scan continues (e.g. when an SD instance is used without MS, mostly for testing, but there's at least one fork of SD which is explicitly using it without MS). We *definitely* don't want more than a single log entry for that case, and then it's not an error, but might reasonably be a warning.

B) Even when we're in a mode where the `metasmoke_ws_t` thread is not supposed to die, we should not be filling up the log with a large number of error messages that the thread is (still) dead. In the cases where `metasmoke_ws_t` should be running, the log definitely should include an entry, we should even send a message to the debug room, but it shouldn't be every time we receive data from the SE WebSocket.

There are probably a few things that could be done wrt. the `metasmoke_ws_t` thread. One thing would be to adjust the `not has_succeeded` case such that more than a single attempt is made to connect to the MS WebSocket. It's also reasonable to monitor that the thread is still alive and try restarting it if both it's dead and it's not supposed to be dead.

user12986714:
New version proposed as PR #3851. Closed for reasons suggested by Makyen.

PR #3851, opened 2020-05-16
Proposed potential bug fixes

user12986714:
Changed those code initiating websocket into a function init_websocket;  changed recovery using the newly created init_websocket

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -205,14 +205,20 @@ def setup_websocket(attempt, max_attempts):
         return None
 
 
+def init_websocket(max_tries):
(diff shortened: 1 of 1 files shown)


makyen:
Thanks for trimming out the other portion.

Another time, please use a commit message that has a first line which very briefly summarizes what the commit does. Extra detail in additional lines is good, but the first line is what is seen the most. Sometimes it's the only thing shown. If it's ever necessary to go searching through the commit history, it gets really difficult to differentiate between commits if all they say is "fixed a bug" in that first line.

PR #3852, opened 2020-05-16
Added complaining to send_status_ping

user12986714:
Added complaining in chat functionality to send_status_ping: will complain if 0 post was scanned in the last minute.


user12986714:
Ah it eventually passed...

makyen:
I'm sorry, but this I'd already implemented something along these lines and like my implementation better.

A significant problem in this implementation is that you would send a message into the debug rooms upon every ping when the instance is in standby mode, or the mode where we don't scan the SE WebSocket. It also doesn't account for the periods of time where there might not be any SE WebSocket messages for more than a minute. While that's a relatively rare occurrence, it does happen.

user12986714:
Closed for reasons suggested by Mayken and potential merging conflicts.

Issue #3844, opened 2020-05-13
"English text on localized site" should ignore HTML tags


tripleee:
### What problem has occurred? What issues has it caused?

Magnificent demo: https://m.erwaysoftware.com/posts/uid/rus/460730

The post is in Russian but it used code formatting which was replaced with English text internally.

### What would you like to happen/not happen?

Don't examine code blocks for this rule?

makyen:
Note: This detection already ignores code blocks. What appears to be the issue is that the detection caught the HTML tags, not the code. I've edited the title of this issue to reflect that.

gparyani:
Duplicate of #2640 (@makyen)

Issue #2500, opened 2018-10-29
findspam.py: refactor and unify link extraction (and perhaps overall post handling)


tripleee:
There are multiple overlapping and sometimes conflicting attempts to enumerate all the links in a post in `findspam.py`.  See below for a sampling.

We should unify these, and ideally reduce the number of times we iterate over the message text looking for more or less the same information.

An object-oriented approach to the entire problem would seem like a natural but somewhat involved solution. Instead of scanning the raw text of the post over and over, use the `_Post` object we already have to store the links once, and then just use the object's methods to retrieve the links you want.

This should also make it easier to keep information about a post's features between different methods in `findspam.py` which look for distinct but related features in a post.

Just to give you an idea of the scope of the problem, here are a few of the methods which attempt to analyze links.

* [`misleading_link`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L158) looks for an `<a href=...">...</a>` snippet using a fairly straightforward regex.
* [`link_at_end`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L249) has a different regex which looks for a URL in the anchor text, with some simple whitelisting of a few domains.
* [`non_english_link`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L259) has a similar structure, with a different regex, without whitelisting.
* [`keyword_link`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L472) has a fairly simple regex for the actual link, and a messy regex over a few lines of strings for the keywords in the anchor text.
* [`bad_link_text`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L492) has a different link, more similar to the `link_at_end`/`non_english_link` regex, and a bit over half a dozen lines of messy "bad keyword" regex.
* [`bad_pattern_in_url`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/findspam.py#L525) has a different version of the `<a href=...>...</a>` regex, and two lines of bad patterns. (Confession: This one is mine.)
[...]

PR #3853, opened 2020-05-16
Ignore code blocks for English-text-on-localized-site

user12986714:
According to issue #3844 

Changed regex so any code block will be excluded when evaluating whether there is English text on localized site.

Warning: the updated regex will exclude all pattern "sometext>". May be exploitable by spammers, by appending a ">" to the blacklisted word to circumvent blacklisting.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1962,7 +1962,7 @@ create_rule("Chinese character in {}", r"(?i)\p{Script=Han}.*\p{Script=Han}.*\p{
 create_rule("Hindi character in {}", r"(?i)\p{Script=Devanagari}",
             sites=["hinduism.stackexchange.com"], body=False)
 # English text on non-English site: rus.SE only
-create_rule("English text in {} on a localized site", r"(?i)^[a-z0-9_\W]*[a-z]{3}[a-z0-9_\W]*$",
+create_rule("English text in {} on a localized site", r"(?i)^[\w\W]*[a-z]{3}(?![^<>]{0,16}>)[\w\W]*$",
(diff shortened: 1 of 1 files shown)


makyen:
Just FYI: the following regex will match valid HTML tags in HTML received from SE (and then pre-processed by SD): 

    /<((?:\/?(?:b|blockquote|code|del|dd|dl|dt|em|h[123]|i|kbd|li|p|s|sup|sub|strong|strike|ul|br|hr))|(?:\/(?:a|div|img|ol|pre|span))|(?:a|div|img|ol|pre|span)\b.*?)\s*\/?>/

Note that this is a JavaScript based regex, so you'll want to drop the `/` at the start and end. 

The above regex could be used to strip the HTML tags from the text prior to running the detection. You may need to look at the interaction of how we strip code blocks. There was an issue in how that was done for another detection. I'd need to double check...

user12986714:
There are some issues with my second commit. The regex won't work as intended and the line was too long. Decided to revert the second commit.

user12986714:
@makyen Looks like this regex won't work with negative lookahead. Yeah as you have said some magic may be needed in stripping, but a temporary solution may be disable some phrases ended with '>'. (Commit 1c89 won't work with urls, they are too long so <a href="blah blah blah very long"> won't be disabled)

user12986714:
After checking the code base, it is not a trivial task for me to use that regex in stripping without major refactoring of the entire stripping/spam detection code logic. This may be possible when resolving issue #2500, in which refactoring is proposed. (@makyen )

makyen:
Closed by request [per chat](https://chat.stackexchange.com/transcript/message/54407398#54407398).

PR #3854, opened 2020-05-16
Added lock protection

user12986714:
Added lock protection to metasmoke.py and chatcommands.py regarding to the read operation of apiquota from GlobalVars.apiquota. The lock may be needed as the write operation in bodyfetcher.py may not be atomic, and hence without lock, half-written and corrupted data may be read - the same reason why metasmoke.py in method send_statistics acquired posts_scan_stats_lock before reading post scan statistics.

### DIFF
--- a/bodyfetcher.py
+++ b/bodyfetcher.py
@@ -248,6 +248,7 @@ class BodyFetcher:
         self.api_data_lock.release()
 
         message_hq = ""
+        GlobalVars.apiquota_rw_lock.acquire()
         if "quota_remaining" in response:
             if response["quota_remaining"] - GlobalVars.apiquota >= 5000 and GlobalVars.apiquota >= 0:
                 tell_rooms_with("debug", "API quota rolled over with {0} requests remaining. "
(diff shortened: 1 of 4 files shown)


makyen:
If this is done, then bodyfetcher.py needs to be modified to release the lock during the time it sleeps for backoff, or this thread could stall waiting for the lock for over a minute, potentially an unbounded period of time (usually backoffs are 10 seconds, but they can be any value).

A better overall solution would be to implement getters and setters in GlobalVars that automatically acquire and release the locks when the values are read/written (this could be generically applied to all values stored in GlobalVars which may be changed somewhere other than in GlobalVars, or maybe, during startup elsewhere). That would make it such that the individual threads wouldn't need to worry about getting and releasing the locks, unless they wanted to implement a atomic read/modify/write.

user12986714:
(@Makyen) I am now trying to implement a separate apiquota lock. (Also that maybe we should have documentation about which lock is about which fields)

makyen:
Yeah, documentation would be good, or at least comments in the code that explain what's what. The issue of knowing which lock is for which value is one of the things that would be solved by using getters and setters (which are written to both automatically acquire/release locks) for any of the values which are permitted to be changed outside of globalvars.py (and *maybe* the startup portion of ws.py).

user12986714:
The major problem about getters/setters is that it requires a major refactoring all over the code base, as by the original design, now there are countless many direct read/write to GlobalVars fields, and hunting them down would be a huge trouble. Will try to read the code again and document those locks.

makyen:
Using getters and setters specifically means you *don't* need to hunt those down. They mean that code can just use the variables as they normally would and have the locks handled "behind the scene".

makyen:
!!/approve

PR #3855, opened 2020-05-16
Multiple retries to initiate MS websocket

user12986714:
According to [this](https://github.com/Charcoal-SE/SmokeDetector/pull/3850#issuecomment-629596302) message

Implemented multiple retries to initiate MS websocket rather than only one attempt. Also implemented logging when initiation failed.

### DIFF
--- a/metasmoke.py
+++ b/metasmoke.py
@@ -82,7 +82,8 @@ class Metasmoke:
                 if not has_succeeded:
                     failed_connection_attempts += 1
                     if failed_connection_attempts > MAX_MS_WEBSOCKET_RETRIES:
-                        chatcommunicate.tell_rooms_with("debug", "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
+                        chatcommunicate.tell_rooms_with("debug", "Cannot initiate MS websocket." +
+                                                        "  metasmoke_ws_t is now dead.")
                         log('warning', "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
                         break
                     else:
--- a/metasmoke.py
+++ b/metasmoke.py
@@ -32,6 +32,7 @@ from socketscience import SocketScience
 import metasmoke_cache
 
 
+MAX_MS_WEBSOCKET_RETRIES = 5
 MAX_FAILURES = 10  # Preservative, 10 errors = MS down
 NO_ACTIVITY_PINGS_TO_STANDBY = 8
 NO_ACTIVITY_PINGS_TO_REPORT = 4
@@ -56,6 +57,7 @@ class Metasmoke:
     @staticmethod
     def init_websocket():
         has_succeeded = False
+        failed_connection_attemps = 0
         while GlobalVars.metasmoke_key != "":
             try:
                 Metasmoke.connect_websocket()
@@ -78,7 +80,13 @@ class Metasmoke:
                 GlobalVars.metasmoke_failures += 1
                 log('error', "Couldn't bind to MS websocket")
                 if not has_succeeded:
-                    break
+                    failed_connection_attempts += 1
+                    if failed_connection_attempts > MAX_MS_WEBSOCKET_RETRIES:
+                        log('warning', "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
+                        break
+                    else:
+                        #Wait and hopefully network issues will be solved
+                        time.sleep(10)
                 else:
                     time.sleep(10)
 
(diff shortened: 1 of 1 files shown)


metasmoke.py
```diff
@@ -78,7 +80,13 @@ def init_websocket():
                 GlobalVars.metasmoke_failures += 1
                 log('error', "Couldn't bind to MS websocket")
                 if not has_succeeded:
-                    break
+                    failed_connection_attempts += 1
+                    if failed_connection_attempts > MAX_MS_WEBSOCKET_RETRIES:
+                        log('warning', "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
+                        break
```
makyen, line 86: Even though it wasn't in the original, it would be good to have a

    chatcommunicate.tell_rooms_with("debug", ...

here prior to the `break`, so that we're informed the current instance won't be connected to MS (and won't continue to try to connect to MS).

As I'm thinking about it, it might be better to not limit the number of attempts to connect (e.g. so SD auto-connects when SD is rebooted when MS is down for more than 50s; as it is and was SD would need to be manually rebooted). I'm not recalling the use-case of why we're specifically not doing so when MS is enabled. I'd have to look at the commit history for this file to see what might be explained there.

metasmoke.py
```diff
@@ -56,6 +57,7 @@ def connect_websocket():
     @staticmethod
     def init_websocket():
         has_succeeded = False
+        failed_connection_attempts = 0
         while GlobalVars.metasmoke_key != "":
```
makyen, line 61: If we're going to retry, then `while` should probably be:

    while GlobalVars.metasmoke_key != "" and GlobalVars.metasmoke_ws_host !="":

because a user may have modified the config file to remove either of those values thinking that it will disable the MS WebSocket.

user12986714:
According to review

Implemented complaining in chat functionality; adjusted while loop conditions for multiple retries.

metasmoke.py
```diff
@@ -82,6 +82,7 @@ def init_websocket():
                 if not has_succeeded:
                     failed_connection_attempts += 1
                     if failed_connection_attempts > MAX_MS_WEBSOCKET_RETRIES:
+                        chatcommunicate.tell_rooms_with("debug", "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
```
makyen: Most of the people in chat won't know or care what `metasmoke_ws_t` is, or what it means that it's dead. It would be be better to explain what this situation means to the people in chat. For example:

    "Cannot initiate MS WebSocket. A manual `!!/reboot` will be needed once MS is up."

user12986714: Would you think it is better to log metasmoke_ws_t while chat initiation-failed (or both log and chat initiation-failed)?

makyen: In this instance, we should both `log` and `tell`. However, in this location, we shouldn't talk about `metasmoke_ws_t`, because the code here is completely independent from what the Thread happens to be called in ws.py. Also, we don't want to have to remember to change a string here if we happen to change a variable name that's used on two lines in another file.

In general, when looking through a log, it's going to be a person much more familiar with the code, and how programs work, than someone who's reading things in chat. I don't know about other people, but in the log, I'm looking at cues as to where the code is currently operating and its state; mentally going through the process of running the code as I scan through what's there.

In chat, people are not expected to know the details of the code. What helps them is to know the status of SD and how a change of status affects them (i.e. what they might need to do).

user12986714:
According to review

Adjusted chat and log messages to realistic needs.

metasmoke.py
```diff
@@ -83,8 +83,8 @@ def init_websocket():
                     failed_connection_attempts += 1
                     if failed_connection_attempts > MAX_MS_WEBSOCKET_RETRIES:
                         chatcommunicate.tell_rooms_with("debug", "Cannot initiate MS websocket." +
-                                                        "  metasmoke_ws_t is now dead.")
-                        log('warning', "Cannot initiate MS websocket. metasmoke_ws_t is now dead.")
+                                                        "  Manual `!!/reboot` is required once MS is up")
+                        log('warning', "Cannot initiate MS websocket. metasmoke_ws_t is terminating.")
```
makyen, line 87: This log really shouldn't mention `metasmoke_ws_t`. Doing so puts a reverse dependency on the code this is being called from, which is in a different file/class. This function doesn't know anything about how it's being called. What if the Metasmoke class is used in some other way?

user12986714:
According to review

Adjusted log message to prevent cross dependencies. The log message now is:
Cannot initiate MS websocket. init_websocket() in metasmoke.py is terminating.

makyen:
!!/approve

PR #3856, opened 2020-05-16
Added waiting to init_se_websocket

user12986714:
Added waiting to init_se_websocket in the hope that network issues may be resolved after some time.


makyen:
!!/approve

Issue #3631, opened 2019-12-23
Fetching whitelisted domains from MS doesn't handle MS being down well


makyen:
In the most recent instance when MS was down, the active instance got stuck trying to fetch the current list of whitelisted domains. It resulted in the instance repeatedly failing and reverting to older commits. The [reboots started here](https://chat.stackexchange.com/transcript/message/53008599#53008599). The error logs are [here](https://chat.stackoverflow.com/transcript/message/48175386#48175386), with some more a couple of messages later.

Once `!!/ms-down` was executed, the instance was able to operate, but it wouldn't detect that MS was down on it's own, or recover from the reverting to older commits loop without manually executing `!!/ms-down`.

PR #3857, opened 2020-05-16
Catch exceptions in generator and fetch

user12986714:
According to issue #3631 

Added exception handler in generator() and fetch().
The handler in generator() will log the exception, and re-raise it
The handler in fetch() will terminate fetch() and return None, 'MISS-NOGEN'

This change may help to solve the problem, as the log message provided [here](https://chat.stackoverflow.com/transcript/message/48175386) indicates that the problem was caused by an uncaught exception repeatedly terminating SD, and hence triggering reversion before request_sender() in metasmoke.py was able to turn off MS.

### DIFF
--- a/metasmoke_cache.py
+++ b/metasmoke_cache.py
@@ -3,6 +3,7 @@ import datahandling
 import metasmoke
 import globalvars
 import tasks
+from helpers import log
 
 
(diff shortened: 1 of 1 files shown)


makyen:
!!/approve

Issue #3738, opened 2020-02-14
Show a dummy title to allow clicking to access posts that consist solely of non-printing characters


gparyani:
Just now, we had a spate of spam posts whose titles consisted solely of soft hyphens, which are non-printing characters. This meant that in the messages posted by SmokeDetector, the link to access the post on the site simply would not render, leaving no way to access the actual post.

Can we make it so that if a post's (or parent question's) title consists solely of characters that will not display, leading to a blank link, the title will not show, but instead a dummy title like *consists only of non-printing characters* shows up (in the same manner that we post a dummy title for offensive titles)? This will allow a clickable link to the post.

makyen:
It will be difficult for SD to determine that the browser will render the title with zero, or very little, width. From SD's POV, that title has 27 characters, which it did include in its report message. In order to determine that the link is not going to be displayed with enough width to be clickable, SD would need to know all possible ways for someone to include characters which will render with little of no width. It's possible to know some of the more common ways that the OP might do this with titles, but a complete solution is difficult.

OTOH, it's relatively easy for a userscript to determine that the width of the link is "too narrow" (i.e. below some threshold). I already have something prototyped which does this for MS. It would be not that much work to have it done by FIRE or AIM.

Would a userscript solution be acceptable?

tripleee:
Another workaround of course is to click through to the metasmoke report, then click "Visit on site" there. It's clearly not very elegant, but when workarounds exist and the logic is already complex (in the opposite direction, to make sure the message isn't too long for chat) I would be disinclined to put in additional handling for a marginal case like this.

user12986714:
Maybe we can consider to unconditionally display a dummy title, like this:
[ SmokeDetector | MS ] Reasons (Weight): Title of the post [visit] by Author on Site
where [visit] would be the dummy title.

makyen:
[This comment is copied from my comment on the above PR.]

Having SD add something to every report to compensate for this issue is noise on 99.99% of reports.

The overall issue actually affects us in two general places: SD's chat reports and on MS (on post pages and in post listings on multiple pages).

This is something which is easily handled in a userscript, or in JavaScript running on MS<sup>1</sup>, but is annoying to compensate for by adding something to every single reported post, just to save people from having to first click to MS to get a link to the post on SE, particularly when they only need to do that 0.01% of the time.

It would be possible for SD to take a *SWAG* at determining when it needs to add something to chat reports, but it's not going to be 100%.

-------
[...]

PR #3858, opened 2020-05-17
Include a [visit] link in SD reports

user12986714:
According to stale issue #3738 

Implemented [visit] link in SD reports, which links to the post. This link can be useful when the title of the post is of zero or very narrow width when rendered/displayed.

### DIFF
--- a/spamhandling.py
+++ b/spamhandling.py
@@ -169,7 +169,8 @@ def handle_spam(post, reasons, why):
 def build_message(post, reasons):
     # This is the main report format. Username and user link are deliberately not separated as with title and post
     # link, because we may want to use "by a deleted user" rather than a username+link.
-    message_format = "{prefix_ms} {{reasons}} ({reason_weight}): [{title}]({post_url}) by {user} on `{site}`"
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
This should be implemented metasmoke-side, not on Smokey's end. Moreover, it should be something _small_ if it's displaying on every single post: some common but non-ASCII, non-emoji Unicode symbol, probably.

makyen:
I agree. Adding something like this is noise on 99.99% of reports.

The overall issue actually affects us in two general places: SD's chat reports and on MS (on post pages and in post listings on multiple pages).

This is something which is easily handled in a userscript, or in JavaScript running on MS<sup>1</sup>, but is annoying to compensate for by adding something to every single reported post, just to save people from having to first click to MS to get a link to the post on SE, particularly when they only need to do that 0.01% of the time.

It would be possible for SD to take a *SWAG* at determining when it needs to add something to chat reports, but it's not going to be 100%.

-------
1. I have code in a userscript I use for MS which does this. My intent with the userscript is to test things there and then move them onto MS. I've been ... slow ... to actually get things moved to MS. It feels like every time I take some time to get something accomplished, there's yet another fire that develops somewhere.

PR #3860, opened 2020-05-19
Major refactoring of globalvars.py

user12986714:
According to [this](https://chat.stackexchange.com/transcript/message/54406786#54406786) message, perform major refactoring over globalvars.py.

Detailed changes are in commit messages.

Please don't merge this PR until all refactoring is done.

Note: draft PR won't work as it will not request review, which defeats the point.


user12986714:
flake8 is unhappy about my code...

user12986714:
I suppose it is okay to append `# noqa: N801`?

ArtOfCode-:
Ideally fix the flake8 failures instead of ignoring them, if you can - if not, add the ignore and a comment to explain why ignoring is better than fixing.

globalvars.py
```diff
@@ -129,9 +129,72 @@ class GlobalVars:
 
     api_request_lock = threading.Lock()
 
-    num_posts_scanned = 0
-    post_scan_time = 0
-    posts_scan_stats_lock = threading.Lock()
+    class post_scan_stat:  # noqa: N801
```
ArtOfCode-, line 132: Python class names are `CamelCase`, `snake_case` is for variable names

user12986714: Let's see if flake8 is happy now :-P

user12986714: Err... I did not notice the "View changes" button and conducted a forced push after renaming the class. Looks like your commits disappeared. Sorry. (@ArtOfCode- )

ArtOfCode-: I never had any commits here...

user12986714: image
I suppose that means you have made some changes to those source code? (Anyways now flake8 is happy, and I am waiting for additional reviews and then move on to commit code sitting on my computer for next inner class.)

user12986714:
There is some problem with my documentation.`#Title` won't work and need to be replaced by `# Title`.

user12986714:
Commit message still says "post_scan_stat" rather than "PostScanStat". Need to be fixed.

ArtOfCode-:
> There is some problem with my documentation.`#Title` won't work and need to be replaced by `# Title`.

I believe that's a Markdown standard - heading definitions must start with a space. It looks better, anyway.

user12986714:
circular import...

user12986714:
Added return msg to chatcommands.py to make test cases happy.
Now by running `ms-up`, one will get:
Metasmoke status is now set to up.
metasmoke is now considered up.
(Somewhat duplicate...)
Also by running `ms-up-`, one still gets:
Metasmoke status is now set to up.
(This msg is sent by Metasmoke.ms_up(), which is called every time metasmoke is switch to up except a few cases documented. This is important debug msg anyways, so better not to let `-` silence it)

user12986714:
Tests are still not happy. This is expected as in test, no room was set and hence `chatcommunicate.tell_rooms_with()`, when being called by `ms_up()`, will return an error. Hence I don't think this is a bug in the code.

metasmoke.py
```diff
@@ -62,18 +154,18 @@ def init_websocket():
                     a = GlobalVars.metasmoke_ws.recv()
                     try:
                         data = json.loads(a)
-                        GlobalVars.metasmoke_last_ping_time = datetime.utcnow()
```
user12986714, line 67: Is there really a use case of metasmoke_last_ping_time? Searching through the code base, this field does not appear to be read. Hence there is no calls to `get_last_ping()` Anyways I implemented a getter/setter to replace that. If there is no use case, maybe we can consider to completely get rid of that field, and also my `set_last_ping()` and `get_last_ping()`.

user12986714:
Anyways ignoring test results, as there is only one failure, and that is by design.

user12986714:
### This stage of refactoring is completed. Request reviewing and merging.

user12986714:
Ideally I should have refactored all api_quota/api_quota_per_site using getters/setters to enforce locking. However, the lock is implemented in PR #3854 and if implemented, can cause merging conflicts difficult to solve. Hence this PR ends here. Refactoring api_quota/api_quota_per_site is planned in the future, once #3854 get merged.

user12986714:
Completely no idea why CI failed. Local building succeeded.
Local build log:
user@pc:~/SmokeDetector1$ python3 -W default::Warning -m pytest --cov=chatcommunicate --cov=findspam --cov=globalvars --cov=spamhandling --cov=datahandling --cov=chatcommands --cov=helpers test
/home/user/.local/lib/python3.7/site-packages/_pytest/compat.py:333: PytestDeprecationWarning: The TerminalReporter.writer attribute is deprecated, use TerminalReporter._tw instead at your own risk.
See https://docs.pytest.org/en/latest/deprecations.html#terminalreporter-writer for more information.
  return getattr(object, name, default)
============================= test session starts ==============================
platform linux -- Python 3.7.3, pytest-5.4.2, py-1.8.1, pluggy-0.13.1
rootdir: /home/user/SmokeDetector1
plugins: timeout-1.3.4, cov-2.6.1
collected 304 items                                                            

test/test_blacklists.py .....                                            [  1%]
test/test_chatcommands.py ..............s........                        [  9%]
test/test_chatcommunicate.py .......                                     [ 11%]
test/test_datahandling.py ..                                             [ 12%]
test/test_findspam.py .................................................. [ 28%]
...........................................................              [ 48%]
test/test_globalvars.py .                                                [ 48%]
test/test_helpers.py ...                                                 [ 49%]
test/test_parsing.py ................................................... [ 66%]
....................................................................     [ 88%]
test/test_socketscience.py .                                             [ 88%]
test/test_spamhandling.py ..................................             [100%]

=============================== warnings summary ===============================
/usr/lib/python3/dist-packages/socks.py:58
  /usr/lib/python3/dist-packages/socks.py:58: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Callable

/usr/lib/python3/dist-packages/html5lib/_trie/_base.py:3
  /usr/lib/python3/dist-packages/html5lib/_trie/_base.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Mapping

/usr/lib/python3.7/importlib/_bootstrap.py:219
  /usr/lib/python3.7/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
    return f(*args, **kwds)

-- Docs: https://docs.pytest.org/en/latest/warnings.html

----------- coverage: platform linux, python 3.7.3-final-0 -----------
Name                 Stmts   Miss  Cover
----------------------------------------
chatcommands.py       1040    606    42%
chatcommunicate.py     407    114    72%
datahandling.py        441    210    52%
findspam.py            906    115    87%
globalvars.py          205     36    82%
helpers.py             160     72    55%
spamhandling.py        146     41    72%
----------------------------------------
TOTAL                 3305   1194    64%

============ 303 passed, 1 skipped, 3 warnings in 340.02s (0:05:40) ============

user12986714:
I see. flake8 again...

user12986714:
One conflict in metasmoke.py due to "code privilige" vs "blacklist manager". (line 423)

REFDOC.md
```diff
@@ -0,0 +1,131 @@
...
+> Note: methods return `None` unless noted otherwise.  
+
+> Warning: if the last change time indicated in this documentation is earlier than that indicated in corresponding code sections, this documentation is out of sync (and hence almost useless).
+
+# File globalvars.py  
+## Class GlobalVars.PostScanStat  
+Tracking post scanning data.  
+### Public interface  
+- `add_stat(posts_scanned, scan_time)`: Add `posts_scanned` to total numbers of posts scanned. Add `scan_time` to total time spent on scanning.
+- `get_stat()`: Get total numbers of posts scanned, total time spent on scanning and posts scanned per second. If total time spent is zero, posts scanned per second is set to `None`. Returns a tuple `(posts_scanned, scan_time, posts_per_second)`.
```
tripleee, line 17: Not a blocker, but creating excessively long lines makes it hard to edit and review this file in the future. Markdown easily lets you split source to short lines which are merged when rendering the document to HTML or other formats. Short lines which contain related phrases make for excellent diffs going forward.

tripleee:
I notice you kept `GlobalVars`, but also broke out e.g. `Metasmoke` into a separate (presumably global) object. Perhaps this should go all the way, and obey standard Python conventions for globals.

`GlobalVars.metasmoke` -> `METASMOKE`

`GlobalVars.PostScanStats` -> POST_SCAN_STATS`

etc

user12986714:
> I notice you kept `GlobalVars`, but also broke out e.g. `Metasmoke` into a separate (presumably global) object. Perhaps this should go all the way, and obey standard Python conventions for globals.

Hopefully the separate `Metasmoke` class you are referring to is the one in metasmoke.py, on line 88 in the documentation. This class already existed before this refactoring. What I actually did is encapsulated auto metasmoke status switch into a inner class Metasmoke.AutoSwitch, and added two method to the outer Metasmoke class, namely ms_up() and ms_down(). (@tripleee )

user12986714:
There is a plan in the future to completely get rid of globalvars.py/greatly reduce what is in it, and spread those variables among actual classes. Example: move all metasmoke global variables into class Metasmoke. However, this does not appear to be achievable in a short term... Current plan is to first group those variables in globalvars into inner classes (Example: this time grouped all post scan statistics data into a inner class GlobalVars.PostScanStat), and after all that is done, it is much easier to move those inner classes out of globalvars.py

makyen:
If you can avoid it, please don't merge the master branch into a PR, or a branch which you are working on. Doing so results in a [large commit](https://github.com/Charcoal-SE/SmokeDetector/commit/930e21b526f18855c58e8ff594de35234322cd05) that makes it look like you're performing a *lot* of other changes (i.e. it makes it look like this PR is changing everything that's been changed in the master branch since the PR was opened). Effectively, having such a merge commit can make reviewing the PR by looking at the overall diffs difficult, depending on the tools used. In some cases, it results in effectively requiring review by going commit by commit, which is definitely non-trivial given the history of this PR, and *may* significantly increase the amount of work for doing a review.

Merging in the master branch also: limits the options available wrt how the work is accepted (e.g. potentially makes it difficult to merge the PR into some other branch, which may, or may not, already incorporate the existing master branch); tends to result in more convoluted commit history graphs; and tends to make it not as clear as to what's going on if you need to make additional changes to the PR prior to it being merged.

If it is prior to a PR being opened (i.e. prior to your changes being public), you should almost always rebase onto the master branch, rather than merge the master branch into your branch.

Note: for many reviews, one does commonly look at each commit, particularly when those commits separate out changes made that encompass everything necessary for a particular functional change.

user12986714:
(@makyen ) Will revert that commit. Takes some time to clone the repo onto my computer.

user12986714:
Fixed

chatcommands.py
```diff
@@ -719,28 +719,45 @@ def errorlogs(count):
...
+            return status_text[1]
     # The next aliases/functionalities require privilege
     if not is_privileged(msg.owner, msg.room):
         raise CmdException(GlobalVars.not_privileged_warning)
 
     if alias_used == "ms-down":
-        GlobalVars.metasmoke_down = True
-        GlobalVars.metasmoke_failures = 999
-        return "metasmoke is now considered down."
+        Metasmoke.AutoSwitch.switch_auto(True)
```
makyen, line 742: There's a lot of redundant code and text in this next section. It seems that it would be possible to consolidate this all down into a few statements like:

```
is_force = "force" in alias_used
going_up = 'up' in alias_used
Metasmoke.AutoSwitch.switch_auto(is_force)
if going_up:
    Metasmoke.ms_up()
else:
    Metasmoke.ms_down()
return "Successfully set metasmoke status to **{}**.".format("up" if going_up else "down") +\
       " Automatic metasmoke status switch is currently **{}abled**.".format("dis" if is_force else "en")
```

globalvars.py
```diff
@@ -129,9 +128,66 @@ class GlobalVars:
 
     api_request_lock = threading.Lock()
 
-    num_posts_scanned = 0
-    post_scan_time = 0
-    posts_scan_stats_lock = threading.Lock()
+    class PostScanStat:
+        """ Tracking post scanning data """
+        # Last change was on 19 May 2020.
```
makyen, line 133: Let's not have comments which state the last change date. While knowing that information is helpful, such comments commonly become out of date. Even when such a comment exists, one has to go look at the git blame or git history to actually know if you can believe such a comment.

OTOH, it's not unreasonable to leave a comment that something changed on a date and why, if it's quite relevant to why the code is doing what it's doing.

user12986714: The hope is that this comment will be updated every time one changes code logic in the section. By doing so, we can tell if the documentation is out of sync (there is a similar comment on every section of the doc, and hopefully one will update the doc if he/she performed a major change (i.e. logic change) in code)

user12986714:
CI is failing for no reason... (again)

globalvars.py
makyen: While I understand the desire, wanting to know when something was changed is one of the reasons we have source code control systems, which maintain this type of information automatically.

The realty is that you can almost never trust that such a comment is accurate. Such comments, unless automatically maintained, will never be 100% accurate. The only way you know if it's accurate is to go to the `git` record and see when the code was last actually modified, which is exactly what you would have had to do to get the information when the comment isn't there. So, when you see such a comment, you can either ignore it, verify it for the same effort it would take to find the accurate information, or trust it and potentially have the wrong information.

user12986714:
3 conflicts. By preserving `MAX_MS_WEBSOCKET_RETRIES = 5` in metasmoke.py and deleting all other conflicting lines of master branch, one can resolve such issues.

REFDOC.md
```diff
@@ -0,0 +1,156 @@
...
+The former is a wrapper of the latter which offers logging and chat messages.
+### Implementation  
+> Not completed yet.
+### Known bugs  
+- None.
+> Last change in this section was on 20 May 2020.
+## Class Metasmoke.AutoSwitch  
+Automatically switch metasmoke status.
+### Public interface  
+- `to_off()`: Indicate a status ping failure.
```
makyen, line 114: Let's use names for `to_on` and `to_off` that more clearly indicate what they are doing. Just `to_on` and `to_off` are unclear and confusing. Maybe something like `record_ping_success_and_turn_on_ms_if_meet_criteria()` and `record_ping_failure_and_turn_off_ms_if_meet_criteria()`.

metasmoke.py
```diff
@@ -32,16 +32,103 @@
...
-    last_ping_scan_time = 0
-    last_ping_posts_scanned = 0
+
+    class AutoSwitch:
+        """ Automatically switch metasmoke status """
+        # Last change was on 20 May 2020.
+        MAX_FAILURES = 10  # 10 failures == ms down
+        MAX_SUCCESSES = 1  # 1 success == ms up
+        counter = 0
+        auto = True
```
makyen, line 48: How about something like `auto_on_off_permitted`? Just `auto` doesn't convey what it represents.

metasmoke.py
```diff
@@ -32,16 +32,103 @@
...
     status_pings_since_scan_activity = 0
-    last_ping_scan_time = 0
-    last_ping_posts_scanned = 0
+
+    class AutoSwitch:
+        """ Automatically switch metasmoke status """
+        # Last change was on 20 May 2020.
+        MAX_FAILURES = 10  # 10 failures == ms down
+        MAX_SUCCESSES = 1  # 1 success == ms up
+        counter = 0
```
makyen, line 47: How about `failure_count`?

metasmoke.py
```diff
@@ -32,16 +32,103 @@
...
+                # Negative values for success
+                current_counter = -Metasmoke.AutoSwitch.counter
+                current_auto = Metasmoke.AutoSwitch.auto
+            # MAX_SUCCESSES is constant so no lock.
+            if current_counter > Metasmoke.AutoSwitch.MAX_SUCCESSES and\
+               GlobalVars.MSStatus.is_down() and current_auto:
+                # Why use warning? Because some action may be needed if people don't think metasmoke is up.
+                log("warning", "Last {} connections to metasmoke succeeded".format(Metasmoke.AutoSwitch.counter) +
+                               " Setting metasmoke status to up.")
+                chatcommunicate.tell_rooms_with("debug", "**Warning**. {}:".format(GlobalVars.location) +
```
makyen, line 88: While having this be sent as a "warning" level for the `log` is fine, turning MS on shouldn't be presented as a warning in chat. Presumably, turning MS on is a good thing, which we should tell people about, but shouldn't warn them about it.

metasmoke.py
user12986714: The issue is that the counter is used for both success and failure (and we don't want super long names like `ping_success_failure_counter`)

REFDOC.md
```diff
@@ -0,0 +1,156 @@
...
+### Known bugs  
+- None.
+> Last change in this section was on 20 May 2020.
+# File metasmoke.py  
+## Class Metasmoke  
+> Note: the documentation of this class is not yet completed.
+### Subclasses  
+- `AutoSwitch`
+### Public interface  
+- `ms_up()`: Switch metasmoke status to up.
```
makyen, line 92: These two should probably be `set_ms_up()` and `set_ms_down()`. With them not having `set_` (or something similar), it's ambiguous as to if they change the state or return the current state.

metasmoke.py
user12986714: (and that variable is private anyway)

makyen: I'd argue that a verbose name is better than one which is short enough to be unclear what it does. I'm fine with `ping_failure_count`, `net_failure_count`, etc. One of the points of having descriptive variable names is so that code is more self-documenting and requires fewer comments to guide the reader's understanding of what's being done and why. I probably wouldn't use `ping_success_failure_count`, because having success in the name may make it seem that it should be positive for successes, whereas a negative number in a `failure_count` is hard to interpret as something other than successes.

makyen: That it's private, doesn't really matter. The point of having variable names be descriptive is so that when the next programmer comes through needing to change something, even if that's you a couple years from now, they know exactly what the variable is doing, instead of seeing `counter` and then thinking "counter? what counter? what's it counting?"

metasmoke.py
```diff
@@ -32,16 +32,103 @@
...
+                                                         " Last {} connection to metasmoke".format(current_counter) +
+                                                         " succeeded. Setting metasmoke status to **up**.")
+                Metasmoke.ms_up()
+
+        @staticmethod
+        def switch_auto(on):
```
makyen, line 94: This might be better as `permit_auto_switching(on=True)` or `permit_auto_on_off(on=True)`, etc.

metasmoke.py
```diff
@@ -32,16 +32,103 @@
...
+            """ Enable or disable auto status switch """
+            with Metasmoke.AutoSwitch.rw_lock:
+                Metasmoke.AutoSwitch.auto = on
+            if on:
+                switch_auto_msg = "Metasmoke status autoswitch is now enabled."
```
makyen, line 99: Consolidate these down to not repeating the portion of the final string which is "Metasmoke status autoswitch is now ".

metasmoke.py
```diff
@@ -224,7 +310,7 @@ def send_feedback_for_post(post_link, feedback_type, user_name, user_id, chat_ho
         if GlobalVars.metasmoke_host is None:
```
makyen, line 304: Note for the not too distant future:

In GlobalVars, we really should have something like:

    GlobalVars.metasmoke_host_valid = GlobalVars.metasmoke_host and GlobalVars.metasmoke_key
    GlobalVars.metasmoke_ws_host_valid = GlobalVars.metasmoke_ws_host and GlobalVars.metasmoke_key

and then use `GlobalVars.metasmoke_host_valid` and `GlobalVars.metasmoke_ws_host_valid` in almost all the various places where we're testing for it being possible to use MS or the MS WebSocket. As it is, we inconsistently test throughout various parts of the code for a Truthy/Falsy value, against `None`, or `!= ""` and for only one of the required pieces of data or the other.

user12986714: Hopefully we can get all metasmoke_* variables refactored into a class MSConfig and implement is_host_valid() and is_ws_host_valid().

metasmoke.py
```diff
@@ -502,43 +586,30 @@ def func(url, *args, ignore_down=False, **kwargs):
...
         in_standby_mode = GlobalVars.standby_mode or GlobalVars.no_se_activity_scan
         if not in_standby_mode:
             # This is the active instance, so should be scanning. If it's not scanning, then report or go to standby.
-
-            # We're not changing the scan stats, so shouldn't need a lock.
-            current_scan_time = GlobalVars.post_scan_time
-            current_posts_scanned = GlobalVars.num_posts_scanned
-            if current_scan_time == Metasmoke.last_ping_scan_time \
-                    and current_posts_scanned == Metasmoke.last_ping_posts_scanned:
+            if GlobalVars.PostScanStat.get_stat() == GlobalVars.PostScanStat.get_snap():
```
makyen, line 606: I'm concerned that implementing this as `PostScanStat.get_snap()` and `PostScanStat.snap()` invites those methods to be used elsewhere than in this single piece of code, when their use anywhere else will disrupt their use here. If it was me, I'd keep it just as a separate variable stored within this class. Obviously, it still needs a lock, but that can be implemented within this class.

metasmoke.py
```diff
@@ -395,23 +475,22 @@ def stop_autoflagging():
...
             payload = {'key': GlobalVars.metasmoke_key,
-                       'statistic': {'posts_scanned': GlobalVars.num_posts_scanned, 'api_quota': GlobalVars.apiquota,
+                       'statistic': {'posts_scanned': posts_scanned,
+                                     'api_quota': GlobalVars.apiquota,
                                      'post_scan_rate': posts_per_second}}
         else:
             payload = {'key': GlobalVars.metasmoke_key,
-                       'statistic': {'posts_scanned': GlobalVars.num_posts_scanned, 'api_quota': GlobalVars.apiquota}}
+                       'statistic': {'posts_scanned': posts_scanned,
+                                     'api_quota': GlobalVars.apiquota}}
```
makyen, line 491: Note to self: adjust the above `if`/`else` after it's merged and the merge conflicts are resolved. Probably to:

            payload = {'key': GlobalVars.metasmoke_key,
                       'statistic': {'posts_scanned': GlobalVars.num_posts_scanned, 'api_quota': current_apiquota,
                                     'post_scan_rate': posts_per_second}}
            if not posts_per_second:
                del payload['statistic']['post_scan_rate']

user12986714:
It is fortunate that we find out the typo (`Autoswitch` ---> `AutoSwitch`) before throwing it into production...

user12986714:
To reviewers: I squashed an change onto 6493e63 Minor improvements, namely 22337677, that fixed a typo. No need to review that file again.

metasmoke.py
```diff
@@ -32,16 +32,98 @@
...
+                                                         " succeeded. Setting metasmoke status to **up**.")
+                Metasmoke.set_ms_up()
+
+        @staticmethod
+        def enable_autoswitch(to_enable):
+            """ Enable or disable auto status switch """
+            with Metasmoke.AutoSwitch.rw_lock:
+                Metasmoke.AutoSwitch.autoswitch_is_on = to_enable
+            switch_auto_msg = "Metasmoke status autoswitch is now {}abled.".format("en" if to_enable else "dis")
+            log("info", switch_auto_msg)
```
makyen, line 100: Let's only `log` and `tell` the debug rooms if we're actually changing state. We should probably only `log` and `tell` when the state actually changes (i.e. not log when we're changing it to the same state that it was already in) for `enable_autoswitch`, `set_ms_up` and `set_ms_down`

user12986714: Hopefully. Let's have the counter issue fixed first though :-P

user12986714: As otherwise the git tree will become very messy

metasmoke.py
```diff
@@ -32,16 +32,98 @@
...
+
+        @staticmethod
+        def ping_failed():
+            """ Indicate a metasmoke status ping connection failure """
+            with Metasmoke.AutoSwitch.rw_lock:
+                if Metasmoke.AutoSwitch.ping_failure_counter < 0:
+                    # Consecutive counter. Switch sign.
+                    Metasmoke.AutoSwitch.ping_failure_counter = 0
+                Metasmoke.AutoSwitch.ping_failure_counter += 1
+                current_counter = Metasmoke.AutoSwitch.counter
```
makyen, line 60: What's `Metasmoke.AutoSwitch.counter` here? It appears to only ever be used in line 107, where it's set to 0. It looks like this (and line 107) should be `Metasmoke.AutoSwitch.ping_failure_counter`.

user12986714: Yes, this is the bug! (we literally noticed that at the same time.)

user12986714:
Looks good. Let's squash it

user12986714:
To add to wiki if this is merged:

Note that the failure count provided by`!!/ms-status` is not the same as that used in auto status switch. The former counts all consecutive metasmoke connection failures, and will become zero upon success. The latter, however, only counts metasmoke status ping successes/failures, which is performed once per minute, and goes both positive and negative, while a negative value indicates consecutive successes. There is currently no way to get the value of the latter counter. If there are good reasons this should be implemented, complain in chat.

PR #3868, opened 2020-05-21
Adjusted auto reboot message

user12986714:
Appended instance location to auto reboot message string to avoid confusion.


user12986714:
Demo: https://chat.stackexchange.com/transcript/message/54442170#54442170

PR #3871, opened 2020-05-22
Blacklist pruning

ArtOfCode-:
This PR is a significant pruning of the blacklisted websites list, removing old items, items that haven't had a hit in a while or have had lots of FPs since their last good hit.

The list of items to remove was generated by a script, which is available for review [here](https://gist.github.com/ArtOfCode-/7765eb7a1c4e92e8d5756d80486b5493); in essence, it calculates a value based on how long it's been since a good hit, how many false positives there have been since a hit, and how accurate the item has been, and if the value hits above a certain value it nominates the item for removal.

PRing this for visibility; I don't want to just drop this in on my own.


makyen:
The blacklists and watchlists definitely need this type of attention, so it's great to be doing this.

This is just my *very* rough, *initial* thoughts, and is considerably biased by my pre-existing thoughts on pruning the blacklists. I need to go through the code to see what it's actually doing wrt. how decisions are being made prior to being able to comment reasonably.

I certainly agree that entries which are getting too many FP (or which don't have enough TP) should be removed, usually with placing them on the watchlist, unless the only reason they are getting the FP is a result of us not having bookending on the website blacklist.<sup>1</sup>

#### For the blacklists, accuracy is king
To me, the overriding criteria for being on, or off, the blacklist is accuracy: is the entry maintaining a high enough %TP rate to justify being on the blacklist? If it's not accurate enough, then it should be off the blacklist. I'm having a hard time coming up with exceptions to needing to maintain a high enough %TP, there probably are some, but they are quite rare.

There are a significant number of  these entries that I'm at a loss as to why they were **ever** added to the blacklist, because they don't appear to come remotely close to qualifying, even at the time they were added.

#### Removal based on time to last detection
Unless we have a need to reduce the amount of CPU cycles which SD uses per post for scanning, I'm not sure that time since last detection should be a significant criteria. It's something to warrant additional investigation, as the domain may have become unregistered, but we've definitely seen spam domains that have been dormant for years return to spamming.

What's the goal of trimming entries which haven't gotten recent hits?

#### Moving to the watchlist
It's unclear to me if this is proposing just straight removal of all identified entries, or if the intent is to move them to the watchlist after. It currently doesn't show putting the removed entries on the watchlist, but it's possible that the intent is to add a commit doing so after the list is finalized.

------
1. In which case, they should be adjusted to have bookending.

   I'd really like to see a comparison of results between the no bookending we're currently using and using bookending that prevents entries in the website blacklist from matching domains that happen to encompass the entry. While that issue won't come close to accounting for all the FP, it's definitely been a noticeable number of those I've investigated.

ArtOfCode-:
> There are a significant number of these entries that I'm at a loss as to why they were ever added to the blacklist, because they don't appear to come remotely close to qualifying, even at the time they were added.

There's a few, particularly near the start of the list, that don't appear to have _ever_ had a hit. I suspect the cause of that, and the cause of some of the early very-low-accuracy items, is a lack of metasmoke at the point they were added.

> What's the goal of trimming entries which haven't gotten recent hits?

Seemed like a good idea at the time? :)

Part of the reason that I wrote the code with a formula for calculating a removal value instead of just writing conditions based on each value was just this. I could've said "remove everything that hasn't had a hit in 2 years", but that didn't seem right; the calculation means that if an item's getting removed for old age, it's either _really_ old and hasn't had hits in 4-5 years, or it's newer, but low accuracy.

user12986714:
This PR need to be either merged or closed quickly. As people keep blacklisting things, there will be an increasing amount of conflicts every day, and we'd better resolve them when we still can.

user12986714:
Script for generating watchlist lines from lines in remove.txt:
`cat remove.txt | awk '{print systime() "\tArtOfCode\t" $1}' >> watched_keywords.txt`

user12986714:
Given 22 days has past since this PR is opened, I would recommend closing it and if necessary, open a new one, due to severe merging conflicts.

PR #3875, opened 2020-05-23
Add logging when sending auto comments failed

user12986714:
In the hope that this will find out where the bug is.

### DIFF
--- a/metasmoke.py
+++ b/metasmoke.py
@@ -446,8 +446,12 @@ class Metasmoke:
                       "text": msg[:1].upper() + msg[1:],  # Capitalise the first letter of the comment
                       "chat_user_id": user.id,
                       "chat_host": user._client.host}
-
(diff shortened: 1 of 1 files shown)


makyen:
This won't provide useful information unless you're running the instance yourself and can look at the console output. `log` just outputs to the console, it doesn't store the data into what's provided by `!!/errorlogs`.

The method that you're looking for is `log_exception()`. However, if it's previously raised an exception, then it should already have been logged, but I'm not seeing an exception in the log where it should be for the most recent case.

makyen:
To be a bit more clear: It would be good to have additional logging here. However, the `except` shouldn't swallow the exception. It can do some logging, but the exception should be re-raised, or just allow the exception to be caught by the general exception handling (which *should* already be logging just the exception).

If there's going to be a `else: log()` then it should include output that indicates why it wasn't possible to send the autocomment (e.g. no `url` and no `ids`, or just no `response`), not just that it couldn't be sent.

PR #3876, opened 2020-05-23
Tell user why pattern is invalid

user12986714:
Adjusted CmdException message for do_blacklist(), to show what pattern it is processing that triggered the error, for debug reasons. Theoretically, this error should never be triggered, as the command requires one argument and hence must be in the form "!!/blacklist-* some_regex", and hence the split() call will return two items and item [1] can be accessed as usual. The fact that an IndexError is raised indicates that msg.content_source contains no whitespace, which means that no argument is provided (hence there is some bug in other part of the code, likely in chatcommunicate.py)

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -323,7 +323,8 @@ def do_blacklist(blacklist_type, msg, force=False):
     try:
         pattern = rebuild_str(msg.content_source.split(" ", 1)[1])
     except IndexError:
-        raise CmdException("An invalid pattern was provided, please check your command.")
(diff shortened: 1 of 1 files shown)


makyen:
This is good, but what you're sending back isn't the pattern. It's the full command. In that location, the problem is that we were unable to find the pattern. Also, it should be past tense, not present tense in order to match the first sentence. So, it should be "The command was", instead of "The pattern is".

makyen:
Based on the chat discussion starting [here](https://chat.stackexchange.com/transcript/11540?m=54465107#54465107), I substantially changed this.

Issue #3874, opened 2020-05-23
Reporting reasons are no longer being added as comments on MS


gparyani:
Recently, #2398 was implemented, which automatically adds reporting reasons (i.e. `!!/report [post link] "reporting reason"`) as a comment on the relevant metasmoke record for the post.

However, this seems to have broken at some point in the last month or two.

Can this please be fixed?

makyen:
Please provide an example of this not functioning as designed.

The most recent example from CHQ appears to be [this `!!/report`](https://chat.stackexchange.com/transcript/11540?m=54454342). The two MS reports ([1](https://m.erwaysoftware.com/posts/uid/stackoverflow/61962932), [2](https://m.erwaysoftware.com/posts/uid/stackoverflow/61962991)) created from the `!!/report` contain the reporting information in the `why` data, which is the way this has always worked since it was implemented.

gparyani:
@makyen That's how it *used* to work prior to the implementation of the linked issue. After its implementation, those used to not only be added to the `why` data, but *also* as a comment posted on the MS record on behalf of the reporting user. It is not working as designed.

See these two posts showing how it was working after that issue was implemented:

- https://metasmoke.erwaysoftware.com/post/209777
- https://metasmoke.erwaysoftware.com/post/209983

Was it de-implemented at some point?

makyen:
OK. My bad. It looks like it's supposed to be doing so, but is not.

gparyani:
@user12986714 Unfortunately, it seems your above PR didn't fix the issue. I [tried using the manual report feature just now](https://metasmoke.erwaysoftware.com/post/243998) and it didn't post as a comment.

gparyani:
Figured out how to [search](https://metasmoke.erwaysoftware.com/search?page=2&utf8=%E2%9C%93&why=with+reason) for manually reported posts with custom reasons specified.

Looking at that list, it seems that this change *never really worked* except for the two examples I gave in my prior comment here. I guess that's why you @makyen thought it was by design at first.

It seems that while that request was implemented in October 2019, it didn't work until those two reports in November 2019, and stopped working again since.

makyen:
This should be resolved.

PR #3881, opened 2020-05-25
Adjusted async comment call on report post call

user12986714:
Adjusted async comment call to actually post the custom reason for reporting.

Potentially fixed issue #3874

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1844,7 +1844,7 @@ def report_posts(urls, reported_by, reported_in=None, blacklist_by=None, operati
             comment = report_info + scan_why.lstrip()
             handle_spam(post=post, reasons=scan_reasons, why=comment)
             if custom_reason:
-                Tasks.later(Metasmoke.post_auto_comment, custom_reason, reported_by, url=url, after=300)
+                Tasks.later(Metasmoke.post_auto_comment, custom_reason, reported_by, url=url, after=15)
             continue
 
         # scan_spam == False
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -1843,7 +1843,8 @@ def report_posts(urls, reported_by, reported_in=None, blacklist_by=None, operati
         if scan_spam and operation != "report-direct":
             comment = report_info + scan_why.lstrip()
             handle_spam(post=post, reasons=scan_reasons, why=comment)
-            Tasks.later(Metasmoke.post_auto_comment, comment, reported_by, url=url, after=300)
+            if custom_reason:
```
ArtOfCode-, line 1846: I think we can probably drop the `after=300` down to some small value here - say 10.

chatcommands.py
```diff
@@ -1862,7 +1863,8 @@ def report_posts(urls, reported_by, reported_in=None, blacklist_by=None, operati
             handle_spam(post=post,
                         reasons=["Manually reported " + post_data.post_type + batch],
                         why=comment)
-            Tasks.later(Metasmoke.post_auto_comment, comment, reported_by, url=url, after=300)
+            if custom_reason:
+                Tasks.later(Metasmoke.post_auto_comment, custom_reason, reported_by, url=url, after=300)
```
ArtOfCode-, line 1867: Likewise, this can be adjusted to the same value.

chatcommands.py
user12986714: I am also confused why the original implementation would like to wait for 5 minutes... I am not even sure how the original implementation managed to work as intended. It should not, as it passes the comment data instead of the custom_reason data to the post_auto_comment() call, hence should post a comment like "manually reported by *user* in room *1234* for reasons *reason*" instead of "reason".

user12986714: However I do believe that there are some good reasons to wait that long when people first implemented it.

gparyani: I believe the threshold should be reduced to 30 or lower, per ArtOfCode.

user12986714:
Reduced waiting time to 15 seconds according to comments.

makyen:
!!/approve

Issue #2795, opened 2019-03-27
!!/allspam should not resubmit posts which were recently reported


tripleee:
Example;

https://chat.stackexchange.com/transcript/message/49665526#49665526

I submitted `!//allspam` to sweep up the remaining posts by this spammer which had not already been reported ... but the command ended up rereporting the two which had also been reported just recently, resulting in a "which post did you mean" disambiguation page in Metasmoke for those two.

* https://metasmoke.erwaysoftware.com/posts/uid/superuser/1300995
* https://metasmoke.erwaysoftware.com/posts/uid/superuser/1300983

tripleee:
Again: https://chat.stackexchange.com/transcript/message/51648522#51648522

PR #3882, opened 2020-05-25
Check recent reports in allspam

user12986714:
Implemented check for recent reports before a post is reported in allspam, to prevent duplicate reports.
Potentially fixed issue #2795

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1618,6 +1618,7 @@ def allspam(msg, url):
     :return:
     """
 
+    output = []
     api_key = 'IAkbitmze4B8KpacUfLqkw(('
     crn, wait = can_report_now(msg.owner.id, msg._client.host)
     if not crn:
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -1735,13 +1736,28 @@ def allspam(msg, url):
         batch = ""
         if len(user_posts) > 1:
             batch = " (batch report: post {} out of {})".format(index, len(user_posts))
+        if has_already_been_posted(post.site, post.post_id, post.title) and not is_false_positive(
+                (post.post_id, post.site)):
+            if GlobalVars.metasmoke_key is not None:
```
makyen, line 1741: This should be:

    if GlobalVars.metasmoke_key and GlobalVars.metasmoke_host:

alues appear to be

makyen:
Rather than having a lot of special purpose logic in `allspam`, which not only fetches a list of post URLs, but also actually goes and fetches the posts, why are we not just having `allspam` fetch the list of post URLs, and then call `report_posts()`, which would share the logic that's there with `!!/report`? `report_posts()` might need some moderate adjustments, but conceptually `!!/allspam` is just "go fetch what posts this user has and `!!/report` them". It looks like the logic in SD could just be that, and it would clean `allspam()` up quite a bit.

user12986714:
Closed for reasons suggested by Makyen and the next step is outlined [here](https://chat.stackexchange.com/transcript/11540?m=54481117#54481117).

PR #3884, opened 2020-05-26
Refactor blacklists; implement !!/watch-ip, !!/watch-ns, !!/watch-asn and !!/blacklist-xx for the first two

tripleee:
Refactor `blacklists.py` to have a set of classes implementing the various blacklist types, with behavior implemented by way of mixins for the individual classes.

Based on this refactoring, simplify `chatcommands.py` and extend to support the new blacklist types.

Some related refactoring in `gitmanager.py` to let the blacklist take care of its own updates, and in `findspam.py` to correspondingly update how we query for matches.

Finally, simplify `globalvars.py` by putting all the Github-managed black- and watchlists in a single global dict. This arrangement simplifies handling of the lists as a whole, and finding the correct list based on a keyword like "watch-number".


tripleee:
I have done some testing in chat, but this should probably still be given a quick whirl by someone who has access to Smokey's Github password.

tripleee:
Tangentially, I was getting some really crazy tracebacks in https://chat.stackexchange.com/rooms/108325/room-for-tripleee-and-ad-oisee -- it looks like the line numbers are correct but the code that Python displays is from somewhere else in the file. I'm pretty sure this is unrelated to any changes I made.

user12986714:
Those test cases were angry at your new code (because they only tests the old implementation). Ideally we can adjust that or just delete those unhappy ones.

tripleee:
Gaaaah, still need to update test cases /-:

user12986714:
Does not appear that I can open a PR over a PR, yet we need to change all bisect related functionalities so they use the new blacklist.

tripleee:
I generally use an editor window which is 80 characters wide. Keeping lines short also improves usability on Github etc.

tripleee:
I have tried to test this but there really seems to be no way to commit blacklist changes without touching the production system. (I have been planning for a long time to refactor the Github code so you could override the repo URL etc, but that's outside the scope of this PR.) My proposal at this point would be to merge this and then deal with any fallout - plastic knives and all that. Any yeas or nays?

makyen:
The repo which SD uses is already configurable/automatic. I routinely use that ability to run tests against my fork of the SD repo, including operation of `!!/watch` / `!!/blacklist-*` commands.

The repo URL is automatically picked up from the URL defined for the `git` `origin`, using the line:

    git_url = git.config("--get", "remote.origin.url").strip()
in [globalvars.py](https://github.com/Charcoal-SE/SmokeDetector/blob/master/globalvars.py#L25), if that URL starts with `https://github.com/`. This can be overridden by the `SMOKEDETECTOR_REPO` environment variable.

All you have to do is create the directory you're using for SD testing as a clone of the repository you want to use:

    git clone https://github.com/{my clone of the SD repo}

user12986714:
@tripleee  Since PR 3951 is merged, you may want to adjust `gitmanager.py: _anchor()` function accordingly after this is merged using the newly implemented *list dict.

makyen:
I still need to go through this PR and do a detailed inspection of the code.

However, prior to being merged, the commands are going to need to be tested on an SD instance running against a fork of the SD repo. However, there was a security issue with SD which could disclose the GitHub password in chat in response to a `!!/errorlogs` command. That issue has been resolved, and the password changed, but that fix is in a newer commit than this PR is based upon.

I would be more comfortable doing that testing if the code here was rebased to the current SD `master` `HEAD` (would require resolving merge conflicts and a `--force` push), or otherwise incorporates the changes which have been made (e.g. merging master into this PR: still requires resolving merge conflicts). My personal preference is rebase, but that's really a personal preference, based largely on GitHub not giving a good diff for PRs after the master has been merged.

@tripleee I'm happy to coordinate testing running in an instance against a fork of the SD repo. If you'd like, we can pick a time when we are both available.

user12986714:
For "dead code" one: really, we can just insert a raise RuntimeError('Not dead yet') clause at the beginning of that function, and during tests we can see if the code is really called nowhere. grep -rnis may also be a good idea.

tripleee:
@Daniil-M-beep points out that bisect and the `sdc` prefix are broken so some work still remains to be done.

tripleee:
Actually adding network checks to `!!/bisect` is outside the scope of this ticket. Perhaps it should be done, but not in this PR.

Daniil-M-beep:
> Actually adding network checks to `!!/bisect` is outside the scope of this ticket. Perhaps it should be done, but not in this PR.

I disagree, you are making a major change by adding 5 commands, this needs to be done properly including bisect.

user12986714:
> > Actually adding network checks to `!!/bisect` is outside the scope of this ticket. Perhaps it should be done, but not in this PR.
> 
> I disagree, you are making a major change by adding 5 commands, this needs to be done properly including bisect.

I don't really see the point here. The logic implemented in `!!/bisect` is highly regex list specific, and implementing it for network lists like cidr or as is not a simple code change.

However, we are certain that the commands already implemented in this PR are going to be very useful. I personally think it is far more important to get things done first and then improve them, rather than trying to make things perfect in the beginning.

tripleee:
Force pushed with fix for lost `sdc` change and some new tests to cover that, as well as slight refactoring of the `sdc` logic.

Daniil-M-beep:
I don't think we should be doing multiple changes in one PR. For example, you refactored the `sdc` logic who should be done separately as it doesn't relate to this.

Daniil-M-beep:
You're also refactoring a lot of existing code which I'm not a fan of since this should be done separately.

teward:
If you are doing a large code refactor, **do not** do the major refactoring of many many modules in a singular PR please.  Instead, only refactor the code you *must* refactor and refactor other code in separate PRs.  Otherwise to revert a change that is as large as this, we have to revert the entire code commit change if you break something.

!!/reject

ArtOfCode-:
Agree, can we cherry-pick off commits that aren't related to this PR to another branch and PR them separately?

tripleee:
Any particular tests you would still like to see? I will go ahead and merge without reviews next week before this goes completely stale again.

Daniil-M-beep:
@tripleee I'd like to see that is actually works

tripleee:
> @tripleee I'd like to see that it actually works

You have tested it yourself, what more are you looking for? Is there a particular test case you'd still like to implement, or try in a private room?

tripleee:
I would like to merge this before it bit rots completely. Unless somebody wants to propose a different timeline, I'll proceed and merge this next week, probably early next week.

tripleee:
Observations:
* `!!/watch-ns foo.bar.` requires the final dot
* You still can't watch an NS pair /-:
* The commit message somewhat confusingly doesn't tell you what kind of watch or blacklist this is; probably this should be changed in all watch/blacklist commit messages

tripleee:
This was backed out in #5113 because of some lingering issues. I will create a new PR once I have fixed those .

Issue #3243, opened 2019-07-22
Heavy disk write in long run


iBug:
I set up a local instance (iBot) of Smokey in a Docker container, and recently I noticed unintelligibly high disk write amount (read from `Block I/O` column of `docker stats`). At present, the container that's been up for 7 days has accrued a total block write amount of more than 150 GiB, which divides into roughly 20 GiB per day.

Upon closer inspection, I noticed that the file `bodyfetcherQueueTimings.p` was written for once every few seconds, and at the time it had a size of 2.1 MiB. For every update its size grew by a few bytes. Given that it's written once for every API call to fetch posts from SE and that we're using around 10k of our daily quota, it's most likely the culprit for such a huge write amount.

Per my knowledge, that pickle file doesn't have any use for Smokey's primary functionalities (it's for data analysis by human), and this file alone is wearing out SSDs very quickly (one Samsung 970 EVO 500 GB model in less than 3 years if unattended, calculating from raw bytes). Should we take some measures to reduce unnecessary hardware workloads?

ArtOfCode-:
That file does get written a lot, but even so, 20G/day seems excessive. Not sure what's going on there.

iBug:
@ArtOfCode- Referencing [angussidney](https://chat.stackexchange.com/transcript/message/51108310#51108310), we're doing 10k API calls each day, and each API call writes the file once, so 2 MiB * 10k getting 20 GiB is rather accurate.

ArtOfCode-:
Sure, but 2M per write also seems excessive. I'm sure it's not meant to be logging that much data.

ArtOfCode-:
_Definitely_ excessive. Each API call should only be logging a few values - the number of seconds spent in the queue for each post in that site's queue, which is no more than a handful. That's something like 100 _bytes_ per API call, not 2 MiB. Something's wrong here.

angussidney:
Every time `make_api_call_for_site` is run, it logs the queue timings to disk using the `store_queue_timings()` method: 

https://github.com/Charcoal-SE/SmokeDetector/blob/master/bodyfetcher.py#L154

This then calls the `_dump_pickle` method:

https://github.com/Charcoal-SE/SmokeDetector/blob/master/datahandling.py#L333

which passes the `"wb"` option to the built-in `open()` function. According to the [Python documentation](https://docs.python.org/3/library/functions.html#open), this truncates and overwrites the entire file.

[...]

PR #3886, opened 2020-05-26
Use plain text for queue timings data

user12986714:
Used plain text format for storing queue timings data, rather than pickle, to reduce unnecessary disk I/O.

Potentially solved issue #3243 .


user12986714:
One conflict in bodyfetcher.py with master branch. Simply deleting the conflicting code in master branch would resolve this issue.

user12986714:
@ArtOfCode- I believe that it is generally not a good idea to merge master into a PR, as learned [here](https://github.com/Charcoal-SE/SmokeDetector/pull/3860#issuecomment-633127220). Therefore I reverted your last commit. If there are good reasons that I am not aware of for doing so, you may want to reset HEAD to `1e74aae6745b10fb44f429bba075d1d03faf880b` and force push to recover it.

ArtOfCode-:
Merging master into a PR is how you resolve merge conflicts. Removing that commit means there are now merge conflicts with this PR again; you're welcome to resolve them another way.

makyen:
This currently appears to grow the bodyfetcherQueueTimings.txt log file without bound, and without handling errors or potential corruption in the file. At an absolute minimum, we need to have a handle on how rapidly this will consume disk space, and some way for the file to be truncated or reset. As to how much disk space it will use, I plan to run this on an instance for a few hours to get an idea of the amount of space used.

Note: My instances already encountered a disk space issue, which has been temporarily alieviated on my end. I have a planned change to SD, which should solve that separate issue, but which I still need to test.

user12986714:
Given the limit in API quota, SD will make no more that 20k requests per day. As seen in queuing config, each API call should log no more than 10 lines of data. Each line of data, given the worst case, would not be exceeding 100 Bytes. Hence the disk write cost for one day is no more that 20 MB, which is a significant reduce compared to 2 MB *per API call* in the previous implementation.  

As for the disk space problem, the plain text format will not be occupying order of magnitude of more disk space than the serialized pickle. Hence it is expected that in this implementation the disk space cost per day is no more that 9 times of that of the original implementation, and my personal guess is no more than 1.2 times, which is not really a severe issue.  

It is also true that logs, on POSIX, is cleaned manually by an admin or by a cron job, not by the application generating it. Implementing the feature of log size control would add unnecessary logic to the code, as cron is already doing a good enough job in automating log size control.  

Stepping back, given the calculation in the first paragraph, which indicates no more than 20 MB writes per day, effectively tells us that for 1 GB of disk space SD may run at least 50 days. Given that on modern computers, even on RPis, it is common to have disk space of several GBs, I don't really see the point for auto log size control.

makyen:
This is currently broken. The merge of the master branch into this one broke it (posts are not scanned).

This can be reopened once the problem is resolved.

PR #3899, opened 2020-06-03
Implemented chat message queue dumping on exit

user12986714:
Implemented chat message queue dumping on exit. Upon restart, chat messages that were not posted in the last session will be restored.


makyen:
In general, I'd appreciate more comments that explain what your *intent* is. In reviewing, I shouldn't need to dig through the code cold, having to figure out how what you are implementing is intended to accomplish your overall goal; nor should it be necessary to dig through each section of code or function in order to get a general overview of what is supposed to be happening. It doesn't take a lot of comments, but there should be some to guide the person reading the code as to what is intended to be happening, and in general how that's being accomplished.

PR #3929, opened 2020-06-05
Entry "rock\W?hard(?:\W?bull)?" removed from blacklist

user12986714:
Too many fps. Move to watchlist instead.


NobodyNada:
!!/approve

PR #3941, opened 2020-06-07
Minor changes in chatcommands.py: Added comments for threads

user12986714:
Added comments for threads, so next time someone runs this command, they do not need to figure all this out again.
"message sender" is started by SD.
"message_sender" is started by CE, one for each server.


makyen:
!!/approve

PR #3942, opened 2020-06-07
Split link-at-beginning watch into new reason

NobodyNada:
[Discussed in CHQ](https://chat.stackexchange.com/transcript/11540?m=54586828#54586828). The reason has its own whitelist for now, until we have unified whitelist-checking.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -2265,6 +2265,14 @@ create_rule("potentially bad keyword in {}",
             all=False, sites=["workplace.stackexchange.com", "workplace.meta.stackexchange.com"],
             username=True, body_summary=False, body=False, title=False,
             max_rep=93, max_score=1)
-
+# Link at beginning of post; pulled from watchlist
+create_rule("link at beginning of {}",
+            r'^\s*<p>\s*(?:</?\w+/?>\s*)*<a href="(?!(?:[a-z]+:)?//(?:[^" >/.]*\.)*(?:(?:quora|medium'
+            r'|googleusercontent|youtube|microsoft|unity3d|wso|merriam-webster|oracle|magento|example'
+            r'l|apple|google|github|imgur|stackexchange|stackoverflow|serverfault|superuser|askubuntu)\.com'
+            r'|(?:(?:lvcharts|php|jsfiddle|mathoverflow)\.net)|github\.io|youtu\.be|edu|(?:(?:arxiv|drupal'
+            r'|python|isc|khronos|mongodb|open-std|dartlang|apache|pydata|gnu|js|wordpress|wikipedia)\.org))'
+            r'[/\"])\W*(?![\W\w]*?</(?:code|blockquote)>)',
+            title=False, username=False, body=True)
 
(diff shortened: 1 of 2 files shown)


user12986714:
If we are implementing this as a separate rule (which I think we should), I would just duplicate the code used for link-at-end, which takes whitelists into considerations.
```python
@create_rule("link at end of {}", title=False, all=False, sites=[
    "superuser.com", "askubuntu.com", "drupal.stackexchange.com", "meta.stackexchange.com",
    "security.stackexchange.com", "patents.stackexchange.com", "money.stackexchange.com",
    "gaming.stackexchange.com", "arduino.stackexchange.com", "workplace.stackexchange.com"])
def link_at_end(s, site):   # link at end of question, on selected sites
    s = regex.sub("</?(?:strong|em|p)>", "", s)
    match = regex.compile(r"(?i)https?://(?:[.A-Za-z0-9-]*/?[.A-Za-z0-9-]*/?|plus\.google\.com/"
                          r"[\w/]*|www\.pinterest\.com/pin/[\d/]*)(?=</a>\s*$)").search(s)
    if match and not is_whitelisted_website(match.group(0)):
        return True, u"Link at end: {}".format(match.group(0))
    return False, ""
```

NobodyNada:
See the CHQ discussion; my understanding is that this reason will eventually adopt the MS whitelist once we have the new whitelist-checker.

makyen:
This should have a `max_rep` and a `max_score`. The only data we have is for what the watchlist is using, which is `max_rep=32, max_score=1`.

findspam.py
```diff
@@ -2265,6 +2265,14 @@ def format_with_city_list(regex_text):
             all=False, sites=["workplace.stackexchange.com", "workplace.meta.stackexchange.com"],
             username=True, body_summary=False, body=False, title=False,
             max_rep=93, max_score=1)
-
+# Link at beginning of post; pulled from watchlist
+create_rule("link at beginning of {}",
+            r'^\s*<p>\s*(?:</?\w+/?>\s*)*<a href="(?!(?:[a-z]+:)?//(?:[^" >/.]*\.)*(?:(?:quora|medium'
+            r'|googleusercontent|youtube|microsoft|unity3d|wso|merriam-webster|oracle|magento|example'
+            r'l|apple|google|github|imgur|stackexchange|stackoverflow|serverfault|superuser|askubuntu)\.com'
```
makyen, line 2272: There's a minor editing error in this line. It should start with `|`, not `l`. The first letter `l` is extra.

NobodyNada: Oops, good catch!

NobodyNada:
@makyen Thanks. I went ahead and put in a max_rep and max_score for now; I've noticed enough false-positives from this that I think it's worthwhile.

makyen:
I agree, I'd rather have the `max_rep` and `max_score`, at least for now. Maybe once we have the whitelist working, it will reduce the FP rate sufficiently such that we can look at adjusting those, if it's felt there's a need.


!!/approve

PR #3955, opened 2020-06-08
location command redirected to rev

user12986714:
The latter is a super set of the former. Doing so may also reduce the work in future maintenance/refactoring/development, as LoC is reduced and there is one less place to search/replace in case there is a change.


iBug:
Agreed. LGTM.

In fact, I've always thought about this.

PR #3961, opened 2020-06-08
chatcommands.py: Implemented new commands stat and counter

user12986714:
Implemented new chat commands stat and counter.
- stat: get post scan statistics.
- counter: get metasmoke ping failure counter.

counter is to be distinguished from ms-status.


makyen:
!!/approve

PR #3971, opened 2020-06-10
Use plain text for queue timings data

user12986714:
Used plain text format for storing queue timings data, rather than pickle, to reduce unnecessary disk I/O.

Potentially solved issue #3243 .


bodyfetcher.py
```diff
@@ -46,7 +46,6 @@ class BodyFetcher:
     api_data_lock = threading.Lock()
     queue_modify_lock = threading.Lock()
     max_ids_modify_lock = threading.Lock()
-    queue_timing_modify_lock = threading.Lock()
```
NobodyNada, line 49: I think we still want to keep the lock. AFAIK writing to a file is **not** atomic, so if two threads try to add timing data simultaneously we could have some wonderful and fun concurrency problems.

datahandling.py
```diff
@@ -340,8 +338,12 @@ def store_bodyfetcher_max_ids():
     _dump_pickle("bodyfetcherMaxIds.p", GlobalVars.bodyfetcher.previous_max_ids)
 
 
-def store_queue_timings():
-    _dump_pickle("bodyfetcherQueueTimings.p", GlobalVars.bodyfetcher.queue_timings)
+def add_queue_timing_data(site, time_in_queue):
+    with open("bodyfetcherQueueTimings.txt", mode="a", encoding="utf-8") as stat_file:
```
NobodyNada, line 342: Can't we just open the file once on startup, instead of every time we write to it?

queue_timings.py
```diff
@@ -5,20 +5,21 @@
 
 import os.path
 # noinspection PyPep8Naming
-import pickle
 import math
 
 
 def main():
-    if os.path.isfile("bodyfetcherQueueTimings.p"):
```
NobodyNada, line 13: Maybe we still want to support both formats? If both files are present we can just merge the data.

bodyfetcher.py
```diff
@@ -1,7 +1,7 @@
 # coding=utf-8
```
NobodyNada, line 1: We don't use the `queue_timings` variable on line 23 anymore, right? Can we just get rid of it?

NobodyNada (changes requested):
Looks great, just a couple minor things.

datahandling.py
ArtOfCode-: We could, but that would result in Smokey holding exclusive access to the file while it's running, whereas opening on demand leaves the file available - for instance if you want to edit/correct it while Smokey is running.

user12986714: To open the file on startup means that we need to somehow close it when SD shuts down. Maybe we can use some hooks but I am not sure how to do it.

---
> but that would result in Smokey holding exclusive access to the file while it's running

It appears that (at least under linux/bsd) it is not the case, and we are not even using POSIX advisory lock. (And we can ignore the advisory lock anyway since it is not a mandatory lock.)

NobodyNada: Aren't open files closed automatically when the process exits?

makyen: How about we hold this data in RAM for a relatively short while and just write once every N times that `add_queue_timing_data()` is called, or if M time has passed since the last time which it was called (and upon SD exit)? Do we _need_ this data to never have lost some data points (i.e. do we need to be _certain_ we don't loose any data if SD hard crashes)?

NobodyNada: That seems reasonable to me -- it sure would save a lot of disk traffic.

I don't think we care too much about preserving every data point. AFAIK all we ever do with this data is average it all up (and use the results to periodically adjust queue sizes), so dropped data points shouldn't have a significant impact.

user12986714: > Aren't open files closed automatically when the process exits?

Well, (I think) data integrity is not guaranteed in such cases.

user12986714: > How about we hold this data in RAM for a relatively short while and just write once every N times that `add_queue_timing_data()` is called, or if M time has passed since the last time which it was called (and upon SD exit)? Do we _need_ this data to never have lost some data points (i.e. do we need to be _certain_ we don't loose any data if SD hard crashes)?

Do we really need that much extra control structure/logic for negligible savings in total disk I/O?

NobodyNada: > Well, (I think) data integrity is not guaranteed in such cases.

Aw, looks like you're right about that. The OS guarantees data integrity, but the Python runtime does its own buffering and because we use `os._exit` instead of `sys.exit`, Python can't guarantee those buffers get flushed.

> Do we really need that much extra control structure/logic for negligible savings in total disk I/O?

Well, the purpose of this PR is to reduce disk I/O after all. It's possible that those changes would be unnecessary because of the OS's own caching/buffering, but I'm not sure how much we can rely on that. Opening a file, writing a single line, and closing a file for *every post we scan* is a lot of work.

ws.py
```diff
@@ -86,6 +87,8 @@
 if not GlobalVars.metasmoke_ws_host:
     log('info', "No metasmoke websocket host found, which is okay if you're anti-websocket")
 
+# Register actually_add_queue_timings_data hook
+atexit.register(actually_add_queue_timings_data)
```
NobodyNada, line 91: `atexit` unfortunately isn't going to work here, because we exit Smokey with `os._exit` which bypasses `atexit` handlers (instead of using `sys.exit`). I'm not sure why we use that...I'd suggest changing it to `sys.exit`, if not for [Chesterton's fense](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence).

We probably want to be sure to actually write the batches of data every ~100 entries or so, so that even if we crash we don't lose *too* much data.

NobodyNada (changes requested):

ws.py
NobodyNada: 👍 If you're changing the exit call, be sure to change the one in [`metasmoke.py`](https://github.com/Charcoal-SE/SmokeDetector/blob/549ae6ba8582821cf4fa0a27b4aea23ed445ed1c/metasmoke.py#L230) too.

metasmoke.py
```diff
@@ -212,7 +212,7 @@ def handle_websocket_data(data):
             chatcommunicate.tell_rooms(event["message"], ("debug", "site-" + event["site"]),
                                        ("no-site-" + event["site"],), notify_site="/autoflag_fp")
         elif "exit" in message:
-            os._exit(message["exit"])
+            exit_mode(message["exit"])
```
NobodyNada, line 215: I think this should be `exit_mode(code=message["exit"])`

NobodyNada (changes requested):

NobodyNada (approved):
Looks great! I'll test this when I get a chance, but that may not be for a few days.

tripleee:
@NobodyNada Do you want to test something still?

NobodyNada:
@tripleee No thanks, I don't remember what I wanted to test :) If y'all are working on merging this and need someone to test it, I'm happy to help out though.

tripleee:
I just noticed that this has been pending for almost a year, and figured we should do something about it. Given the fairly detailed background in the linked bug report, I guess it should not be hard to assess whether this reduces writes. Do you want to volunteer, or shall I?

makyen:
I haven't looked at this in quite a while, and don't remember what changes were made to the code.

Using text to represent this data in the file will be beneficial to reduce the overall number of writes/erasures on the media. It would be good to also (or alternately) just not wrtting the data every single time we scan posts. Writing the data every 2nd time would reduce the amount written by 2x. Writing the data every 10th time would reduce the amount written by an order of magnitude.

The main issue is writes on a FLASH based storage medium. With FLASH, there will be a minimum block size which must be erased and rewritten at the hardware level. In many/most cases, writes are not aggressively cached by the system, so at the end of our writing the information the data will be flushed from whatever system level cache exists to the media. So, even if we are writing only a single line of text, the underlying media will still be doing a [read larger block with partial data]-> [modify what's in that block to add the line] -> [write modified larger block into a different location] (the old block is typically marked as trimmed and the physical device erases it asynchronously, as the erase process takes substantially more time).

NobodyNada:
I've got a bit of time today, so I'll give it a smoke-test to verify that it doesn't break things and does indeed reduce disk activity.

NobodyNada:
There were a few merge conflicts, so I took care of those. There were also a couple bugs:

- We really did use `os._exit` for a reason: someone thought it would be a good idea for`sys.exit` to exit only the current *thread*, not the current process. This means we can't use `atexit` handlers and must put our cleanup code in `exit_mode` (20117c3).
- This code missed writing a newline at the end of each batch of timings (15db8db)

With these fixed, this PR appears to work well and greatly reduce I/O traffic. The fixed code is available at my fork: https://github.com/NobodyNada/SmokeDetector/tree/asciitiming

makyen:
@NobodyNada Please submit a new PR with the additional changes to be merged into the  "PR3971-plain-text-queue-timings" branch.

NobodyNada:
@makyen Done: #6705

PR #3984, opened 2020-06-13
vision[\W_]*+20 removed from blacklist

user12986714:
vision[\W_]*+20 removed for low accuracy 0.389. Should be moved to watchlist.


makyen:
!!/approve

PR #4006, opened 2020-06-17
chatcommunicate.py: Ignore reply command if provided too many arguments

user12986714:
Those replies are unlikely intended as actual commands.


user12986714:
Maybe :-/

PR #4040, opened 2020-06-21
Blacklist dns?.namecheaphosting.com

user12986714:
Enough tp on MS.


iBug:
There were enough FPs (~3%) for this NS. We want to keep it watched, not blacklisted. (cc @tripleee)

PR #4042, opened 2020-06-21
bad_keywords.txt: Allow plural in support-number spam

user12986714:
As a preventive method

### DIFF
--- a/bad_keywords.txt
+++ b/bad_keywords.txt
@@ -44,7 +44,7 @@ service\W?solahart
 junisse
 profactor\W?t
 (?:fake|novelty|quality|buy(?:ing)?|sell(?:ing)?|offer|comprar|fabricante|vender|venta).{0,20}(?:passport|driver'?s? licen[cs]e|ID card|green card|residence permit|toefl|Ielts|pasaportes|licencias[\W_]*+de[\W_]*+conducir|tarjetas[\W_]*+de[\W_]*+identificación)s?
-(?:support|service|helpline)(?:\W?phone)?\W?number(?<=^.{0,225})
+(?:support|service|helpline)(?:\W?phone)?\W?numbers??(?<=^.{0,225})
 [a-z_]*+(?:1_*+)?866[\W_]*+978[\W_]*+(?:6819|6762)[a-z_]*+
 (?:mcafee|hotmail|gmail|outlook|yahoo|lexmark (?:printer)?) ?(?:password(?: recovery)?|tech)? ?(?:(?:customer|technical) (?:support|service))? (?:support|contact|telephone|help(?:line)?|phone) number
 kitchen for sale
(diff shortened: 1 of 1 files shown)


bad_keywords.txt
```diff
@@ -44,7 +44,7 @@ service\W?solahart
 junisse
 profactor\W?t
 (?:fake|novelty|quality|buy(?:ing)?|sell(?:ing)?|offer|comprar|fabricante|vender|venta).{0,20}(?:passport|driver'?s? licen[cs]e|ID card|green card|residence permit|toefl|Ielts|pasaportes|licencias[\W_]*+de[\W_]*+conducir|tarjetas[\W_]*+de[\W_]*+identificación)s?
-(?:support|service|helpline)(?:\W?phone)?\W?number(?<=^.{0,225})
+(?:support|service|helpline)(?:\W?phone)?\W?numbers??(?<=^.{0,225})
```
NobodyNada, line 47: Are there supposed to be two question marks there?

user12986714: Lazy match?

user12986714:
```
>>> import regex
>>> a = regex.compile(r'aa??b')
>>> a.match('a')
>>> a.match('ab')
<regex.Match object; span=(0, 2), match='ab'>
>>> a.match('aab')
<regex.Match object; span=(0, 3), match='aab'>
```

NobodyNada: Thanks, I didn't know about that. Does it make a difference in this case? It seems to me that the performance would be the same regardless of whether you check "number" or "numbers" first, but I'm probably missing something.

user12986714: Literally negligible performance difference... Just out of habit :-P

NobodyNada:
!!/approve

bad_keywords.txt
user12986714: Oh no... This will decrease performance quite a bit.
If the string is `numbers`, it will first be matched as `number`, and then a lookbehind, and then `numbers`, and another lookbehind.
If one does not use lazy matching, such issue won't exist.

NobodyNada:
!!/approve

bad_keywords.txt
makyen: Please use `s?+` (possessive) there, so the look-behind is only tested once at that spot. In that spot, a lazy match is never needed. The "s" either is, or is not, present. If it is present, then it *must* be included in the match, in order to match the ending `\b` (i.e. we know apriori  that it's not possible to match the rest of the regex if an "s" is present and given back after a failed match).

As long as you are adjusting the line, it would probably be good to adjust the `\W?` to `[\W_]*+`.

makyen:
Closure was because the page didn't update showing the additional commit. Please change that `s?` to `s?+`.

PR #4044, opened 2020-06-21
Allow plural for support spam websites

user12986714:


makyen:
It looks like it would be reasonable to have `recover(?:y|ies)` be `recover(?:y|ies)?+`.

PR #4074, opened 2020-06-24
Send post Markdown from SE API to MS

iBug:
Context: Charcoal-SE/metasmoke#734

This PR implements the Smokey side of stuff. Apparently depends on the deployment of the target MS PR.

---

Changes to filters: Added `body_markdown` for questions and answers where `body` is already selected.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1731,7 +1731,7 @@ def allspam(msg, url):
         # Fetch posts
         request_url = "https://api.stackexchange.com/2.2/users/{}/posts".format(u_id)
         params = {
-            'filter': '!)Q4RrMH0DC96Y4g9yVzuwUrW',
+            'filter': '!*0S0--txpIwJHBnEqJR9px_-b(4TI_dAeSMiPUcCA',
             'key': api_key,
             'site': u_site
         }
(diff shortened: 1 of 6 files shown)


iBug:
cc @ArtOfCode- Ready for merge after MS deploy

chatcommands.py
```diff
@@ -1731,7 +1731,7 @@ def allspam(msg, url):
         # Fetch posts
         request_url = "https://api.stackexchange.com/2.2/users/{}/posts".format(u_id)
         params = {
-            'filter': '!)Q4RrMH0DC96Y4g9yVzuwUrW',
+            'filter': '!*0S0--txpIwJHBnEqJR9px_-b(4TI_dAeSMiPUcCA',
```
makyen, line 1734: This filter does not appear to contain `post/body_markdown`, which is what you are trying to get here, at least as far as I understand.

iBug: Oops, I only looked into `answer` and `question` and even wondered for a while why it included neither `answer/body` nor `question/body`.

iBug: Fixed in abd4b5011ee82b206586411b11b616ca7fdb832f

PR #4078, opened 2020-06-25
Add room for Arqade

Draakonraev:
Add chat room named Chat Restricted for feed for gaming.stackexchange.com


quartata:
!!/approve

Issue #4083, opened 2020-06-26
Let Smokey take commands as SD


Daniil-M-beep:
Currently, Smokey takes commands with `!!/` and `sd` for aliases, e.g. `k` etc.

Can we let Smokey take all `sd` commands? For me, this is much easier and quicker to type, especially on mobile.

e.g.:

`sd watch test`

`sd approve 1000`

etc.

NobodyNada:
I'm not saying that we should or shouldn't do this, but there are some complexities we do need to take into account. `sd` commands have different semantics than `!!/` commands (see [Shortcut commands](https://charcoal-se.org/smokey/Commands#shortcut-commands)). The `sd` shorthand is used to invoke commands that operate on chat messages (such as `k`, `f`, `delete`), while `!!/` is used to invoke global commands such as `!!/watch` or `!!/reboot`. Invoking a command using `sd` syntax is shorthand for replying to Smokey's most recent chat message.

The tricky bit is that `sd` syntax allows you to specify multiple commands on one line. For example, `sd k 2f - delete` would mark the most recent message as a TP, mark the second- and third- most recent messages as FPs, skip the fourth, and delete the fifth-most recent message. Because of this, commands written with `sd` syntax cannot take parameters; e.g. `sd watch test` would be interpret `watch` and `test` as two separate commands.

There may be ways around this challenge -- perhaps we could recognize non-reply commands such as `watch` and bypass the normal shorthand command logic -- but adding another special case would make shorthand command syntax a bit convoluted and harder to learn, and there might be additional edge cases I haven't thought of.

Daniil-M-beep:
Implemented: https://github.com/Charcoal-SE/SmokeDetector/pull/4096

PR #4096, opened 2020-06-28
chatcommunicate.py: Using sdc to invoke standard commands

user12986714:
Note that "sh " and "!!/" are of the same length.

Linked issue #4083

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -296,7 +296,7 @@ def on_msg(msg, client):
         if result:
             s = ":{}\n{}" if "\n" not in result and len(result) >= 488 else ":{} {}"
             _msg_queue.put((room_data, s.format(message.id, result), None))
-    elif message.content.startswith("!!/") or message.content.lower().startswith("sh "):
+    elif message.content.startswith("!!/") or message.content.lower().startswith("sdc "):
         result = dispatch_command(message)
 
         if result:
@@ -515,12 +515,12 @@ def get_message(id, host="stackexchange.com"):
 def dispatch_command(msg):
     command_parts = GlobalVars.parser.unescape(msg.content).split(" ", 1)
     try:
-        if command_parts[0] == 'sh':
+        if command_parts[0] == 'sdc':
             command_parts = command_parts[1].split(" ", 1)
         else:
             command_parts[0] = command_parts[0][3:]
     except IndexError:
-        return "Invalid command: Use either `!!/cmd_name` or `sh cmd_name`" +\
+        return "Invalid command: Use either `!!/cmd_name` or `sdc cmd_name`" +\
                " to run command `cmd_name`."
 
     if len(command_parts) == 2:
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -517,9 +517,11 @@ def dispatch_command(msg):
     try:
         if command_parts[0] == 'sh':
             command_parts = command_parts[1].split(" ", 1)
-            command_parts[0] = "!!/" + command_parts[0]
+        else:
+            command_parts[0] = command_parts[0][3:]
     except IndexError:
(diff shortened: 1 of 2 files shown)


chatcommunicate.py
```diff
@@ -296,7 +296,7 @@ def on_msg(msg, client):
         if result:
             s = ":{}\n{}" if "\n" not in result and len(result) >= 488 else ":{} {}"
             _msg_queue.put((room_data, s.format(message.id, result), None))
-    elif message.content.startswith("!!/"):
+    elif message.content.startswith("!!/") or message.content.lower().startswith("sh "):
```
NobodyNada, line 299: Has this been tested? I don't think it will work because of [the first line of `dispatch_command`](https://github.com/Charcoal-SE/SmokeDetector/blob/81075a1c743be259665c2b28f4af181a647c1758/chatcommunicate.py#L516):

    command_parts = GlobalVars.parser.unescape(msg.content).split(" ", 1)

That splits the command on the first space, separating it into the command name and arguments. If you passed in `sh somecommand args`,  the command will be parsed as `sh` and the arguments will be `somecommand args`, which is not what you want.

Next, the function then strips the first three characters off of the command (to get rid of the `!!/`), which will probably throw an exception or something since the "command" `sh` is only two characters long.

chatcommunicate.py
```diff
@@ -514,6 +514,12 @@ def get_message(id, host="stackexchange.com"):
 
 def dispatch_command(msg):
     command_parts = GlobalVars.parser.unescape(msg.content).split(" ", 1)
+    try:
+        if command_parts[0] == 'sh':
+            command_parts = command_parts[1].split(" ", 1)
+            command_parts[0] = "!!/" + command_parts[0]
```
NobodyNada, line 520: Why not just set `command_name` and `args` to `command_parts[0].lower()` and `command_parts[1]` respectively, and then have the logic for parsing `!!/` in an else statement? It would be functionally equivalent but a little easier to read IMO

makyen:
Let's not use `sh`, due to the fact that there is a Unix/Linux shell of the same name. When I first saw this, I thought you were trying to allow running arbitrary commands on the SD instance, so there's definitely room for confusion here. I also don't want to be training people's muscle memory to be typing `sh reboot`, etc.

makyen:
Because you mentioned it in a commit message:  
Allowing people to run arbitrary shell commands using the SD chat command interface isn't going to happen, because it's just too big of a security hole. Yeah, it might be convenient in some cases. Yes, running SD effectively gives people with write access to the GitHub repo the ability to run arbitrary code, but that is a *much* smaller subset of people and leaves a record on GitHub which is difficult to eliminate (you can cover over it, but it's hard to actually eliminate the record).

user12986714:
So... What I actually mean is to implement something like
```python
def grep(in, exp):
    try:
        regex_exp = regex.compile(r"^.*" + exp + r".*$")
    except Exceptions:
        raise CmdException("Invalid regex")
    in_lines = in.split('\n')
    out = list()
    for line in in_lines:
        if regex_exp.match(line):
            out.append(line);
    return "\n".join(out)
```

Daniil-M-beep:
@user12986714  I think this is a good idea. 

I tested this new code with [my personal Smokey instance](https://chat.stackexchange.com/rooms/109232/daniil-sd) and found issues with commands that used **raw data**.

Examples:

- `sdc bisect test` `searches for `bisect test` in the watchlist/blacklist instead of `test
- `sdc watch-number 123456789` tries to add `watch-number 123456789` to the `watched_numbers.txt` list as opposed to `123456789`
- `sdc blacklist-keyword test` tries to blacklist `blacklist-keyword test` to the blacklist

Happy to do any more testing after the code has been modified!

PR #4127, opened 2020-06-29
findspam.py: Partial performance optimization for regex

user12986714:
Replace capturing groups with non-capturing groups.


findspam.py
```diff
@@ -903,15 +903,15 @@ def pattern_email(s, site):
 def keyword_link(s, site):   # thanking keyword and a link in the same short answer
     if len(s) > 400:
         return False, ""
-    link = regex.compile(r'(?i)<a href="https?://\S+').search(s)
+    link = regex.compile(r'(?i)<a href="https?://[\w_-]++').search(s)
```
makyen, line 906: The `_` character is part of the word, `\w`, character class. I believe the intent here was to get the entirety of the URL, which can include characters other than `[\w-]`, so this change doesn't do what was intended.

findspam.py
```diff
@@ -903,15 +903,15 @@ def pattern_email(s, site):
...
+    praise = regex.compile(r"(?i)\b(?:nice|good|interesting|helpful|great|amazing) (?:article|blog|post|information)\b|"
                            r"very useful").search(s)
-    thanks = regex.compile(r"(?i)\b(appreciate|than(k|ks|x)|gratidão)\b").search(s)
-    keyword = regex.compile(r"(?i)\b(I really appreciate|many thanks|thanks a lot|thank you (very|for)|"
-                            r"than(ks|x) for (sharing|this|your)|dear forum members|(very (informative|useful)|"
-                            r"stumbled upon (your|this)|wonderful|visit my) (blog|site|website))\b").search(s)
+    thanks = regex.compile(r"(?i)\b(?:appreciate|than(k|ks|x)|gratidão)\b").search(s)
+    keyword = regex.compile(r"(?i)\b(?:I really appreciate|many thanks?|thanks? a lot|thank you (?:very|for)|"
+                            r"than(?:ks|x) for (?:sharing|this|your)|dear forum members|(?:very (?:informative|useful)|"
+                            r"stumbled upon (?:your|this)|wonderful|visit my) (?:blog|site|website))\b").search(s)
```
makyen, line 914: I haven't looked at the full text of the code here, but from what is visible in the GitHub diff view, `keyword.group(0)` and `link.group(0)` are used in the code. In this instance, the capture groups are actually being used, so changing every group to non-capturing is going to break things. It's likely that a number of the groups could and should be changed to non-capturing, but determining which ones would require more review time than I've spent here.

Using non-capture groups when capturing isn't needed does save processing time. However, we can only change things wholesale when it's clear that the results from the capture are not actually being used.

user12986714:
```
Help on built-in function group:

group(...) method of _regex.Match instance
    group([group1, ...]) --> string or tuple of strings.
    Return one or more subgroups of the match.  If there is a single argument,
    the result is a single string, or None if the group did not contribute to
    the match; if there are multiple arguments, the result is a tuple with one
    item per argument; if there are no arguments, the whole match is returned.
    Group 0 is the whole match.
```

makyen: So, I should have RTFM. :; Thanks.

makyen:
!!/approve

PR #4128, opened 2020-06-29
findspam.py: Anchor pattern_websites

user12986714:
So `spam.co` does not match `notspam.com`

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -2052,7 +2052,7 @@ create_rule("bad keyword in {}", r"(?is)(?:^|\b|(?w:\b))(?:(?:poker|casino)\W*on
 # Category: Suspicious links
 # Suspicious sites 1
 create_rule("pattern-matching website in {}",
-            r"(?i)(?:{})(?![^>]*<)".format("|".join(pattern_websites)),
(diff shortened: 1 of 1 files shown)


NobodyNada:
I'm pretty sure pattern-matching websites is intentionally non-anchored, because we're looking for specific keywords in a domain name rather than trying to match specific domain names. For instance, we currently catch the domain [xxxfreehdporn DOT com](https://metasmoke.erwaysoftware.com/domains/15807), but AFAICT we would not if the reason was anchored.

makyen:
My understanding is that many of the regexes in `pattern_websites` are explicitly intended to start matching from anywhere within a word, not just at word boundaries. In other words, the lack of having `\b` bookends is intentional. Some of the regexes may benefit, but that would be on a case by case basis. It appears that adding this would break the intended operation of at least some of those regexes.

What is the problem that this change is attempting to address?

makyen:
Also, if bookending is going to be added, it's probably preferable to do it once in the string in which the `.format()` is substituting into, rather than add `\b` to the beginning/end of each sub-pattern.

user12986714:
The specific issue is that almost all [fp](https://metasmoke.erwaysoftware.com/reason/35?filter=fp)s from this reason is due to fragment matches; however I agree that this may reduce detection and may be harmful to the overall goal.

PR #4134, opened 2020-06-30
Add visit my channel to bad keyword with a link

user12986714:
Reference: PR #4133


makyen:
I'm only seeing 6 hits on SE for the new patterns.

PR #4144, opened 2020-07-04
findspam.py: Minor improvements in mostly-image reason

user12986714:
Improved image block finding regex; specify the digits to put into `why` data.


makyen:
!!/approve

PR #4158, opened 2020-07-13
findspam.py: Temporary covid-19 trolling detection

user12986714:
The magic numbers in the code are not supported by any data/reasoning/calculation; they would be adjusted according to the accuracy data in the future.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1547,12 +1547,14 @@ covid_keyword = r"(?-i:covid(?:-?19)?|corona\W*+virus|pendamic|epidemic)"
 covid_location = r"(?-i:china|america|wu\W*+han|he\W*+bei|new\W*+york)"
 covid_syndrome = r"(?-i:hospitalized?|respirat(?:ory|ion)|oxygen\W*+saturation|fever|septics?\W*+shock)"
 covid_likely_troll = r"(?-i:millions?|thousands?|global|(?:organ|lung|respiratory)\W*+failure|septics?\W*+shock|" +\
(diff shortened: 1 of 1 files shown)


teward:
I'm going to suggest that we reject this on the basis of this:

> The magic numbers in the code are **not supported by any data/reasoning/calculation**; they would be adjusted according to the accuracy data in the future.

I'm hesitant to add a new detection specific for COVID-19 things if we don't have accurate statistics on how we should be handling the detection numbers.  My two cents is that without the support/statistics on this it's not suitable for inclusion at this time.

Happy to defer to the majority, but this is my two cents.  (Chances are we're going to get a bunch of false positives because this is a Rapidly Evolving Event)

teward:
This was brought up by @makyen when I poked them in the CHQ channels on chat, but not only is this wondering the bits I identified, but what does this catch that we aren't already catching?  What does this solve?  What kinds of posts does this detect?

user12986714:
Context: https://chat.stackexchange.com/transcript/message/54928872

This PR is mainly inspired by the MS search [here](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=%28%3F%3Acorona%7Ccovid%29&why_is_regex=1&why=manual), which indicates that quite a lot of posts need to be manually reported.

A simple blacklist/watchlist does not work as seen in [this search](https://stackexchange.com/search?q=coronavirus).

The current problem is that we don't have sufficient data to either prove the effectiveness or to apply statistical method for finding out the constants.

A potentially better way to do this once we have enough data is to use some basic ml algorithms, with a demo available [here](https://gist.github.com/user12986714/3fe22ad05dd918ad6999ca33186b5c67).

PR #4165, opened 2020-07-13
Update README.md

Daniil-M-beep:
Correct code error

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ git checkout deploy
 
 python3 -m venv env
 env/bin/pip3 install -r requirements.txt --upgrade
-env/bin/pip3 install --user -r user_requirements.txt --upgrade
+env/bin/pip3 install -r user_requirements.txt --upgrade
 ```
 
 Next, copy the config file and edit as said above.


makyen:
I was under the impression that this was set up this way because the `tld` package *required* the ability to update its TLD list, which is within its file structure. Doing so when installed globally would require SD to be running as root/admin, which we don't want.

Has something changed, so that it updating its file is no longer required?

Daniil-M-beep:
@makyen  I'm not sure but running SD no longer works with the old code, says it's missing dependencies or something

user12986714:
I don't think this is really an error, given that `man 1 pip3` gives
```
       --user Install using the user scheme.
```

Daniil-M-beep:
> I don't think this is really an error, given that `man 1 pip3` gives
> 
> ```
>        --user Install using the user scheme.
> ```



> I don't think this is really an error, given that `man 1 pip3` gives
> 
> ```
>        --user Install using the user scheme.
> ```

Running SD doesn't work with the old code

ArtOfCode-:
> Running SD doesn't work with the old code

[status-norepro]. I installed the packages as specified by the README and can run Smokey normally.

Daniil-M-beep:
Art, you need to do it in a vnev in order for it not to work

makyen:
If it's not working in a particular environment, then the instructions need to explain that, and have an additional, alternate line indicating what's needed. As I understood it, the `--user` there is actually *required* for normal functionality in other environments, but I don't recall specifically testing that.

Daniil-M-beep:
> If it's not working in a particular environment, then the instructions need to explain that, and have an additional, alternate line indicating what's needed. As I understood it, the `--user` there is actually _required_ for normal functionality in other environments, but I don't recall specifically testing that.

@makyen  I did edit the vnev section, didn’t I?

makyen:
@Daniil-M-beep Yes, you did, but it sounds like it may be a situation where it works under some conditions, but not under others. If that's the case, it would be good for us to determine the differences between the environments. For example, is the difference that you and @ArtOfCode- are testing under different operating systems, even though it's a venv?

ArtOfCode-:
My test was WSL 2 Ubuntu. But regardless - the `--user` flag is _required_ - if it's not working, we need to find a way to work around that, but we can't just drop it because that _also_ makes it not work.

Daniil-M-beep:
Closing as I deleted the fork so I am no longer able to commit to the PR

PR #4185, opened 2020-07-16
Draft PR: Avoid merging PR's where the PR string is already watched.

Daniil-M-beep:
Not tested yet. Avoids allowing merging of PRs when the PR string is already watched.


gitmanager.py
```diff
@@ -315,7 +316,15 @@ def merge_pull_request(cls, pr_id, comment=""):
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        string = pr_info["title"]
         ref = pr_info['head']['ref']
+        string = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", str).group(1)
+        string = string.replace("\\", "")
+        is_watchlist = "watch" in pr_info["title"]
+        is_phone = string.isdigit()
```
makyen, line 324: This is not a valid test for it being a number watch. There are plenty of number watches which include characters other than numbers. There are also times when we will watch numbers as keywords.

gitmanager.py
```diff
@@ -315,7 +316,15 @@ def merge_pull_request(cls, pr_id, comment=""):
...
+        reasons = chatcommands.check_blacklist(string, False, is_watchlist, is_phone)
```
makyen, line 325: The goal of this PR is to prevent the tests that are in `blacklist_integrity_check()` in helpers.py from failing once the PR is merged. To accomplish that, tests similar to what are done in `add_to_blacklist()` in gitmanager.py need to be re-run on the added watch/blacklist entry. That is not what `chatcommands.check_blacklist()` does. The checks which `chatcommands.check_blacklist()` performs are not something that needs to be re-run on the entry that is in the PR. They have already been done when the PR was created. The user has already done a `-force`, or not, as is required to pass those checks.

So, there's no need to resolve your circular import, because the thing you're importing that's causing you a problem shouldn't be used or imported.

gitmanager.py
```diff
@@ -306,20 +306,58 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        string = pr_info["title"]
+        file = files_info["filename"]
         ref = pr_info['head']['ref']
-
+        string = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", str).group(1)
```
makyen, line 323: `str` here doesn't appear to be correct. Shouldn't that be `string`?

gitmanager.py
```diff
@@ -306,20 +306,58 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+        string = string.replace("\\", "")
```
makyen, line 324: You shouldn't be removing `\\` characters from regexes. Doing so will result in not matching identical regexes. While what you are matching will eventually be a regex, you're looking for exact text matches.

gitmanager.py
```diff
@@ -306,20 +307,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        string = pr_info["title"]
+        file = files_info[0]["filename"]
+        print(file)
         ref = pr_info['head']['ref']
+        string1 = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", string).group(1)
```
makyen, line 325: This appears to strip off any regex comment. While we may want the general case to be comparison without comments, we currently are not doing that for validation. If we do that, then it should be implemented in `Blacklist.exists()`.

gitmanager.py
```diff
@@ -306,20 +307,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+        username = regex.match(r".*(?=:)", string).group(0)
```
makyen, line 326: The regex here needs to be `r".*?(?=:)"`. Currently, it will find the last `:` in `string`, which may be in the regex text which is being watched.

gitmanager.py
```diff
@@ -306,20 +307,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+                item_to_blacklist = "\t".join([now, username, item])
+            else:
+                op = 'blacklist'
+                item = item_to_blacklist
 
+            exists, line = blacklister.exists(item_to_blacklist)
+            if exists:
+                raise RuntimeError('Already {}ed on line {} of {}'.format(op, line, blacklist_file_name))
+        except Exception:
+            print(sys.format_exc())
```
makyen, line 357: While it's OK to have the above code here for testing, this is effectively duplicating what's in `add_to_blacklist()` (which is what is needed), but it would be better to pull most of this out of both functions into a helper function, so it's only performed in one location, rather than having to maintain nearly identical code in two places.

PR #4188, opened 2020-07-16
Implement Art's `dig` implementation in a much less exploitable way than using `dig` directly.

teward:
Use DNSPython to handle DNS request lookups.  This removes much of the exploit risk because arguments passed into this will not execute in a way that exposes local data.

I am setting this as a **DRAFT** because I am still working on this and making sure we can implement it in a more secure fashion and avoid any type of data leak via argument injection.


tripleee:
Not in a place where I can properly review right now; but once you are closer to done, I'd like to be invited to take a look. The `dig` command jumps a bunch of "do what I mean"  hoops when the immediate result is not a straightforward A record (so a CNAME or a PTR etc) which would be convenient to preserve.

teward:
@tripleee as is we're returning an A record.  This is what Art's `dig` implementation did.  PTR is only on IP addresses, and I'm simply attempting to reproduce what Art did.  It's far from mergeable in its current state but "soon".  CNAME needs some work to be implemented, but PTR is only on IPs so... that's out of scope for what we had initially implemented `dig` for

teward:
So some discoveries.

When you execute an A check on a CNAME it actually returns the CNAME and the A.  Gotta write it to get the A records out of results...

teward:
This should now be ready for review, since I've gotten CNAME resolution working fine as well - looks like it's inbuilt to an A or AAAA lookup.  Dig will return the list of IP addresses when called in chat, but will also say "No data" or similar when no DNS data available (including NXDOMAIN etc.)

makyen:
!!/approve

PR #4252, opened 2020-07-26
Avoid merging already actioned PRs

Xnero-ai:
This should not allow the merging of PRs that have already been actioned.

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -311,7 +311,7 @@ class GitManager:
         if not response or not files_list:
             raise ConnectionError("Cannot connect to GitHub API")
         pr_info = response.json()
-        files_info = files.json()
+        files_info = files_list.json()
         if pr_info["user"]["login"] != "SmokeDetector":
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -311,7 +311,7 @@ class GitManager:
             raise ConnectionError("Cannot connect to GitHub API")
         pr_info = response.json()
         files_info = files.json()
-        if pr_info["user"]["login"] != "Daniil-SD":
+        if pr_info["user"]["login"] != "SmokeDetector":
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -306,8 +306,8 @@ class GitManager:
     @classmethod
     def merge_pull_request(cls, pr_id, comment=""):
         response = requests.get("https://api.github.com/repos/{}/pulls/{}".format(GlobalVars.bot_repo_slug, pr_id))
-        files = requests.get("https://api.github.com/repos/{}/pulls/{}/files".format(GlobalVars.bot_repo_slug, pr_id))
-        if not response or not files:
+        files_list = requests.get("https://api.github.com/repos/{}/pulls/{}/files".format(GlobalVars.bot_repo_slug, pr_id))
+        if not response or not files_list:
             raise ConnectionError("Cannot connect to GitHub API")
         pr_info = response.json()
         files_info = files.json()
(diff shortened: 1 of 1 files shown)


gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
     def merge_pull_request(cls, pr_id, comment=""):
         response = requests.get("https://api.github.com/repos/{}/pulls/{}".format(GlobalVars.bot_repo_slug, pr_id))
-        if not response:
+        files = requests.get("https://api.github.com/repos/{}/pulls/{}/files".format(GlobalVars.bot_repo_slug, pr_id))
+        if not response or not files:
             raise ConnectionError("Cannot connect to GitHub API")
         pr_info = response.json()
-        if pr_info["user"]["login"] != "SmokeDetector":
+        files_info = files.json()
+        if pr_info["user"]["login"] != "Daniil-SD":
```
user12986714, line 314: Are you sure?

Daniil-M-beep: Oops

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
         pr_info = response.json()
+        files_info = files.json()
         if pr_info["user"]["login"] != "SmokeDetector":
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        string = pr_info["title"]
+        file = files_info[0]["filename"]
```
ArtOfCode-, line 321: `string` and `file` are not descriptive variable names. Name variables so their purpose is clear without needing to trace code paths.

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+        print(file)
```
ArtOfCode-, line 322: If this is a useful debugging log, use the `log` function from `helpers` instead. If it's not, remove it.

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
         ref = pr_info['head']['ref']
+        string1 = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", string).group(1)
```
ArtOfCode-, line 324: `string1` is also not a descriptive variable name.

Daniil-M-beep:
@ArtOfCode- This is a draft and is not ready for review

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
         if pr_info["user"]["login"] != "SmokeDetector":
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        pr_string = pr_info["title"]
+        file_modified = files_info[0]["filename"]
         ref = pr_info['head']['ref']
+        requested_pr_string = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", pr_string).group(1)
```
NobodyNada, line 324: I'm not a huge fan of using a regex to parse the PR titles, because we might change the format, or we might have trouble with certain usernames or blacklist items. What about looking at the last entry of the changed blacklist file, or parsing the patch returned by the GitHub API?

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+        else:
+            raise CmdException('GitManager: blacklist is not recognized. Blame a developer.')
+        try:
+            blacklister = Blacklist(blacklist_type)
+            blacklist_file_name = blacklist_type[0]
+            item_to_blacklist = requested_pr_string
+            if blacklist_type in {Blacklist.WATCHED_KEYWORDS, Blacklist.WATCHED_NUMBERS}:
+                op = 'watch'
+                item = item_to_blacklist
+                item_to_blacklist = "\t".join([now, username_of_pr_requester, item])
```
NobodyNada, line 347: What's `now`? Presumably it's the current time, but I don't see it declared anywhere?

NobodyNada (changes requested):
Looks pretty good -- just a couple minor things. I assume this has been tested?

gitmanager.py
Daniil-M-beep: I think RegEx seems fine. Format changes -> the RegEx can change.

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
             raise ValueError("PR #{} is not created by me, so I can't approve it.".format(pr_id))
         if "<!-- METASMOKE-BLACKLIST" not in pr_info["body"]:
             raise ValueError("PR description is malformed. Blame a developer.")
         if pr_info["state"] != "open":
             raise ValueError("PR #{} is not currently open, so I won't merge it.".format(pr_id))
+        pr_string = pr_info["title"]
+        file_modified = files_info[0]["filename"]
         ref = pr_info['head']['ref']
+        requested_pr_string = regex.match(r".*?: \S+ (.*?)(?:\Z|\s*\(\?#)", pr_string).group(1)
+        username_of_pr_requester = regex.match(r".*(?=:)", pr_string).group(0)
```
makyen, line 325: This will get the wrong information if there's a `:` within the regex which is watched ([example](https://regex101.com/r/GkWMfO/1/). If you're going to use a regex here, then `regex.match(r"^.*?(?=: )"` is probably reasonable ([example](https://regex101.com/r/B0Zocz/1)).

Daniil-M-beep: You mentioned a few issues with the RegEx which said you'll work on around 2 months ago, I assume this was one of them...

gitmanager.py
Daniil-M-beep: This cannot be tested due to my test instance not being able to get the latest blacklist information, therefore testing the watched string will always result in it being not caught therefore the testing will be inaccurate. **This will not work.** @makyen  Identified some issues with the RegEx a couple of months ago (see previous comment) but has not gotten to working on it yet/forgot about it. It does seem like the time is missing, when Makyen fixes the RegEx I'll look into it.

teward: `now` needs to be defined per-method, this is defined in other methods for blacklisting, etc. but not here.  Fixing the non-declaration of `now` is trivial, ultimately.  See line 113 in this file (another method's declaration) for how it's done.

Daniil-M-beep: @teward I already responded to this, see above.

teward: > @teward I already responded to this, see above.

I saw your response.  I'm still making my point known.

gitmanager.py
makyen: I have no idea how anything which I've said I'll work on affects this specific issue. If you think it does and will remove the concern I've raised here, please elaborate.

gitmanager.py
makyen: It's unclear to me how anything I've said I'll work on will affect this specific concern with this code. Please explain. As you've said, I must have forgotten about something. Please help me by reminding me.

gitmanager.py
```diff
@@ -306,20 +306,56 @@ def remove_from_blacklist(cls, item, username, blacklist_type="", code_privilege
...
+                item_to_blacklist = "\t".join([now, username_of_pr_requester, item])
+            else:
+                op = 'blacklist'
+                item = item_to_blacklist
 
+            exists, line = blacklister.exists(item_to_blacklist)
+            if exists:
+                raise RuntimeError('Already {}ed on line {} of {}'.format(op, line, blacklist_file_name))
+        except Exception:
+            raise CmdException("An error occured")
```
NobodyNada, line 356: Sorry I missed this last time around — but this is almost definitely not what you want to do. If *any* error occurs within the function (including your new “already actioned” error), this will discard the error and only produce the response “An error occurred”.

Instead, you probably want to:

1. Delete this except block. That way, the error will be handled by chatcommunicate which will log the error and post the “I hit an error [...] run `!!/errorlogs` for details.”
2. Change the `RuntimeError` on line 354 to a `CmdException`. That way the “already actioned” message will be posted to chat directly (instead of going to errorlogs)

gitmanager.py
Daniil-M-beep: @makyen You said that you needed to change the RegEx, I can't remember what exactly as the conversation is lost.

tripleee:
Just a brief note that this will require an overhaul if #3884 gets merged first; then the logic to figure out which blacklist we are talking about is encapsulated into the `blacklists` module.

PR #4253, opened 2020-07-26
Delete force in CHQ

Daniil-M-beep:
Force `delete-force` for all messages in Charcoal HQ.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -2043,6 +2043,12 @@ def delete(msg):
                "`sd delete-force`. See [this note on message deletion]"\
                "(https://charcoal-se.org/smokey/Commands"\
                "#a-note-on-message-deletion) for more details."
+    elif msg.room.id == 11540:
+        return "Messages here are usually kept as records."\
+               "If you really need to delete a report, please use "\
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -2043,6 +2043,12 @@ def delete(msg):
                "`sd delete-force`. See [this note on message deletion]"\
                "(https://charcoal-se.org/smokey/Commands"\
                "#a-note-on-message-deletion) for more details."
+    elif msg.room.id == 11540:
+        return "Messages here are usually kept as records."\
+               "If you really need to delete a message, please use "\
+               "`sd delete-force`. See [this note on message deletion]"\
+               "(https://charcoal-se.org/smokey/Commands"\
+               "#a-note-on-message-deletion) for more details."
```
ArtOfCode-, line 2051: Instead of adding a whole new block here, simply remove the check for `post_data` on line (currently) 2040, so that it becomes simply `if msg.room.id == 11540:`

Daniil-M-beep: No, because the 2 messages are different, unless you want them as one?

PR #4264, opened 2020-07-28
Leave out reason when not provided

Daniil-M-beep:
Previously, when a reason was not provided, the reason in the comment will be left blank. Now it says "Reason not provided."

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -587,9 +587,15 @@ def reject(msg, args, alias_used="reject"):
         message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
         chat_user_profile_link = "https://chat.{}/users/{}".format(
             msg._client.host, msg.owner.id)
+        if comment != '':
         comment = "[Rejected]({}) by [{}]({}) in {}. Reason provided: '{}'\n\n![Rejected with SmokeyReject]({})".format(
             message_url, msg.owner.name, chat_user_profile_link, msg.room.name,
             reason,
+            "https://camo.githubusercontent.com/77d8d14b9016e415d36453f27ccbe06d47ef5ae2/68747470733a2f2f7261737465722e"
+            "736869656c64732e696f2f62616467652f626c61636b6c6973746572732d72656a65637465642d7265642e706e67")
+        else:
+            comment = "[Rejected]({}) by [{}]({}) in {}. No reason provided.\n\n![Rejected with SmokeyReject]({})".format(
+            message_url, msg.owner.name, chat_user_profile_link, msg.room.name,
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -583,17 +583,20 @@ def reject(msg, args, alias_used="reject"):
...
-            message_url, msg.owner.name, chat_user_profile_link, msg.room.name,
-            reason,
-            "https://camo.githubusercontent.com/77d8d14b9016e415d36453f27ccbe06d47ef5ae2/68747470733a2f2f7261737465722e"
-            "736869656c64732e696f2f62616467652f626c61636b6c6973746572732d72656a65637465642d7265642e706e67")
-        message = GitManager.reject_pull_request(pr_id, comment)
-        return message
+        if comment != '':
+            comment = "[Rejected]({}) by [{}]({}) in {}. Reason provided: '{}'\n\n![Rejected with SmokeyReject]({})"
+                .format(message_url, msg.owner.name, chat_user_profile_link, msg.room.name,reason,rejected_image)
+        else:
```
makyen, line 595: This path appears to never actually run `reject_pull_request`. That's most likely an indenting issue in those lines below.

teward: there's other indenting issues in the code too makyen, as well as a "Variable Not Found" type error because comment isn't defined anywhere earlier, so it's a nonexistent variable at the tie of the `if` test.

makyen:
It looks better. There's still quite a bit that's repeated in the different clauses of the code. When maintaining the code, those portions would need to be changed in parallel. In general, it's preferable to avoid, if possible and reasonable, having sections of code that need to be changed in lock-step, because they tend to end up getting out of sync.

I'd prefer to see something like:
```
    rejected_image = "https://camo.githubusercontent.com/" \
                     "77d8d14b9016e415d36453f27ccbe06d47ef5ae2/68747470733a" \
                     "2f2f7261737465722e736869656c64732e696f2f62616467652f626c6" \
                     "1636b6c6973746572732d72656a65637465642d7265642e706e67"
    message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
    chat_user_profile_link = "https://chat.{}/users/{}".format(msg._client.host, msg.owner.id)
    rejected_by_text = "[Rejected]({}) by [{}]({}) in {}.".format(message_url, msg.owner.name,
                                                                  chat_user_profile_link, msg.room.name)
    reject_reason_text = " No rejection reason was provided.\n\n"
    if reason:
        reject_reason_text = " Reason: '{}'".format(reason)
    reject_reason_image_text = "\n\n![Rejected with SmokeyReject]({})".format(rejected_image)
    comment = rejected_by_text + reject_reason_text + reject_reason_image_text
    try:
        message = GitManager.reject_pull_request(pr_id, comment)
        return message
    except Exception as e:
        raise CmdException(str(e))
```

PR #4275, opened 2020-08-01
Update config.sample

Daniil-M-beep:
Updating according to NobodyNada: https://chat.stackexchange.com/transcript/message/55118462#55118462

### DIFF
--- a/config.sample
+++ b/config.sample
@@ -13,6 +13,8 @@ location=location_here
 # metasmoke_key=gibberish_here
 
 # Set GitHub keys
+# git_username=SmokeDetector
+# git_useremail=smokey@erwaysoftware.com
 # github_username=username@domain.com
 # github_password=p@55w0rd
 # github_access_token=some_token
--- a/config.sample
+++ b/config.sample
@@ -13,7 +13,9 @@ location=location_here
 # metasmoke_key=gibberish_here
 
 # Set GitHub keys
-# git_username=SmokeDetector
-# git_useremail=smokey@erwaysoftware.com
 # github_username=username@domain.com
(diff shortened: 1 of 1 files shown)


NobodyNada:
!!/approve

config.sample
```diff
@@ -13,7 +13,9 @@ location=location_here
 # metasmoke_key=gibberish_here
 
 # Set GitHub keys
-# git_username=SmokeDetector
-# git_useremail=smokey@erwaysoftware.com
```
makyen, line 17: These should not be removed. They are actively used on every commit and merge.

NobodyNada: They're not in the `sample-config` in Keybase. Are they optional and defaulting to the values in `git config` or something like that?

NobodyNada: Ah, they [have default values](https://github.com/Charcoal-SE/SmokeDetector/blob/e3e35ba4ac0642f72e42e2a20f77da857d629b4a/globalvars.py#L246). However, it might be worth keeping them in for the sake of documentation.

teward: I disagree with the removal of these keys, leave them there for documentation purposes.  This needs adjusted to *not* remove those lines, please.

makyen: Yes, the config.sample file in the repository is intended as documentation. It shows what information is needed to set up the bot for someone running SD without the SmokeDetector credentials. These values need to be set whenever the bot is run and is making commits to `git`/GitHub using an account other than SmokeDetector. The default values are fine to be used by people running authorized instances.

makyen:
!!/approve

PR #4279, opened 2020-08-01
Less fp's

Daniil-M-beep:
- Excludes StackOverflow, Maths, Mathoverflow and Cross Validated from the "post is mostly images" reason (since StackOverflow can have `` code counted and the other 3 have a lot of MathJax used)
- Adds Cross Validated to the exclusion list for the "mostly punctuation marks in {}" reason (MathJax)

Statistics:

**Excluding StackOverflow, Maths, Mathoverflow and Cross Validated from the "post is mostly images" reason, will result in:**

- 31 fewer fp's
- 0 fewer tp's

The current accuracy of this reason is 17% (17)
New accuracy: 40% (40)

**Excluding Cross Validated from the "mostly punctuation marks in {}" reason, will result in:**

- 0 fewer tp's (all tp's caught by other reasons)
- 30 fewer fp's

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -606,7 +606,7 @@ def len_img_block(string):
 
 
 # max_score=2 to prevent voting fraud
-@create_rule("post is mostly images", title=False, max_rep=201, max_score=2)
+@create_rule("post is mostly images", title=False, max_rep=201, max_score=2, sites=["stackoverflow.com", "math.stackexchange.com", "mathoverflow.net", "stats.stackexchange.com"])
 def mostly_img(s, site):
     if len(s) == 0:
         return False, ""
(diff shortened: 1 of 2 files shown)


Daniil-M-beep:
Note: The failures are not because of my code but because of how the tests are set up

user12986714:
Less fp for the mostly-img reason would be great, but I don't think excluding sites is optimal approach. See https://github.com/Charcoal-SE/SmokeDetector/pull/4190

Daniil-M-beep:
> Less fp for the mostly-img reason would be great, but I don't think excluding sites is optimal approach. See #4190

MathJax would still get caught though

NobodyNada:
> Excludes StackOverflow, Maths, Mathoverflow and Cross Validated from the "post is mostly images" reason (since StackOverflow can have  code counted and the other 3 have a lot of MathJax used)

Isn't there a `stripcodeblocks` option that would help on Stack Overflow? As for the math sites...as far as I can tell, MathJax doesn't render as images, it just embeds the text in a `<span class="math-container">` (which is rendered by client-side JS).

Daniil-M-beep:
> Excludes StackOverflow, Maths, Mathoverflow and Cross Validated from the "post is mostly images" reason (since StackOverflow can have  code counted and the other 3 have a lot of MathJax used)
> Isn't there a stripcodeblocks option that would help on Stack Overflow? As for the math sites...as far as I can tell, MathJax doesn't render as images, it just embeds the text in a <span class="math-container"> (which is rendered by client-side JS).

I have updated the PR and it has nothing to do with SO now, also the issue is MathJax posted as images and not actual MathJax.

PR #4304, opened 2020-08-06
Add reason post is likely nonsense

user12986714:
This PR attempts to find out vandalism and gibberish by calculating the informational entropy of a given text. The constants used in this PR is conservative.


ArtOfCode-:
What's the standard deviation of entropy-per-char? 3.0 seems quite close to 2.6 to me, but it depends - that might be super unlikely or relatively likely; the standard deviation will reveal which it is.

user12986714:
Well, entropy per char for English + space IIRC is ~21...

ArtOfCode-:
That's... not what I'm asking. You have a comment in here that says "Average entropy per char in English is 2.6". If that's the average, what's the stddev?

Daniil-M-beep:
The entropy values here are broken, every post gets caught.

Legit posts:

- “I have seen the discussion about the Turkish Airlines COVID Cabin policy which makes little sense. Regardless though, does anyone know if they are enforcing it? I, like many, will be transferring at a European Airport on two tickets issued separately. I can't check my luggage all the way through from Istanbul to Malaga and can't exit customs to collect the luggage in Brussels (without a forced quarantine or denied entry)” - entropy per char of 0.2332

- “Why all Indian rupee notes are accepted in Nepal and Bhutan, except 500 Rs and 1000 Rs? Why spare those two notes?” - entropy per char of 0.2485

Gibberish posts:

- this this this this spamd dshdshdshdshhds - entropy per char of 0.4045

- “test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test” - entropy per char of 0.4898

- “gibberish dshdshdsaasdlaf,afdasfkkdafkafdkdfkfdskdsfksdkfksd.fk.sdfksfdkfk.fsdk.sdfksfdkfsdkfsdk” + entropy per char of 0.3817

- “dfahdfhdsfsfkdjjsldfksdflfsdlkjfldskjlkfdsjklsfd/jldsfjlsdfjsdfakjsdafjkfsjkldfsaklfdklsalkfdsaklsdadsaklasfdkldfsaljkdfslkaldsfkdslflfsddfskjsdfllsfdaladsflfdsalsdalfksdklafkdsafdsafdsaklkldfsakfdkslaflklfdsakldfsalfdldsaflkasfldjkdfslsdklaflkdsfakjlfsadkljfsakljafsdlkjsdfjjfsdljasdladsfljkfdsjldfsjldsfjsdf“ - entropy per char of 0.4025

I think the entropy values need to be adjusted according to these results

user12986714:
A stat with 12405 fp posts on MS
```
>>> statistics.mean(result)
0.20483261275004847
>>> statistics.median(result)
0.20223865427238322
>>> statistics.stdev(result)
0.031230117152319384
```
So yes, I managed to mess up with the decimal point

findspam.py
```diff
@@ -617,6 +627,26 @@ def mostly_img(s, site):
     return False, ""
 
 
+@create_rule("post is likely nonsense", title=False, sites=["codegolf.stackexchange.com",
+    "stackoverflow.com", "ja.stackoverflow.com", "pt.stackoverflow.com", "es.stackoverflow.com",
+    "islam.stackexchange.com", "japanese.stackexchange.com", "anime.stackexchange.com",
+    "hinduism.stackexchange.com", "judaism.stackexchange.com", "buddhism.stackexchange.com",
+    "chinese.stackexchange.com", "french.stackexchange.com", "spanish.stackexchange.com",
+    "portuguese.stackexchange.com", "codegolf.stackexchange.com", "korean.stackexchange.com",
+    "ukrainian.stackexchange.com"], max_rep=10000, max_score=10000)
```
Daniil-M-beep, line 636: Please add italian.stackexchange.com and make the lint styling tests pass

NobodyNada:
> A stat with 12405 fp posts on MS

That's a lot...unless I'm misunderstanding something, it means we're catching one out of every six non-spam posts.

user12986714:
> > A stat with 12405 fp posts on MS
> 
> That's a lot...unless I'm misunderstanding something, it means we're catching one out of every six non-spam posts.

Well, it is that I took many fp posts out of MS record and analyzed them rather than that this reason will result in those fps.

NobodyNada:
@user12986714 Ah, gotcha. Do you have any stats on how many tps/fps this will result in?

user12986714:
> @user12986714 Ah, gotcha. Do you happen to have any stats on how many tps/fps this will result in over the MS corpus?

W.r.t. result on metasmoke dataset, fp rate is very low. However, since the samples on MS is biased, we cannot really conclude anything.

However, I believe that some test sessions have been run and fp rate is low.

NobodyNada:
> W.r.t. result on metasmoke dataset, fp rate is very low. However, since the samples on MS is biased, we cannot really conclude anything.

Not necessarily in this case — most false-positives on MS are normal English posts, which are what we want to avoid catching.

In that case, I think this is ready for merge (cc @ArtofCode-). If we run into problems we can always revert it.

findspam.py
```diff
@@ -617,6 +627,30 @@ def mostly_img(s, site):
...
+                    "portuguese.stackexchange.com", "korean.stackexchange.com",
+                    "ukrainian.stackexchange.com", "italian.stackexchange.com"],
+             max_rep=10000, max_score=10000)
+def nonsense(s, site):
+    if len(s) == 0:
+        return False, ""
+    if "pytest" in sys.modules:
+        return False, ""
+    probability = [float(s.count(x)) / len(s) for x in s]
+    entropy_per_char = -sum([x * math.log2(x) for x in probability]) / len(s)
```
NobodyNada, line 647: I have had exactly one and a half weeks of formal statistics education in my life, so take what I say with a grain of salt -- but I believe there's a subtle mathematical error here that's leading to a lot of really weird false positives.

This definition of entropy appears to be the same one that Wikipedia uses, which is:



...where X is a [discrete random variable] -- in other words, a variable that can randomly take on elements of a set of discrete events. In this case, X is a discrete random variable representing the possible outcomes of randomly choosing a single character from a post. For example, if 25% of characters in a post are spaces, then X has a 25% chance of being a space.

According to this definition, in order to calculate the entropy of a discrete random variable, first you calculate the "suprisingness" of each possible event, where surprisingness is defined as -log(P(X)) (so that lower-probability events are more surprising). Then, the entropy is the "expected surprisingness" of the random variable, which is the average surprisingness of the events weighted by probability. That's equivalent to the average number of bits of information you learn on each sampling of the random variable -- or the average number of bits of information carried by each character in the input post, which is what we're after.

There are two errors in this implementation:

1. The equation used is for entropy for a single random variable, and our variable is a single character. Therefore, the result **is already entropy-per-character** and dividing by `len(s)` is incorrect. 

2. **`probability` includes duplicate values** whenever the post uses the same character more than once, while the sum in the definition of entropy is supposed to be across *distinct* events of the discrete random variable. By including duplicates, instead of the expected `P(x) log(P(X))`, we're calculating `n * P(x) log(P(x))` (where *n* is the number of occurrences of the character).

Since `P(x) = n / len(s)`, the value we're currently calculating is equal to `-sum(P(x)^2 len(s) log(P(x)) / len(s)`. The `len(s)`'s cancel out, so we end up with <code>-sum(P(x)<strong>**^2**</strong> log(P(x))</code>. Therefore, we can kill two birds with one stone and fix both errors by simply changing the line to: 

    entropy_per_char = -sum([math.log2(x) for x in probability]) / len(s)

With this change, the `entropy_for_char` for most posts ends up being around 4 bits. This value seems a lot more reasonable than 0.25 (a coin flip is 1 bit of entropy, and picking a random character from a post should *definitely* have more entropy than a coin flip). However, this means the `ENTROPY_TOO_HIGH/LOW` constants will have to be tweaked.

NobodyNada (changes requested):
So I ran some tests on this today. It's *really* cool, but it has a couple problems and therefore currently catches a *ton* of FPs. See review comments for details.

findspam.py
user12986714: Looks like I managed to forget that `for x in "string"` does not remove duplicate, and added the `len(s)` part when the result turned out to be erroneous.

NobodyNada:
I ran some more tests today. It's looking a lot better, but we still have problems with:

1) Code. We probably should strip code blocks, but then we'll still have a lot of fp's due to posts with unformatted code.
2) Posts with lots of un-rendered whitespace. IMO we really should collapse repeated whitespace characters.
3) Those constants don't seem to be conservative enough; e.g. https://english.stackexchange.com/a/408724/106362 and https://hermeneutics.stackexchange.com/a/51104 are both caught, with entropies of 4.0233 and 5.6742 respectively.

PR #4315, opened 2020-08-08
Use personal access token

Daniil-M-beep:


makyen:
This change appears to orphan `auth` (i.e. it's no longer used), which should be removed at the same time.

Daniil-M-beep:
> This change appears to orphan `auth` (i.e. it's no longer used), which should be removed at the same time.

I don't see `auth` being used anywhere

makyen:
Yes, exactly. With this change `auth` is no longer used. The line:
```
auth = HTTPBasicAuth(GlobalVars.github_username, GlobalVars.github_password)
```
should be removed at the same time as this change, which is the change that causes it to no longer be used.

Daniil-M-beep:
> Yes, exactly. With this change `auth` is no longer used. The line:
> 
> ```
> auth = HTTPBasicAuth(GlobalVars.github_username, GlobalVars.github_password)
> ```
> 
> should be removed at the same time as this change, which is the change that causes it to no longer be used.

Done

makyen:
!!/approve

PR #4512, opened 2020-08-27
test/test_findspam.py: fail immediately if regex too old

tripleee:
Running with a too old version of the regex library can cause the test to take an excessively long time and even crash the system

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -3,7 +3,6 @@
 
 import sys
 import math
-import regex
 from difflib import SequenceMatcher
 from urllib.parse import urlparse, unquote_plus
 from itertools import chain
@@ -13,6 +12,7 @@ import time
 import os
 import os.path as path
 
+import regex
 # noinspection PyPackageRequirements
 import tld
 # noinspection PyPackageRequirements
@@ -27,6 +27,11 @@ import metasmoke_cache
 from globalvars import GlobalVars
 import blacklists
 
+
+if tuple(int(x) for x in regex.__version__.split('.')) < (2, 5, 82):
+    raise ImportError(
+        'Need regex >= 2020.6.8 (internal version number 2.5.82; got %s)' %
+        regex.__version__)
 
 TLD_CACHE = []
 DNS_CACHE = dict()


teward:
I'm OK with these tests as written, because we DO want to enforce proper versions on modules, and doing so at runtime of the tests is not a bad idea.

Question though, shouldn't we implement this into the system as a whole as well, to version-check regex and make sure we have a recent enough version?

makyen:
Given that this is not just a testing issue, I agree that it's better to put this check in findspam.py, rather than the just the test, as that's the file where we know we have the issue.

Alternately, is there a way for us to enforce at runtime all of the versioning that's in requirements.txt and user_requirements.txt?

If we do enforce all such versioning in only one place, we may _also_ need to enforce it on a per test basis, as not all tests load every file.

teward:
> Given that this is not just a testing issue, I agree that it's better to put this check in findspam.py, rather than the just the test, as that's the file where we know we have the issue.
> 
> Alternately, is there a way for us to enforce at runtime all of the versioning that's in requirements.txt and user_requirements.txt?
> 
> If we do enforce all such versioning in only one place, we may _also_ need to enforce it on a per test basis, as not all tests load every file.

We can add checks on everything in GlobalVars if we want that, or we can import a new function that's called at the beginning of ws.py that tests the versions.

That'd be the 'best approach' but then every time requirements or user_requirements.txt updates we'd have to update the code/tests.  Might get hairy in the long term but I don't think the requirements adjust all that much.

tripleee:
I don't want to over-engineer this, but I agree that it could make sense to move this check to `findspam.py`.  There's a small number of test cases which will pull your machine into a knot if you have the older version of the library, which is what I primarily strive to prevent with this.

The _proper_ solution to enforce this would be to have a `setup.py` with all our dependencies explicitly declared, but I don't think we want to go there at the moment.

makyen:
I understand the specific concern for the test cases. We should be sure we fail early if the `regex` package revision doesn't meet the requirement. OTOH, those test cases exist specifically because the issue with named lists, and other regular expression performance issues, did, in fact, cause SD to crash/lock-up. So, we generally don't want to be running SD without the minimum specified `regex` package, as there _will be_ inputs on which the instance will appear to go offline.<sup>1</sup>

Do we just want to put the specific regex package requirement into findspam.py as the immediate fix and open an issue to remind ourselves that a general fix for testing all dependencies is something we should do?

[Someone's has to have already created such a generic dependency checker, so I'd expect it to be mostly just a drop-in, if we can find an appropriate package or appropriately licensed code.]

-----
1. The user-visible symptoms of the problems which are being tested for are: SD will be working and mysteriously disappear and stop pinging MS. The next instance will be failed-over. The newly active instance will appear to be OK or a brief time, and then it too will go offline, as it tries to test the same post (as part of ensuring that it's not missing posts). This will cascade through most/all of the standby instances until no SD instance is responding. The SD instances will start coming back as they sequentially finish testing the post that caused the problem.  
  
   Code has been added which should significantly mitigate the impact, and the testing was added so we check new regexes against input which was known to cause problems for some previously used regexes (i.e. very basic performance-sanity testing of new regexes).

makyen:
Which filename would be more Pythonic/better:
* requirements.py
* setup.py

My preference would be "requirements.py" and have it *only* deal with verifying actual requirements for running the code. For now, we could just drop the lines which are added here into that file and import it in the tests and wherever else we want to verify the requirements are met. In other words, just handle the current issue in a semi-modular way that we can add to/make more robust some other time.

tripleee:
Let's not reinvent something like [Poetry](https://pypi.org/project/poetry/) here. I'll create a separate ticket for that discussion.

tripleee:
Updated PR. I was afraid I would have to muck with the top-level `except` handlers but it wasn't necessary; `ws.py` now also aborts with an error if your `regex` is too old.

PR #4529, opened 2020-08-28
Updates CI tests for the reject command

Daniil-M-beep:
Updates CI tests for the reject command as the aliases used could have been confusing

### DIFF
--- a/test/test_chatcommands.py
+++ b/test/test_chatcommands.py
@@ -287,7 +287,7 @@ def test_reject(monkeypatch):
         m.setattr("requests.get", lambda *args, **kwargs: None)
         assert chatcommands.reject('8888 "test"', original_msg=msg, alias_used="reject-force") == "Cannot connect to GitHub API"
         m.setattr("requests.get", original_get)
-    assert chatcommands.reject('2518 "test"', original_msg=msg)[:8] in {"Please p", "Cannot c"}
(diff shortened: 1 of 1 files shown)


test/test_chatcommands.py
```diff
@@ -278,16 +278,16 @@ def test_reject(monkeypatch):
 
     # Prevent from attempting to check privileges with Metasmoke
     monkeypatch.setattr(GlobalVars, "code_privileged_users", [])
-    assert chatcommands.reject('8888 "test"', original_msg=msg, alias_used="report").startswith("You need blacklist manager privileges")
+    assert chatcommands.reject('8888 "test"', original_msg=msg).startswith("You need blacklist manager privileges")
```
makyen: You've removed `alias_used="report"`, instead of replacing it with `alias_used="reject"`. All of these should just replace "report" with "reject".

It looks like the `@command` decorator results in the command being actually called as a result of the `__call__` in chatcommunicate.py, which defaults `alias_used` to `None`. This appears to be a larger issue with how all of the commands are organized, which you should just work around here by providing `alias_used` in this line and below.

makyen: And, I'm *way* behind.

Daniil-M-beep: It seems you are reviewing an outdated version.

PR #4553, opened 2020-08-29
[findspam] expand n-word regex

NobodyNada:
A lowercase L looks like an uppercase I: https://metasmoke.erwaysoftware.com/post/261698#preview-tab

This PR amends the offensive regex. 1/0/0 on MS, no legitimate hits on the site


NobodyNada:
As per [Ryan M's suggestion](https://chat.stackexchange.com/transcript/message/55393212#55393212), also catch '1' and '3' in place of 'i' and 'e'.

makyen:
!!/approve

Issue #4291, opened 2020-08-04
SD randomly loses connection to CHQ chat


user12986714:
### What problem has occurred? What issues has it caused?

Examples:  
https://chat.stackexchange.com/transcript/message/55147397
https://chat.stackexchange.com/transcript/message/55127179
https://chat.stackexchange.com/transcript/message/55127158

SD will not post any message to/respond to commands CHQ until a manual reboot in another room.

### What would you like to happen/not happen?
Not to randomly disconnect.

teward:
So, this is a problem, but there's really not that much we can really "do" about it - we're bound by ChatExchange plugin and how it operates, and it has no mechanism to 'reconnect after failure' and such.

user12986714:
This is happening too often now. Maybe we need to spot where the problem is (either here or in CE code base) and fix it.

tripleee:
We are using Art's fork of ChatExchange already, and upstream was making noises which seemed like it would not be receiving further updates from their side. Perhaps the time has come to make ours an official fork, or create a replacement (though the idea scares me).

teward:
> We are using Art's fork of ChatExchange already, and upstream was making noises which seemed like it would not be receiving further updates from their side. Perhaps the time has come to make ours an official fork, or create a replacement (though the idea scares me).

I'm not against having a chat lib that works for us but we need to be careful last time I looked theres a lot of black magic inside ChatExchange to make it work...

makyen:
I'm not sure where the impression was given that ChatExchange wasn't accepting updates. The most recent PR which we (Art) submitted, which was bug fixes, performance improvements, and self-documentation, was merged within 4 hours.

It's reasonable for us to be running off of our own fork, but I'm not aware of ChatExchange being unwilling to accept updates, at least to fix bugs.

I don't know about anyone else, but I haven't even begun to take a look at what might be causing this current issue. It could be in ChatExchange, it could be something we're doing. I think it's quite premature for us to say "let's ditch ChatExchange". Until we're at the point where we *need* something that ChatExchange is set against having, I think we should be targeting any changes/development we do with the intent that it gets merged back into the main ChatExchange repo.

makyen:
[...]

PR #4569, opened 2020-08-31
Tentative async approve/reject

user12986714:
This is an attempt to solve #4291 . The observation of SD behaviour is that it randomly loses connection to CHQ after an approve/reject command is issued. I suspect therefore that this is due to the git operation hanging up the entire command listener thread, as SD is still able to post 'Goodbye, cruel world' message to CHQ but not accept any command. This is an attempt to make approve/reject asynchronous, in order to verify the cause of the bug.

Note: I am not sure what magic is happening for thread raising exceptions. Hence, I am not sure whether the code is exception-safe.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -558,6 +558,12 @@ def approve(msg, pr_id):
         raise CmdException(str(e))
 
 
+@command(int, privileged=True, whole_msg=True)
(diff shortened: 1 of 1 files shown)


user12986714:
No longer needed as the underlying issues have been solved.

PR #4595, opened 2020-09-02
Add reason newly posted youtube video and medium posts

user12986714:
This PR tries to catch newly posted youtube videos.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -646,7 +646,7 @@ def new_video(s, site):
     # Youtube ID regex is by brunodles, https://stackoverflow.com/a/31711517
     youtube_ids = regex.findall(r"(?:https?:\/\/)?(?:www\.)?youtu\.?be(?:\.com)?\/" +
                                 r"?.*(?:watch|embed)?(?:.*v=|v\/|\/)([\w\-_]+)\&?", s)
-    youtube_links = ["https://" + x for x in youtube_ids]
+    youtube_links = ["https://youtu.be/" + x for x in youtube_ids]
     return scrap_and_check(youtube_links,
                            r'"dateText":{"simpleText":"(Jan|Feb|Mar|Apr|May|Jun|' +
                            r'Jul|Aug|Sep|Oct|Nov|Dec)[a-z]? (\d++), (\d++)"}',
--- a/findspam.py
+++ b/findspam.py
@@ -11,6 +11,7 @@ from datetime import datetime
 import time
 import os
 import os.path as path
+import requests
 
 import regex
 # noinspection PyPackageRequirements
@@ -44,6 +45,7 @@ CHARACTER_USE_RATIO = 0.42
 PUNCTUATION_RATIO = 0.42
 REPEATED_CHARACTER_RATIO = 0.20
 IMG_TXT_R_THRES = 0.7
+OLD_VIDEO_THRES = 5
 EXCEPTION_RE = r"^Domain (.*) didn't .*!$"
 RE_COMPILE = regex.compile(EXCEPTION_RE)
 COMMON_MALFORMED_PROTOCOLS = [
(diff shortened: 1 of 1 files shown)


Daniil-M-beep:
Nice idea, have you tested the code?

user12986714:
@Daniil-M-beep I would be happy if someone can help me to test the code. The problem is that I am not capable of creating useful test accounts.

Daniil-M-beep:
@user12986714  Ok. I'll try and organise some testing later today.

Daniil-M-beep:
@user12986714 Apologies but it might not be today but I'll try and get it done relatively soon.

Daniil-M-beep:
This has been tested by me and neither of the 2 rules work.

NobodyNada:
I *really* like the idea, but I'm skeptical of rolling-our-own YouTube, just because it is very much subject-to-change and might end up being troublesome to maintain down-the-road. There may be a couple of more stable alternatives:

- The [YouTube Data API](https://developers.google.com/youtube/v3/getting-started) -- looks like we can make 10k requests/day for free. (Of course Google isn't exactly known for keeping their APIs operational long-term either.)
- [youtube-dl](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#embedding-youtube-dl) is a popular and frequently-updated Python command-line-tool/library for scraping YouTube. I haven't tried it and there doesn't seem to be much of an API documentation, but it looks like you should be able to import the library and call `extract_info(url, download=False)`, which will return a dictionary with an `upload_date` parameter.

user12986714:
> I _really_ like the idea, but I'm skeptical of rolling-our-own YouTube scraper, just because it is very much subject-to-change and might end up being troublesome to maintain down-the-road. There may be a couple of more stable alternatives:
> 
> * The [YouTube Data API](https://developers.google.com/youtube/v3/getting-started) -- looks like we can make 10k requests/day for free. (Of course Google isn't exactly known for keeping their APIs operational long-term either.)
> * [youtube-dl](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#embedding-youtube-dl) is a popular and frequently-updated Python command-line-tool/library for scraping YouTube. I haven't tried it and there doesn't seem to be much of an API documentation, but it looks like you should be able to import the library and call `extract_info(url, download=False)`, which will return a dictionary with an `upload_date` parameter.

I agree it would be a great idea if we can use the API, however it requires coordination w.r.t. API keys and limits the extensibility of such detection mechanisms. For example, it would be great if we later expand to other blog sites like blogspot or something else.

NobodyNada:
> I agree it would be a great idea if we can use the API, however it requires coordination w.r.t. API keys and limits the extensibility of such detection mechanisms. For example, it would be great if we later expand to other blog sites like blogspot or something else.

@user12986714 Those are both true. However, we've had to do API key deployment in the past (e.g. for Perspective), and it's pretty simple:

1. Add a new config entry for the API key, but make sure Smokey still runs fine without it (just without rule disabled). That way, test instances/instances that haven't updated the API key will still work.
2. Update `config.sample` with a placeholder, and update the `config` on Keybase with the real key. That way, all future instances will include the key.
3. Send a message in the runner Keybase chat reminding the runners to add the key to their existing instances.

As far as extensibility goes: yes, we're writing special code to use the YouTube API, but that doesn't stop us from using regexes on Medium or Blogspot. I'm worried about using regex *specifically* on YouTube because YT is not scraper-friendly and I would prefer to stay out of that cat-and-mouse game.

NobodyNada:
I've just taken a closer look at the Medium one, too. I'm concerned by the `class="bh bi at au av aw ax ay az ba fu bd bl bm"`, as obfuscated classes like that are usually anti-adblock measures and are therefore periodically randomized. (Also, I'm just a little bit concerned by the regex; what if someone requests the page from Europe and gets a date of 4 Oct instead of Oct 4?)

However, there's a MUCH easier way to get the date out of a Medium post. Every Medium post has the following meta-tag:

    <meta data-rh="true" property="article:published_time" content="[an ISO 8601 timestamp]">

Finally, since we already have BeautifulSoup, I'd suggest using that [instead of regex to parse HTML](https://stackoverflow.com/a/1732454/3476191), as it will be easier and more reliable.

skprtest666 (changes requested):
F

PR #4596, opened 2020-09-02
Add blacklisters to config file for testing

makyen:
One of the things that's hampered testing is that it's necessary to connect to MS to get a list of blacklist manager approved users. This has been handled in a few cases by copying a codePrivileges.p file around. That's a bit clunky.

This PR adds 3 new optional values to the config file:
* `se_blacklisters`
* `mse_blacklisters`
* `so_blacklisters`
Each value is a comma separated list of IDs which are to be blacklist manager approved on the respective chat server. If there is an existing list of blacklist managers available from MS, then the IDs are added to it. If there is no list of blacklist managers available from MS (i.e. there's no MS key), then the lists that are provided in the config file by chat server are used alone.

### DIFF
--- a/config.sample
+++ b/config.sample
@@ -21,3 +21,8 @@ location=location_here
 
 # Perspective API key
 # perspective_key=some_key
+
+# Manually add blacklist privileged users (primarily for testing):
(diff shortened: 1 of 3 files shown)


Daniil-M-beep (approved):
Good idea.

Issue #565, opened 2017-03-03
!!/report sometimes breaks Smokey


j-f1:
Sometimes, Smokey will silently die when `!!/report` is called. This does not happen every time, and Smokey doesn’t restart or reply to the message. This seems to happen more frequently when `!!/report`ing Ask Ubuntu questions.

AWegnerGitHub:
Thank you for posting this issue.

---

Developers/Troubleshooters/Users of Smokey/Bug Causers:

Due to the nature of this bug and our difficultly in reliably replicating it, we need to keep this issue updated with a few key details when the issue occurs. Hopefully this will help us track down the bug directly or a pattern we can use to replicate the issue:

 - Details on WHEN it failed, with a link to the chat message and a link to the post itself (either in metasmoke or on Stack Exchange).
 - Information on whether we know smokey was alive at the time of the report. When was the last time `!!/report` was run or the last time Smokey posted a message?

teward:
Also include a link to what you reported, so we can see if it's an issue with odd characters or such in the reported item's content that could have broken the chat socket or something.  (As well, any additional information you can provide could be useful as well).

Note that we did some testing on this to try and replicate the chat failure with a debugger attached and couldn't replicate the failures that we've seen.

j-f1:
Since spam posts usually get deleted quickly, you might want to inline them like this:

<details><summary><a href="//charcoal-se.org">Post Link</a></summary>

> Post contents

</details>

Code:
```html
<details><summary><a href="//charcoal-se.org">Post Link</a></summary>

> Post contents

</details>
```

AWegnerGitHub:
Failure:

 - Site: english.stackexchange.com
 - Link: http://english.stackexchange.com/questions/376565
 - Content: Unavailable
 - Chat message: http://chat.stackexchange.com/transcript/message/35810201#35810201
 - Time: Approximately 4:52pm UTC
 - Last Smokey Report: Approximately 4:50pm UTC

Undo1:
Success:

 - Site: stackoverflow.com
 - Link: stackoverflow.com/q/42592564/
 - Content: https://metasmoke.erwaysoftware.com/post/59497
 - Chat message: https://chat.stackexchange.com/transcript/message/35821970#35821970
 - Time: Approximately 5:41am UTC
 - Last Smokey Report: 5:39am UTC

Glorfindel83:
[...]

PR #4690, opened 2020-09-09
[gitmanager] Fix deadlock caused by leak in reject_pull_request

### LINKED ISSUE #4291: SD randomly loses connection to CHQ chat

NobodyNada:
[Once again](https://github.com/Charcoal-SE/SmokeDetector/issues/565), leaked locks cause Smokey to randomly die.

In this case, the `!!/reject` command acquired the gitmanager lock but never released it. This means that when the *next* gitmanager operation attempts to acquire the lock, it deadlocks forever, permanently stalling the calling thread.

Resolves #4291 (or at least one possible cause of the issue).

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -392,16 +392,16 @@ class GitManager:
         if comment:  # yay we have comments now
             GitHubManager.comment_on_thread(pr_id, comment)
 
-        cls.gitmanager_lock.acquire()
-        origin_or_auth = cls.get_origin_or_auth()
-        git.fetch(origin_or_auth, '+refs/pull/{}/head'.format(pr_id))
-        payload = {"state": "closed"}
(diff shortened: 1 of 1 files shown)


NobodyNada:
Also supersedes #4569. However, @user12986714 might have the right idea long-term: we could change command dispatch such that each command gets its own thread (with some sort of `!!/commandstatus` to enumerate running commands, and perhaps a warning if a command takes longer than ~60s to complete). That way, this sort of issue will be both easier to find and easier to work around.

makyen:
!!/approve

PR #4751, opened 2020-09-11
Run commands asynchronously

NobodyNada:
Run each command on its own thread instead of blocking the chat watcher thread. This way, commands that freeze/deadlock won't render Smokey unresponsive. If a command runs for more than 60 seconds without completing, a warning message is printed.

Also adds a `!!/traceback`/`!!/backtrace` command which prints a traceback for a given thread ID; intended to ease debugging of deadlocks.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -23,6 +23,7 @@ import os
 import time
 import collections
 import subprocess
+import traceback
 from html import unescape
 from ast import literal_eval
 # noinspection PyCompatibility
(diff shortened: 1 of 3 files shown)


Daniil-M-beep:
I think the backtrack command should be privileged, plus, have you tested this?

NobodyNada:
@Daniil-M-beep I have tested it.

Making the backtrace command privileged would be easy enough, but why? There's not really a way to cause harm with it as it just retrieves debugging information; and similar commands like `threads` and `errorlogs` aren't privileged.

Daniil-M-beep:
@NobodyNada Judging by the CI test, not sure you have.

NobodyNada:
@Daniil-M-beep @user12986714 is right, I [most definitely tested](https://chat.stackoverflow.com/transcript/167908?m=50439102#50439102) but I forgot to update the unit tests. I'll push those as soon as I'm done banging my head against a wall at Python's mocking system...

NobodyNada:
Phew. Tests are fixed and flake is happy.

user12986714:
What about we just add a kill-thread so we can stop jobs without rebooting the whole thing?

NobodyNada:
@user12986714 In general, it's not a good idea to kill a thread, since it would interrupt whatever code is running without giving it a chance to clean up. It could even cause problems with in the Python interpreter itself -- some Googling says that it's possible to completely deadlock *all* threads by killing a thread which holds the global interpreter lock. There's really no drawbacks to doing a reboot (we even have logic to catch up on missed post IDs), so I believe it's best not to implement a `!!/kill-thread`.

PR #4999, opened 2020-10-04
Change Daniil's username on GitHub

double-beep:
Caused the check to fail, see https://pullapprove.com/Charcoal-SE/SmokeDetector/pull-request/4998 for example.


makyen:
While we hope things get to the point where Daniil returns, this isn't a list on which we keep lines commented out. We should just remove the entry.

PR #5071, opened 2020-10-20
Changed !!/reject 

Victor-VosMottor:
"Please provide an adequate reason for rejection (at least 20 characters long) so the user can learn from their mistakes. Use `-force` to force the reject"
https://chat.stackexchange.com/transcript/message/55884729

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -579,7 +579,7 @@ def reject(msg, args, alias_used="reject"):
     if not code_permissions:
         raise CmdException("You need blacklist manager privileges to reject pull requests")
     if len(reason) < 20 and not force:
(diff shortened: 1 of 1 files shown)


chatcommands.py
```diff
@@ -579,7 +579,7 @@ def reject(msg, args, alias_used="reject"):
     if not code_permissions:
         raise CmdException("You need blacklist manager privileges to reject pull requests")
     if len(reason) < 20 and not force:
-        raise CmdException("Please provide an adequate reason for rejection so the user"
+        raise CmdException("Please provide an adequate reason for rejection (at least 20 characters long) so the user"
```
tripleee, line 582: This line was already longer than I like for them to be, but I guess in this barbaric age less than 120 is okay.

PR #5086, opened 2020-10-23
Split Health-themed spam detection from bad keywords

makyen:
This is sufficiently different from bad keywords such that it should be its own detection.

On MS, this is currently: [2216 Total / 2190 TP (98.83% TP) / 27 FP / 0 NAA](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&why_is_regex=1&why=Health-themed+spam).

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -798,7 +798,7 @@ def has_customer_service(s, site):  # flexible detection of customer service
 
 
 # Bad health-related keywords in titles, health sites are exempt
-@create_rule("bad keyword in {}", body=False, all=False, sites=[
+@create_rule("health-themed spam in {}", body=False, all=False, sites=[
     "stackoverflow.com", "superuser.com", "askubuntu.com", "drupal.stackexchange.com",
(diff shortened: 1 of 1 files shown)


NobodyNada:
Sounds like a good idea to me! However, is there a risk of overlap with other "Bad keyword" reasons that might cause unduly high reason weights?

makyen:
There probably will be some overlap. The detection does look at some similar things which are also detected by a couple/few bad keywords. However, it appears to be intended to detect based on the prevalence of multiple different keywords, rather than a single regex.

As to the FP specifically, there are a total of 27 FP. Of those, there are 3 reports where there is any overlap: [1](https://metasmoke.erwaysoftware.com/post/188924), [2](https://metasmoke.erwaysoftware.com/post/186964), and [3](https://metasmoke.erwaysoftware.com/post/186961). The latter two reports are the same post detected twice. That question was already autoflagged, but was full of content which made it look very close to spam. That particular one would be hard to differentiate.

PR #5097, opened 2020-10-29
Advocate good Git commit messages in `CONTRIBUTING_PULLREQS.md`

iBug:
Since we're already advocating good code comments and PR descriptions, why not roll in commit messages as well? I often admire Makyen for their detailed commit messages for seemingly simple or straightforward changes, and I see it particularly helpful in the long run, so I want to add that as an option for "description of purpose" in our contributing guidelines.

Plus a grammar fix.

### DIFF
--- a/CONTRIBUTING_PULLREQS.md
+++ b/CONTRIBUTING_PULLREQS.md
@@ -10,8 +10,9 @@ watchlist changes) must have all or most of the following information:
 1. What does your pull request change or introduce to the project?
 2. What is the justification for the inclusion of your Pull Request (or, what 
    problem does it solve?)
-3. What does your code (if new) actually do?  Summarize what your new code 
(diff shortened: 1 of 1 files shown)


NobodyNada:
I like it! However, we already talk about comments in the next bullet point, so how about we re-word this one to focus more specifically on Git commit messages?

> 3. Write meaningful commit messages (i.e. do **not** use GitHub's default "Update foo.py" commit message). A commit message must give a brief (but thorough) explanation of what your change does and *why*. See [How to write a Git commit message](https://chris.beams.io/posts/git-commit/). Also include a complete explanation and justification of your proposed change in the body of your pull request.

<sup>Now, if only there was a way to disable those awful default "Update foo.py" commit messages...</sup>

NobodyNada (approved):

makyen:
Thanks. Looks good.

PR #5111, opened 2020-11-12
Fix watch (regression from #3884)

tripleee:
Reinstate proper formatting for watch expressions; required some additional refactoring.

Add a test case for watchlists.


tripleee:
Backed out for now together with #3884 in #5113

PR #5112, opened 2020-11-12
Revert #3884

tripleee:
This rolls back changes emanating from PR #3884 for the time being.

Sorry for the mess /-:


tripleee:
Nah, this won't work.

PR #5113, opened 2020-11-12
Revert #3884 (proper revert)

tripleee:
Roll back the changes for now; still needs more work.


makyen:
I'm a bit late, but looks OK.

PR #5114, opened 2020-11-12
chatcommands.py: trap U+202D invisible character

tripleee:
Display a warning; require -force to override it.

For the record, I have repeatedly submitted commits to remove this character where it had slipped in; most recently, https://github.com/Charcoal-SE/SmokeDetector/commit/7c9aa02d9407dc110a934a44271a586a0c5d92ce

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -359,6 +359,12 @@ def do_blacklist(blacklist_type, msg, force=False):
 
     pattern = get_pattern_from_content_source(msg)
 
+    has_u202d = ""
+    if '\u202d' in pattern:
+        has_u202d = (
+            "The pattern contains an invisible U+202D whitespace character;"
+            " - in most cases, you don't want that")
+
     has_unescaped_dot = ""
     if "number" not in blacklist_type:
         # Test for . without \., but not in comments.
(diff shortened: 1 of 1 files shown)


tripleee:
One more: https://github.com/Charcoal-SE/SmokeDetector/commit/dc49528bcc906108e6b3571ed4c94ea5c637b86b

makyen:
This is helpful, but I'd like us to check for other invisible characters too. If I was doing it, I'd create a regex including all of the invisible characters which we want to check for/not permit without `-force`. I'd then create a new string (e.g. `no_invisible_characters`) having replaced/removed all occurrences of those invisible characters in the current proposed watch/blacklist command. I'd then compare the two strings. If they are equal, then there's nothing there we want to catch. If they are not equal, then there's something there that requires `-force`.

When there are invisible characters, then in addition to the message posted to chat telling the user that they need to use `-force` due to invisible characters, I'd also output to chat `no_invisible_characters` in code format telling them something like "here's a copy of your watch/blacklist with the invisible characters removed". Doing that will probably save people time trying to figure out where the invisible characters are. This later use of `no_invisible_characters` is the reason for going directly for substitution, rather than detection and then also substituting.

As a first stab at what to check for, I'd probably check for everything in the Unicode category "C", which is:

        C   Other                   Cc | Cf | Cs | Co | Cn
        Cc  Control                 a C0 or C1 control code
        Cf  Format                  a format control character
        Cn  Unassigned              a reserved unassigned code point or a noncharacter
        Co  Private_Use             a private-use character
        Cs  Surrogate               a surrogate code point

Because I'm familiar with it, I'd use the list from [my answer here](https://stackoverflow.com/a/26135188/3773011), which is sourced from [here](https://difnet.com.br/opensource/unicode_hack.js).

"Cf" is the group which contains U+202D.

tripleee:
No objection to that plan, but do we actually have other invisible characters which have ever slipped in? This one is a regular problem because it gets copy/pasted from metasmoke etc (though just now I was unable to repro; maybe the metasmoke page was changed? Used to be if you copied the user name it would have this character. Maybe only on some displays).

In the meantime, I'd like to just merge this and think about the wider picture some more. Any objection?

makyen:
Nah, no real objection. For me to personally merge it, I'd want to go over the code again and merge at a time when I was available to test and make sure it didn't cause any problems. If you want to go ahead with that (i.e. merge and test), I don't object and don't have a requirement that I'm around for it. If you don't, I'll probably take a look at it again in ~12 hours, or so.

tripleee:
I'll be in front of my computer in slightly more than 12; feel free to defer back to me if you have better things to do.

tripleee:
Briefly tested: https://chat.stackexchange.com/transcript/message/56165370#56165370

Issue #4841, opened 2020-09-15
Migrate away from Travis


AWegnerGitHub:
Travis-CI.org is shutting down at the end of the year. We have been encouraged to migration to travis-ci.com

---

Hi there!

As you’ll no doubt have noticed, Travis CI has, for some time, been split between travis-ci.com and travis-ci.org. This was initially planned to enable open-source and commercial/private projects to be distinctly separate, but [back in May 2018, we announced](https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps) that open source builders would have the opportunity to build and test their projects under travis-ci.com, meaning your public and private repositories could be managed under one domain with a consistent interface and build environment.

In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely on December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain. 
 

What this means for you

The main thing to expect is that your projects will be built and tested with the same speed and robustness as you’ve come to expect from Travis CI - but also with great new features such as support for AWS Graviton2 which is only available on travis-ci.com. Open-source projects will be free as they always have, with lighting fast builds provided by partners like Packet.

We encourage you to migrate your existing repositories that are currently on travis-ci.org over to travis-ci.com as soon as possible, enabling you to identify any additional changes required to your code or configuration well in advance of the December 31st deadline. We’ve documented everything you need to know in order to migrate your repositories, and of course, you can reach out on the community forum for any additional help.

If you have any questions or concerns, please feel free to contact our support team who will be happy to talk through any migration queries you may have!
 

Happy building on travis-ci.com!
 

The Travis CI Team

NobodyNada:
It's easy to migrate, but is it worth it? We currently test everything with 3 CI providers (Travis, Circle, and GH Actions); is there a reason not to just drop Travis?

NobodyNada:
And...TravisCI [just announced](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) that they’re no longer offering a free plan like they used to. OSS projects now receive 10,000 minutes free (lifetime, not monthly — i.e. they don’t renew every month), and every time we run out we would have to fill out an application for renewal.

I vote we just drop Travis altogether, unless there’s a *really* compelling reason not to.

[...]

Issue #4514, opened 2020-08-28
Improved dependency management


tripleee:
This started as a discussion in #4512 but I'm creating a separate ticket for visibility.

The convention to document dependencies in `requirements.txt` is weak and slightly cumbersome. Ideally, there should be a way to force Smokey instances to verify that they are complying with the dependency requirements of the code base at run time, and ideally fix the situation themselves, or abort if they can't.

tripleee:
I hear [Poetry](https://pypi.org/project/poetry/) gets all the MOT[OS]Ses these days. I haven't looked into whether we can drive dependency installation programmatically but at least it seems to straighten things out a good deal.

makyen:
I was under the impression we had previously rejected the idea of having SD automatically installing dependencies, with the primary reason being that SD is, potentially, a guest sharing a computer and installing dependencies might be disruptive to other things on the machine.

I'm definitely for us verifying that the dependencies which are installed match the requirements specified in requirements.txt and user_requirements.txt. I'm less enthusiastic about automatically installing missing dependencies.

tripleee:
I| you are in a virtual environment, that becomes a non-issue. I know not everyone is excited about requiring virtualenvs, but it's an option if we want it.

teward:
> I was under the impression we had previously rejected the idea of having SD automatically installing dependencies, with the primary reason being that SD is, potentially, a guest sharing a computer and installing dependencies might be disruptive to other things on the machine.
> 
> I'm definitely for us verifying that the dependencies which are installed match the requirements specified in requirements.txt and user_requirements.txt. I'm less enthusiastic about automatically installing missing dependencies.

The way to solve this is to have SD install in userspace - i.e. `pip3 install --user -r requirements.txt -r user_requirements.txt`.

Or alternatively, package up Smoke Detector as a fully fledged application that would be installed in userspace via a setup.py or similar script, which would be able to handle dependencies.

THIS BEING SAID, the initial idea WAS to not have SD handle dependencies automatically because that could cause problems.  Not to mention as currently written, if dependencies update and they would affect NoCrash.py, you then need admin intervention anyways.

I don't think it's a good idea to have SD handle its own dependencies though.  That's why we have dockerfiles for spinning a docker instance or such.

tripleee:
[...]

PR #5117, opened 2020-11-12
Remove dnspython depreciation warning

tripleee:
Where previously we used dns.resolver.query(), switch to d.r.resolve()

Complete compatibility would require d.r.resolve(search=True) but this works without it so far.

Perhaps see also https://github.com/rthalley/dnspython/issues/581

### DIFF
--- a/blacklists.py
+++ b/blacklists.py
@@ -298,12 +298,12 @@ class YAMLParserNS(YAMLParserCIDR):
             if item.get('disable', None):
                 return False
             try:
-                addr = dns.resolver.query(ns, 'a')
+                addr = dns.resolver.resolve(ns, 'a')
                 log('debug', '{0} resolved to {1}'.format(
                     ns, ','.join(x.to_text() for x in addr)))
             except (dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
                 if not item.get('pass', None):
-                    soa = dns.resolver.query(ns, 'soa')
+                    soa = dns.resolver.resolve(ns, 'soa')
                     log('debug', '{0} has no A record; SOA is {1}'.format(
                         ns, ';'.join(s.to_text() for s in soa)))
             except dns.resolver.NoNameservers:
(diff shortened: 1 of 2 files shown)


makyen:
This has dependencies and interrelated issues.

IIRC, the primary issue surrounding this was that the replacement for the deprecated function wasn't available in a dnspython version which is compatible with Python 3.5. So, fixing this issue requires:

* We stop supporting 3.5 (which we can now). This was mostly waiting for Python 3.9.0 to be released, which happened about a month ago.
  * Our officially supported versions and test structure should migrate to testing 3.6, 3.7, 3.8, and 3.9. This should be done in a *separate PR*.
  * We are considering removing CI testing being done by Travis, due to Travis changing how they handle open source projects (basically Travis is changing to make it a pain in the rear to use them). We should go ahead and do this for the SmokeDetector repository. Migrating the userscripts repository is something I've planned. We should also consider migrating the Metasmoke repository away from Travis. We would then still have the ChatExchange repo, which would need to be looked at. However, everything other than SD has a **much** lower CI volume requirement and may be reasonable to support on Travis for a while.  
  Again, this should be a separate PR.
* We need to change the requirements.txt file to indicate a minimum version for dnspython which has the replacement function.
  * Updating requirements and switching to code that requires a new version means that such update **must** be coordinated with all current SmokeDetector runners, so we don't kill active or standby instances.

makyen (changes requested):
See previous comment.

tripleee:
Related: #4514 #4841

teward:
> Complete compatibility would require d.r.resolve(search=True) but this works without it so far.

I'm curious, but why would we need `search=True` to make things work right?  `search` uses **local search domains as set by the system** which for SmokeDetectors should be none.

From the DNSPython configuration documents, this is what 'search' does:

> *search*, a `bool` or `None`, determines whether the search list configured in the system’s resolver configuration are used for relative names, and whether the resolver’s domain may be added to relative names.

We won't need `search=True` configured for our resolutions, because none of them are system-domain-local.

Therefore, that function being added or not is **irrelevant**, I just tested the difference between `.resolve` and `.query` for our code locally and for searches for websites it will not affect the searches.

PR #5125, opened 2020-11-16
Refactor blacklists; add chat commands for ip/ns/asn (#3884 take two)

tripleee:
This contains a major refactoring of `blacklists.py` in order to accommodate the new blacklist types properly and expose a sensible and unified API towards the calling functions in `chatcommands`, `findspam`, etc.

This is take two of #3884 which was merged and then backed out last week. I have fixed two bugs and extended the code slightly.


teward:
NACK on the review - fix your Lint first.

teward (changes requested):
Fix your Lintian problems

tripleee:
I have now tried to rebase this 4-5 times and I end up with really weird results every time. Maybe I'll leave that for tomorrow.

teward:
I would suggest splitting the "New Commands" and "Refactoring' into completely separate PRs.  I say this because I'd rather review the commands themselves independently of the "Refactor blacklists".  So I'm actually going to NACK this PR as is until you can split the two into separate PRs and issues.

teward (changes requested):
See prior comment on the PR

tripleee:
Set to pending: Thomas requested that I split off the blacklist refactoring into a separate PR which can be reviewed separately, so I will proceed to do that, in 6-8.

PR #5136, opened 2020-11-21
GH Actions: Push current ref to deploy branch after test success

iBug:
So we don't rely MS on updating the deploy branch (we still rely it on notifying running Smokey instances).

The "deploy" job only runs after the "build" job succeeds (i.e. the whole build matrix succeeds). Since MS is also relying on CI success (webhook), this should make no difference in terms of software integrity.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,3 +62,12 @@ jobs:
       run: |
         echo -e "## Coverage Report\n" >> "$GITHUB_STEP_SUMMARY"
         coverage report --format=markdown -m >> "$GITHUB_STEP_SUMMARY"
+
+  release:
+    if: ${{ github.ref == 'master' }}
+    needs: build
+    steps:
+    - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


iBug:
@makyen Following up lest this be left in the oblivion:

AFAICT, GitHub Actions *never* had issues with permissions of `GITHUB_TOKEN` from GitHub Actions. The token from Actions running in forks always has read-only access to the source repo (i.e. where it's forked from). Any token has write access only to the repo where it belongs.

So to the best of my knowledge we should be safe to proceed at any time.

iBug:
I just stumbled upon this article. Maybe it's what @makyen was referring to?

[Stealing arbitrary GitHub Actions secrets - Teddy Katz's Blog](https://blog.teddykatz.com/2021/03/17/github-actions-write-access.html)

However, as indicated in the blog post itself, the vulnerability was quickly patched on the public GitHub instance (github\.com) a long time before it's disclosed to the public. So rest assured this is no longer the case.

PR #5145, opened 2020-11-30
Stop posting to room 8089 Mempool

Xekyo:
Hi, 
the Bitcoin folk is not particularly active on the SE chat and we mostly just remove whatever crops up with flags on the site long before we see it in the chat. So, I'd like to stop the smoke detector from posting to the mempool chat (the Bitcoin public chat room, 8089). Hope this PR does that. 

Thank you for your work!


makyen:
!!/approve

PR #5654, opened 2021-01-13
added cigien for SOCVR

rschrieken:


makyen:
The account number that's used in rooms.yml is the user's global account, not the account number on the chat server hosting the room.

I've pushed a different commit.

tripleee:
For the record, that commit was a1e29cc74b8c4378d820e0bcfe4d17c74d72a7ec

PR #5746, opened 2021-01-25
Potential Code Re-Factor and Bug Fixes

HarshCasper:
## Summary

This pull request: 

- Remove unnecessary comprehension
- Change methods not using its bound instance to staticmethods
- Use literal syntax instead of function calls to create data structure

## Importance

I ran a DeepSource Analysis on my fork of this repository. You can see all the issues raised by DeepSource [here](https://deepsource.io/gh/HarshCasper/SmokeDetector/).

DeepSource helps you to automatically find and fix issues in your code during code reviews. This tool looks for anti-patterns, bug risks, performance problems, and raises issues. There are plenty of other issues in relation to Bug Discovery and Anti-Patterns which you would be interested to take a look at.

If you do not want to use DeepSource to continuously analyze this repo, I'll remove the `.deepsource.toml` from this PR and you can merge the rest of the fixes. If you want to setup DeepSource for Continuous Analysis, I can help you set that up.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,8 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
--- a/blacklists.py
+++ b/blacklists.py
@@ -42,7 +42,8 @@ class BlacklistParser:
 
 
 class BasicListParser(BlacklistParser):
-    def _normalize(self, input):
+    @staticmethod
+    def _normalize(input):
         """
         Wrapper to normalize a value. Default method just calls .rstrip()
         """
@@ -202,7 +203,8 @@ class YAMLParserCIDR(BlacklistParser):
         with open(self._filename, 'w', encoding='utf-8') as f:
             yaml.dump(d, f)
 
-    def _validate(self, item):
+    @staticmethod
+    def _validate(item):
         ip_regex = regex.compile(r'''
             (?(DEFINE)(?P<octet>
               0|1[0-9]{0,2}|2(?:[0-4][0-9]?)?|25[0-5]?|2[6-9]|[3-9][0-9]?))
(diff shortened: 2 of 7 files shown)


tripleee:
Thanks for this patch!

The overall diff is surprisingly small; I guess we are in pretty good shape by this metric.

I have a vague objection to the `@staticmethod` changes which I will probably need to articulate better if I really want to argue my case, but other than that, this seems good to go in.

ArtOfCode- (approved):

teward:
This may be approved by Art and others, however I'd like to take your patch and run some runtime tests with TCoM (The Call of Midnight, my dev testing SmokeDetector instance).  the reason for this is I fiddled with making things StaticMethods before and it totally exploded the infrastructure - so forgive me for putting a temporary "Needs Changes" review on this.

teward (changes requested):
See comments - this review may change after testing/review/digging in the dev instance.

teward (approved):

makyen (changes requested):
* What benefits do you see us getting from using the DeepSource service?
* What is your relationship to DeepSource? Why are you recommending it?
* To use DeepSource, it appears we are required to give them write access to our repository. I am not comfortable with that at this time, even though they say they only need write access to a single file.

ArtOfCode-:
@makyen here's the install dialog:

image

IOW, while write access is required, it's enforced to `.deepsource.toml` by GitHub, rather than being "all files but promise we'll only edit one".

FWIW, my review above was purely code-only - agree with your concerns.

teward:
> * What benefits do you see us getting from using the DeepSource service?
> * What is your relationship to DeepSource? Why are you recommending it?
> * To use DeepSource, it appears we are required to give them write access to our repository. I am not comfortable with that at this time, even though they say they only need write access to a single file.

My suggestion: keep the code changes, rip out the deepsource config.  We don't *need* to use the service if we don't want to, for privacy, etc. concerns.

PR #5903, opened 2021-02-18
Change name to Xnero

Xnero-ai:


tripleee:
The Git history will of course still contain the original user name.

Xnero-ai:
> 
> 
> The Git history will of course still contain the original user name.

I know, its just for statistics and consistency.

makyen:
I'm going to close this PR, because the change to watched_keywords.txt changes every line in the file, rather than just the lines with Daniil. This is probably a change to the line endings, but I didn't actually check. I'm not set against the change that appears to be desired here, but we *definitely* don't want that specific commit added to the repository. 

The most likely way for the line endings to have changed is for the file to be edited on a Windows system and then pushed to GitHub. There are a variety of things which might have gone wrong when doing that which could result in a change to the line endings (i.e. there are too many possible issues which might have caused that to be able to say exactly what the cause was without more information/investigation).

PR #5909, opened 2021-02-19
Allow admins to merge all PRs

Xnero-ai:
I have frequently seen admins trying to merge PRs which are code changes. Because these weren't created by SmokeDetector they can't be merged from chat. This PR checks the user ID of the user who sent the `approve` command and if they are a admin (this check is done through matching the user ID against a hard-coded list of all the admins ID's) it allows them to merge all PRs, even those which aren't created by SmokeDetector.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -541,7 +541,7 @@ def approve(msg, pr_id):
     # Forward this, because checks are better placed in gitmanager.py
     try:
         has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050] \
-                               and msg._client.host == "stackexchange.com"
+            and msg._client.host == "stackexchange.com"
         message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
         chat_user_profile_link = "https://chat.{}/users/{}".format(
             msg._client.host, msg.owner.id)
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -540,7 +540,7 @@ def approve(msg, pr_id):
 
     # Forward this, because checks are better placed in gitmanager.py
     try:
-        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050]
+        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050] and msg._client.host == "stackexchange.com"
         message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
         chat_user_profile_link = "https://chat.{}/users/{}".format(
             msg._client.host, msg.owner.id)
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -540,7 +540,8 @@ def approve(msg, pr_id):
 
     # Forward this, because checks are better placed in gitmanager.py
     try:
-        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050] and msg._client.host == "stackexchange.com"
+        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050] \
+                               and msg._client.host == "stackexchange.com"
         message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
         chat_user_profile_link = "https://chat.{}/users/{}".format(
             msg._client.host, msg.owner.id)
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -540,6 +540,7 @@ def approve(msg, pr_id):
 
     # Forward this, because checks are better placed in gitmanager.py
     try:
+        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050]
         message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
         chat_user_profile_link = "https://chat.{}/users/{}".format(
             msg._client.host, msg.owner.id)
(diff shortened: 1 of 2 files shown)


ArtOfCode- (approved):
Approving because looks like it'll work as-is, but some comments:

 * The hardcoded IDs are chat.SE IDs, so this won't work for merging changes from, say, Tavern.
 * Could be further improved by fetching admin IDs from the metasmoke API on startup and caching them; we have IDs for all chat sites both in the MS API and in the config files.

Xnero-ai:
@ArtOfCode- I am not familiar with yaml files so I wasn't sure how to get the ID's from there. What is the path to get admin ID's from the MS API?

ArtOfCode-:
@Xnero-ai `/api/v2.0/users/with_role/:role`, see https://charcoal-se.org/ms/API-Documentation-(V2) for usage details

Xnero-ai:
@Xnero-ai  That doesn’t return the chat ID’s though

ArtOfCode-:
@Xnero-ai use a filter, then it does

Xnero-ai:
@ArtOfCode- Do you think that calling the API is necessary? In my opinion just approving from Chat.SE is enough in order to not add unnecessary code.

ArtOfCode-:
It'd be useful to get it to work everywhere if we can, but it's not essential.

ArtOfCode-:
It'd be useful to get it to work everywhere if we can, but it's not essential.

Xnero-ai:
@ArtOfCode- I think it can be merged as is for now and if needed I can possibly return to it at a later point as RyanM asked me to work on another part of the code.

chatcommands.py
```diff
@@ -540,6 +540,7 @@ def approve(msg, pr_id):
 
     # Forward this, because checks are better placed in gitmanager.py
     try:
+        has_code_permissions = msg.owner.id in [62118, 66258, 73046, 121520, 145827, 156050]
```
makyen, line 543: At a minimum, this needs to check to verify that the command came from chat.SE, as the IDs are not the same on the other chat servers. Without checking which chat server the command came from, it's giving other, different, users authorization to perform this function on the other chat servers.

[Note: I'm going to have more than this single comment, but I'm marking this as requesting changes at this time.]

makyen (changes requested):
This comment is just a placeholder to mark this as requesting changes.

chatcommands.py
Xnero-ai: Done

teward (changes requested):
I cannot support this PR in its current state.

Statically coding who can or can't accept a PR is not a good idea here.  Secondly, I don't want this to be a hardcoded list, I want it to be dynamically updated based on SD Runner or Admin permissions or such on the MS API.

Anyone with code privs should not be allowed to pull in any PR, and I don't really see any benefit for allowing admins to merge any given PR from CHQ.  I'd rather those of us with unrestricted commit/merge access here for SD actually come to GH, review the merge, and approve it or merge it here, unless it's a watch/black list change.

Xnero-ai:
> I cannot support this PR in its current state.
> 
> Statically coding who can or can't accept a PR is not a good idea here. Secondly, I don't want this to be a hardcoded list, I want it to be dynamically updated based on SD Runner or Admin permissions or such on the MS API.
> 
> Anyone with code privs should not be allowed to pull in any PR, and I don't really see any benefit for allowing admins to merge any given PR from CHQ. I'd rather those of us with unrestricted commit/merge access here for SD actually come to GH, review the merge, and approve it or merge it here, unless it's a watch/black list change.

@teward  We don't need the API, it just adds more unnecessary code. Why do you want it? Secondly, you yourself have tried merging PRs not created by SmokeDetector from CHQ. You seem to be conflicting yourself here.

makyen:
While I understand some frustration with people expressing a desire to have the list determined by accessing the API, it's part of limiting how many different places the same permissions list is kept and trying to keep those in sync. The list which holds permissions for the action which this PR is proposing is the Admin role on MS. The list for that is controlled on MS. Repeting that in a hard coded list in SD is definitely non-optimum and is a likely place to be forgotten about when that list needs to change.

Code already exists in SD to get the list of admins from MS. It's used in the `whois admin` command. Requiring that the list of users approve for this command be checked based on the results from MS, similarly to how the list of Blacklist Managers is checked for approving SD's PRs, is not an unreasonable requirement. Yes, it is somewhat more complex, but it's reasonable and not all that difficult, particularly given that it can be largely similar to what exists for the Blacklist Managers list.

makyen:
@Xnero-ai To respond directly to the issue of things like "you've tried to use this feature this way and gotten frustrated by it not working. Why are you not all for this happening?" In general, that someone is frustrated in the moment of doing something doesn't necessarily mean they think that having that capability generally available, even to themselves, is *actually* a good idea. It's quite possible for someone to just be frustrated, even very frustrated, in the moment, but after having some time to think about it, to have come to the conclusion that what happened in that moment really is what should have happened, despite having been frustrated by it at the time.

Bare pointing out that someone's prior positions are in conflict with what they are currently saying tends to be seen as argumentative, particularly by the person you're pointing it out to, which usually isn't a good first, or second, approach to getting agreement by that person to what you're trying to do or the position you're arguing.

If you're going to point that sort of thing out, it's usually better to do so in a more moderate manner. A good approach tends to be to make it about you gaining a clearer understanding of their position, rather than a bare "you're inconsistent". For example, you could try something like:
> OK. I hear what you're saying. However, I'm a bit confused. I created this functionality because I saw you [and others?] be frustrated that it didn't work this way. My intent was to be helpful. What you've just said makes me unsure as to your actual position on if this is useful/desirable. In the past, your frustration appeared to indicate you would want this to work in this manner. From what you're saying now, it seems like you don't want it to work this way. Could you help me by clarifying where you stand on this issue, and, perhaps, explain a bit of your reasoning, so I can better understand the issues as you see them?

makyen:
@Xnero-ai I would suggest pausing development on this for a bit. There are issues which have nothing to do with you or the specific code you've added here which need to be addressed by other people prior to this being appropriate to merge, even if you resolve the admin/API issue. Those issues *could* result in a choice not to have this functionality. Given that, I would prefer that you at least know things are up in the air prior to you choosing to spend more time on this.

Xnero-ai:
@makyen  What are those issues?

makyen:
I understand your curiosity, but I don't have time to go into details at the moment. I understand that might sound a bit trite, but it's going to take considerable time for me to put everything together.

PR #5923, opened 2021-02-22
Remove leading whitespaces from watches

Xnero-ai:


makyen:
Warning about the watch having leading whitespace and requiring `-force` to add it would be good. Just silently removing it limits what someone could add. While it's unlikely to be that someone actually wants a watch to start with whitespace, it's possible and shouldn't be something that's just quietly removed.

PR #5927, opened 2021-02-22
Allow blacklist managers to approve PRs that modify the list files with certain conditions

Xnero-ai:
Allow blacklist managers to approve PRs that modify the list files with certain conditions.

The conditions are:

- The PR only modifies less than 6 lines combined in all files
- The PR has no conflicts and is deemed as "mergeable" by GitHub
- The PR has passed all GitHub Action and CI checks

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -351,15 +351,23 @@ class GitManager:
             list_files = ["watched_keywords.txt", "watched_numbers.txt", "bad_keywords.txt",
                           "blacklisted_numbers.txt", "blacklisted_usernames.txt", "blacklisted_websites.txt"]
             line_changes = 0
-            for file in range (len(pr_files)):
+            for file in range(len(pr_files)):
                 line_changes += int(pr_files[file]["changes"])
                 if pr_files[file]["filename"] not in list_files:
                     raise ValueError("PR #{} modifies files other than the list files, so I can't approve it."
                                      .format(pr_id))
             if line_changes > 5:
-                raise ValueError("PR #{} modifies more than 5 lines, so I can't approve it.".format(pr_id))  
-            if pr_info["mergeable"] != True:
+                raise ValueError("PR #{} modifies more than 5 lines, so I can't approve it.".format(pr_id))
+            if not pr_info["mergeable"]:
                 raise ValueError("PR #{} is not mergeable, so I can't approve it.".format(pr_id))
+            GitHub_checks = requests.get("https://api.github.com/repos/{}/pulls/{}/check-runs".
+                                         format(GlobalVars.bot_repo_slug, pr_id))
+            GitHub_checks = GitHub_checks.json()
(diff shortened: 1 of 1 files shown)


gitmanager.py
```diff
@@ -341,12 +341,43 @@ def merge_pull_request(cls, pr_id, comment=""):
...
+            for file in range(len(pr_files)):
+                line_changes += int(pr_files[file]["changes"])
+                if pr_files[file]["filename"] not in list_files:
+                    raise ValueError("PR #{} modifies files other than the list files, so I can't approve it."
+                                     .format(pr_id))
+            if line_changes > 6:
+                raise ValueError("PR #{} modifies more than 6 lines, so I can't approve it.".format(pr_id))
+            if not pr_info["mergeable"]:
+                raise ValueError("PR #{} is not mergeable, so I can't approve it.".format(pr_id))
+            github_checks = requests.get("https://api.github.com/repos/{}/pulls/{}/check-runs".
```
ThatRyanPerson, line 363: Perhaps I'm missing something, but this doesn't seem to be a working API. See, e.g., https://api.github.com/repos/Charcoal-SE/SmokeDetector/pulls/5922/check-runs

You probably meant 
```
"https://api.github.com/repos/{}/commits/{}/check-runs"
                                    .format(GlobalVars.bot_repo_slug, pr_sha)
```
(example: https://api.github.com/repos/Charcoal-SE/SmokeDetector/commits/c99d4783e6089651cf3f32099cbb14a1754d2a67/check-runs) 

Also, if you tested this and it is currently getting past this code and allowing merging, it needs more error checking, since an API failure should not allow PRs to be merged.

gitmanager.py
```diff
@@ -341,12 +341,43 @@ def merge_pull_request(cls, pr_id, comment=""):
...
+            for check in range(int(github_check_count)):
+                if github_checks[check]["conclusion"] != "success":
+                    raise ValueError("PR #{} has failed one or more GitHub Actions, so I can't approve it."
+                                     .format(pr_id))
+            ci_check = requests.get("https://api.github.com/repos/{}/commits/{}/status"
+                                    .format(GlobalVars.bot_repo_slug, pr_sha))
+            ci_check = ci_check.json()
+            ci_check = ci_check["statuses"]
+            for check in range(int(len(ci_check))):
+                if ci_check[check]["context"] == "ci/circleci: build":
```
ThatRyanPerson, line 377: Only looking for a specific check feels potentially fragile, especially if checks are changed in the future (e.g., more added).  I think you're trying to exclude the PullApprove check here (which is correct, because the user approving it must be code-privileged, so they could approve anyway).  To do that, I'd personally just check if the status is _not_ the PullApprove check to exclude only that one, while still checking for potential future status providers.

ThatRyanPerson (changes requested):
I'm looking forward to being able to use this!  A couple comments.

PR #6540, opened 2021-06-16
globalvars.py: Use ConfigParser instead of RawConfigParser

teward:
According to the documentation for `configparser.RawConfigParser` (https://docs.python.org/3/library/configparser.html#configparser.RawConfigParser), this is a **legacy** API call that allows for non-string section names, option names, and values via the unsafe `add_section` and `set` methods.  We don't use those methods, but it's overall still a legacy compatibility and should be avoided.

The documentation suggests that instead of using `RawConfigParser`, we should use `ConfigParser` and set `interpolation=None` at instantiation instead.

### DIFF
--- a/globalvars.py
+++ b/globalvars.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from html.parser import HTMLParser
 from html import unescape
 from hashlib import md5
-from configparser import NoOptionError, RawConfigParser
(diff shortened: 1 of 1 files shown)


makyen:
Given that this is merged, everyone's still welcome to review it, but it doesn't need to be a task for them.

PR #6541, opened 2021-06-16
Implement DNS Customization

teward:
In SmokeDetector, we use DNS for a variety of spam checks.  This is crucial for our ASN, nameserver, and IP checks on domains.

However, the DNS functionality has been restricted by the ability to only rely on the host system's DNS configurations - that is, if there's a broken of funky resolver in the host system, it will not function properly for SmokeDetector.  In many cases, there are not solutions to fix those issues, and it may be better for SmokeDetector to be configured to use *different* configuration options.  We also don't currently implement thread-safe DNS caching, and instead 'do our own caching' which is not thread safe and does not have a true TTL for data values in the DNS results.

This PR addresses the following:

1. Allow for configuration of DNS Nameservers (or "system" to use system configurations) in config file (via `dns_resolver`) using either the string "system" or a string with comma separated IP values for nameservers (`8.8.8.8,8.8.4.4` for Google DNS for example)
2. Enable or disable DNS lookup caching (via `dns_cache_enabled` in config), and utilize dnspython's inbuilt `dns.resolver.Cache` caching system to have a cache of data
3. Allow configuration of the Cache 'time-to-live' for lookup results (via `dns_cache_cleanup_interval` in config).  This defaults to a 5 minute time to live, but also behaves much like a DNS server's cached lookups would, always returning a cached value until the specified time to live value of the record (from the authoritative server) or the time to live value of a cached result in the DNS server (`dns.resolver.Cache(GlobalVars.dns_cache_interval))`) is reached, at which point the cached record is flushed and a subsequent query for the domain will return a fresh lookup/answer from DNS.

This results in several new configuration variable options being loaded up and optional for inclusion - if the variables are not defined in the config for SmokeDetector, there are fallback values defined for defaults (use System resolver configuration for the resolver, enable caching, use default 5 minute TTL for cached results in the Cache)

This does not take into account the ConfigParser changes in #6540 and will need rebased on that if #6540 is accepted before this.


ws.py
```diff
@@ -91,6 +92,13 @@
 if not GlobalVars.metasmoke_ws_host:
     log('info', "No metasmoke websocket host found, which is okay if you're anti-websocket")
 
+# Initiate DNS
+if GlobalVars.config.dns_nameservers != 'system':
```
makyen, line 96: In this line and the next one, `GlobalVars.config.dns_nameservers` should be `GlobalVars.dns_nameservers`.

With it as it is currently, in a brief test with an SD instance using a config file unmodified from earlier testing (i.e. with none of the new config entries), this line results in a:
```
AttributeError: 'SectionProxy' object has no attribute 'dns_nameservers'
2021-06-19 01:51:41.260326 UTC
  File "ws.py", line 96, in <module>
    if GlobalVars.config.dns_nameservers != 'system':
```

findspam.py
```diff
@@ -18,7 +18,7 @@
 # noinspection PyPackageRequirements
 from tld.utils import TldDomainNotFound
 import phonenumbers
-import dns.resolver
+import dns.exception
```
makyen, line 21: While this appears to work (it actually results in all of `dns` being pulled in, rather than *just* `dns.exception`), it seems like it would be better for this to be stated as `dns.resolver`, as that's what we're actually using.

globalvars.py
```diff
@@ -12,11 +12,13 @@
 # noinspection PyCompatibility
 import regex
 import subprocess as sp
+import json
```
makyen, line 15: It's unclear to me why the json package is being imported. There don't appear to be any lines added or changed which use the json package.

chatcommands.py
```diff
@@ -1,6 +1,7 @@
 # coding=utf-8
 # noinspection PyUnresolvedReferences
-from chatcommunicate import add_room, block_room, CmdException, command, get_report_data, is_privileged, message, \
+from chatcommunicate import add_room, block_room, CmdException, command, get_report_data, \
+    is_privileged, message, \
```
makyen, line 4: Based on this change and another change below, I assume your IDE doesn't like longer lines. I mention this merely because this being changed resulted in me spending time trying to determine if there was a functional change in the lines (i.e. visually comparing each portion of the line, not finding one and then going back and checking it again). That's not a problem, I'm just wanting to be sure I'm not missing a functional change.

ws.py
```diff
@@ -91,6 +92,13 @@
 if not GlobalVars.metasmoke_ws_host:
     log('info', "No metasmoke websocket host found, which is okay if you're anti-websocket")
 
+# Initiate DNS
```
makyen, line 95: Having this at this location in the file will probably be fine. However, we should make sure that we're not relying on the configuration choices having been applied at the time the other packages which use the `dns,resolver` are imported (i.e. it won't have the configuration applied at the time those packages initialize).

That's not a problem. It's just something which we should be aware of to make sure it really isn't a problem.

makyen (dismissed):

classes/_Post.py
```diff
@@ -1,6 +1,5 @@
 # coding=utf-8
 import json
```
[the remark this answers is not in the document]
teward: Circular imports were exploding initially.  We might be able to revert some of these Post object changes now.

ws.py
teward: Thats because I made an oopsies, it'd just be GlobalVars.dns_nameservers

classes/_Post.py
```diff
@@ -78,9 +77,8 @@ def _parse_json_post(self, json_data):
 
         try:
             data = json.loads(text_data)
-        except ValueError:
-            log('error', u"Encountered ValueError parsing the following:\n{0}".format(json_data))
-            return
+        except ValueError as err:
+            raise ValueError("ValueError processing data: {}\n{}".format(str(err), json_data))
```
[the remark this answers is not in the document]
teward: Again  this was to fix a circular import problem and can likely be reverted.

globalvars.py
teward: Leftovers from trying to parse strings then I saw the .getboolean and .getfloat options in configparser.  will remove

chatcommands.py
teward: Correct, IDE enforces PEP8 + 120char maxlimit per line

ws.py
teward: Currently only chatcommands and findspam rely on dns.resolver to function.  And chatcommands' `dig` actually creates a brand new Resolver and not default resolver to return IP data.  Done this way to make sure that we get 'good' DNS and not crap DNS in the event we are digging to debug odd IP values.

PR #6550, opened 2021-06-19
Create DNS comparison command for privileged users.

teward:
Back when we had a discrepancy in DNS results, we were afraid SmokeDetector might be a part of it.  Which is what let ArtOfCode to create the `dig` command, so we could look at DNS results returned by SmokeDetector.

With the recent changes done to SmokeDetector with #6541, we now have the capacity to customize SmokeDetector's DNS settings and configurations.  Which means that having a customizable DNS means we need a mechanism to compare DNS output from SmokeDetector's DNS settings and a reliable DNS server.  In this case, I chose Google DNS as a comparative server for our use to compare against whatever DNS setup is done on SmokeDetector instances, whether system configured DNS or customized DNS with individual SD instances.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -34,6 +34,7 @@ from classes import Post
 from classes.feedback import *
 from classes.dns import dns_resolve
 from tasks import Tasks
+import dns.resolver
 
 
 # TODO: Do we need uid == -2 check?  Turn into "is_user_valid" check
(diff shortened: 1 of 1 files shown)


makyen:
A [quick check indicates there's a problem](https://chat.stackexchange.com/transcript/message/58390449#58390449):
```
### AttributeError on Makyen/EC2-test at 2021-06-20T00:42:32Z: module 'dns.resolver' has no attribute 'Resolve'
File "/home/ec2-user/Charcoal-SE-SmokeDetector/chatcommunicate.py", line 458, in __call__
result = self.__func__(*processed_args)
File "/home/ec2-user/Charcoal-SE-SmokeDetector/chatcommands.py", line 2347, in dnscheck
google = dns.resolver.Resolve(configure=False)
```

teward:
@makyen typo fixed

makyen:
[Brief testing](https://chat.stackexchange.com/transcript/message/58390549#58390549) indicates this is basically working.

PR #6703, opened 2021-09-28
findspam.py: tweak up DNS timeout to 60s when running pytest

tripleee:
Attempt to work around some DNS issues in CI

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -1223,7 +1223,11 @@ def dns_query(label, qtype):
     # If there's no cache then assume *now* is important
     try:
         starttime = datetime.utcnow()
-        answer = dns.resolver.resolve(label, qtype, search=True)
+        # Extenmd lifetime if we are running a test
+        extra_params = dict()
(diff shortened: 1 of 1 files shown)


makyen:
I'm sorry for the delay in responding to this. Overall, 60s seems a bit excessive wrt. length, but probably won't bite us in the rear. There are corner cases which could result in the CI testing taking an excessive amount of time, if there are a very large number of timeouts. But, that happening is probably rare, so, likely isn't worth the time/effort until it happens.

###### Potential future improvement:
I haven't double-checked the testing code. Are we testing all of the domains sequentially or in parallel? If we're doing them sequentially, it would probably help us to do them in parallel, or at least some number greater than one at a time.

tripleee:
Thanks for the feedback. Yes, doing this `async` would be a significant improvement. I noticed the library has acquired some support for this since last time I looked.  I'll create a separate ticket about that.

It would probably be good to make Smokey more `async` too but that might be more challenging.

PR #6704, opened 2021-09-29
test/test_findspam.py: reformat "regex poison" posts for legibility

tripleee:
Distracting and potentially troubling (inefficient syntax highlighting).

My Emacs was freezing briefly when I opened this file ... Blame Emacs, but still, inconvenient


makyen:
I'm surprised Emacs has a problem with such short lines. While, yeah, the lines aren't short, they are dramatically shorter than any point at which I'd expect to see a problem. Clearly, vi wins the editor war if Emacs fumbles on something like this. :) Joking aside, it's something I really would expect Emacs, or any semi-modern editor, to be able to handle. While Emacs is venerable, it is routinely updated, and, frankly, I would have expected it to be able to handle this without batting an eye, even 25 or 30+ years ago. Part of my surprise is that handling lines that are substantially longer than these is something I routinely expect my editor to do when dealing with obfuscated or minimized code.

The strings appear to be identical, so it shouldn't change anything functionally.

For me, the change reduces the visceral feel and immediate high-level understanding of what's being tested. With the change I have to actually *think* about what the string is, rather than just know the high-level approximation of the string content. OTOH, while reading the code, once thinking about the new representation (effectively performing the expansion mentally), the new representation gives a substantially faster grasp of the low-level exact contents of the string, from which a high-level understanding can be derived.

makyen (approved):

tripleee:
Thanks for the review. I too was surprised that the syntax coloring for Python is not more robust, but I'm guessing they have some quick and dirty regexes in there.

PR #6705, opened 2021-09-29
 Use plain text for queue timings data, part 2

NobodyNada:
https://github.com/Charcoal-SE/SmokeDetector/pull/3971#issuecomment-930616269


makyen:
I ended up actually including the commits via cherry-picking. I'll be opening yet a third PR, which includes all the commits along with a bug fix in queue_timings.py

PR #6711, opened 2021-10-15
Files Reformatted for Better Visibility and Understanding.

dhhruv:
Files Reformatted for Better Visibility and Understanding.


makyen:
For you, as someone who is not already actively maintaining the code, globally reformatting the code in a repository is inappropriate.

Even if you are one of the maintainers, when the project is a group project, such reformatting would be something which is normally discussed prior to implementation. If you own the repository, then that's something which you can implement on your own, but when working with other people, it's something which is normally discussed and a consensus reached. In such cases, someone might submit a PR to give people an example of what it would look like, but not without approaching it from the position that it's something for others to discuss.

dhhruv:
Yeah, I'll take care of that. Thank you for reaching me. I'll discuss with
other team members first and then add any new pull request. At first, I
just reformatted the code and now I'm working on it and adding more commits
as my work is done. I'll discuss this with you later.

PR #6736, opened 2021-11-02
worldemart.com no longer resolves (NXDOMAIN)

Spevacus:
Causes a build test failure.

### DIFF
--- a/watched_nses.yml
+++ b/watched_nses.yml
@@ -1527,7 +1527,9 @@ items:
 - ns: intersmartsolution.com.
 - ns: blackhostbd.com.
 - ns: arrowbigapps.com.
-- ns: worldemart.com.
+- comment: 2021-11-02 NXDOMAIN
+  disable: true
+  ns: worldemart.com.
 - ns: chinahoster.com.
 - ns: niagahoster.com.
 - comment: no longer responding 2020-08-31


makyen:
I'm not seeing which build failure you're trying to resolve. Please provide a link to the failure. A `!!/dig worldemart.com` command resulted in 45.114.245.100.

Spevacus:
Lame. It appeared to be a transient issue. For a fair few builds leading up to the [one immediately before I submitted this PR](https://github.com/Charcoal-SE/SmokeDetector/runs/4084517567?check_suite_focus=true), that NS was not resolving (and indeed it didn't resolve on my browser). It appears to be back up now, though.

As this disable is no longer needed, I'm just gonna close this.

makyen:
I'm not seeing a problem for worldemart.com in the check suite you linked. It's quite possible that I'm just missing it, as the DNS failures have a *lot* of extraneous lines in the output. I've recently made a change to SD's CI testing which should prevent about 1k of the lines from printing, so it should be a bit easier to see what's happening. At least it is for me. 

I do see the following two lines which indicate issues with main-hosting.eu.:
```
dns.resolver.NoAnswer: The DNS response does not contain an answer to the question: main-hosting.eu. IN A

dns.resolver.NoNameservers: All nameservers failed to answer the query main-hosting.eu. IN SOA: Server 127.0.0.53 UDP port 53 answered SERVFAIL
```
A similar issue has been in many of the failures. I've commented out main-hosting.eu, but the issues might need further investigation. The strange thing was that they normally failed on Python 3.9, but not Python 3.6. I'm not aware of any change between the two Python versions which would cause that, unless it's an order of testing issue.

Spevacus:
That's odd. In that same link, at [step 1355](https://github.com/Charcoal-SE/SmokeDetector/runs/4084517567?check_suite_focus=true#step:10:1355), it indicates that the DNS lookup timed out.

Screenshot so you don't think I'm crazy:


Another, similar, failure happened a short while before that [here](https://github.com/Charcoal-SE/SmokeDetector/runs/4083308542?check_suite_focus=true).

makyen:
I don't believe that timeout doesn't actually causes an error. The "Captured stderr call" output there is from the `log()` calls in [`_validate`](https://github.com/Charcoal-SE/SmokeDetector/blob/master/blacklists.py#L293) for `class YAMLParserNS(YAMLParserCIDR)`. While some of them are at the `"warn"` level, they don't appear to cause an actual error in the CI testing. For example, "idhostinger.com" also had a similar DNS lookup timeout in the testing you linked (line 1216). Having such a large amount of output there, and having several of them be `"warn"` level makes it unclear where the errors are. The actual errors are reported in the text which is above that "Captured stderr call" output section. When initially looking at errors like this one it took me quite a while to work through what was actually being shown in these, as those extra 1000+ lines of output really confuse things, IMO.

PR #6737, opened 2021-11-03
Adds a TLD check for blogspot domains.

Xnero318:
Adds a TLD check for blogspot domains. Tested here: The pattern is a blogspot watch but keeps the TLD, in most cases this should be removed.


makyen:
!!/approve

PR #6738, opened 2021-11-04
For notifications, don't fetch the users in the room unless needed.

makyen:
The prior code was making a request to chat for the current users in the room  for every report which was posted in chat, regardless of if that information was needed (i.e. that there were users who had requested notifications to be sent with reports to that room only when the user was in the room).

This PR changes the logic to only fetch from chat the list of users currently in the room when that data is actually needed (i.e. there's at least one user who has requested to be notified for reports in that room only when they are present in the room).

Ideally, we'd cache this information after the first report into a room and then only watch for notifications of users joining or leaving the room. However, it's rare for people to request pings only when they are in the room (it's a, currently, undocumented feature). If we have people wanting this capability in our high-traffic rooms (e.g. Charcoal HQ, The Fire Department, etc.), then we should strongly consider at least creating a cache for this information, rather than fetching it for each report that's posted.

### DIFF
--- a/datahandling.py
+++ b/datahandling.py
@@ -527,22 +527,36 @@ def get_user_ids_on_notification_list(chat_site, room_id, se_site):
 
 def get_user_names_on_notification_list(chat_site, room_id, se_site, client):
     names = []
-    try:
-        current_users = client._br.get_current_users_in_room(room_id)
-    except Exception:
-        # ChatExchange had a problem getting the current users. This shouldn't be allowed to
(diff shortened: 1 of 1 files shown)


tripleee:
I can't say I completely grasp the context of this change, but if you are looking for a PullApprove checkmark, I'll be happy to add one. In other words, looks good to me.

makyen:
I'd wanted to make one more pass at at least minimal testing, which I've now done.

PR #6751, opened 2021-11-22
findspam.py: obfuscated_word(): new rule "obfuscated word"

tripleee:
Simple 1337 decoder for a small set of troll / phone support scam keywords

test/test_findspam.py: test case for the above

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse, unquote_plus
 from itertools import chain
 from collections import Counter
 from datetime import datetime
+from string import punctuation
 import time
 import os
 import os.path as path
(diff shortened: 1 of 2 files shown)


tripleee:
I ran an FP scan on some 20,000 old posts and it did not trigger on any of those.

tripleee:
Duh, appease `flake8`

tripleee:
I commented out the phrases with multiple words in them because this first cut only examines individual words. I also commented out AVG because it's so short as to be excessively FP-prone, and I don't have any samples of it being obfuscated anyway. This could be extended to examine multiple words and variations in what counts as a word delimiter but that will increase the overhead. We could also pull in the Unicode homographs library if we wanted to.

tripleee:
I ran this on my laptop for almost 10 hours and it didn't report any false positives. (Granted, it didn't report much else either. I'm guessing the posts were removed by the main Smokey so quickly that they did not make it to my instance, but there could be something wrong with the rule logic too.)

tripleee:
Rebase to resolve conflicts with master

makyen:
I'm sorry I didn't get to reviewing this rapidly. At the time, I was quite busy with things which I'd already committed to accomplish. Then, once you merged it, I assigned looking at it a much lower priority.

findspam.py
```diff
@@ -2413,6 +2414,75 @@ def religion_troll(s, site):
...
+    obfuscation_keywords = [
+        # Trolls && cussing
+        "c" + "ock",
+        "c" + "unt",
+        "d" + "ick",
+        "f" + "uck",
+        "mother" + "f" + "ucker",
+        "p" + "enis",
+        "p" + "ussy",
+        "w" + "hore",
```
makyen, line 2430: Why are you breaking these words up? Functionally, it doesn't affect it, but it's unclear why the actual words shouldn't be here. Yes, they are offensive words, but that's what we're working with.

findspam.py
```diff
@@ -2413,6 +2414,75 @@ def religion_troll(s, site):
...
+    }
+    for p in punctuation:
+        if p not in t:
+            t[p] = ''
+    trans = "".maketrans(t)
+    for word in regex.split(r'[-\s_]+', s):
+        # prevent FP on stuff like 'I have this "number": 1111'
+        word = word.strip(punctuation)
+        translated = word.translate(trans).lower()
+        if translated in obfuscation_keywords and translated != word.lower():
```
makyen, line 2481: It would be somewhat better to reverse the order of the checks in this `if` statement. the `translated != word.lower()` is going to be faster than the `translated in obfuscation_keywords` in most cases, and would short-circuit needing to check every entry in the `obfuscation_keywords` list.

tripleee:
@Makyen Thanks for the review. I have a more drastic refactoring in progress which addresses several of these concerns. I hope to be able to create a new PR next week.

PR #6752, opened 2021-11-22
findspam.py: "several emoji in title" rule

tripleee:
test/test_findspam.py: test for the above

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -2423,6 +2423,10 @@ create_rule("numbers-only title",
             r"^(?=.*[0-9])[^\pL]*$",
             sites=["math.stackexchange.com"],
             body=False, max_rep=52, max_score=5)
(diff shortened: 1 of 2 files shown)


tripleee:
I ran this on my laptop for almost 10 hours and it didn't report any false positives. (Granted, it didn't report much else either. I'm guessing the posts were removed by the main Smokey so quickly that they did not make it to my instance, but there could be something wrong with the rule logic too.)

tripleee:
Rebased to simplify merge.

RobCoding:
> ... there could be something wrong with the rule logic too.)

This seems too simple to reliably detect all emojis, compared to the libraries available for the purpose: https://github.com/NeelShah18/emot or see: https://stackoverflow.com/a/43146653/3648282

We appreciate the effort, it's a good idea to have this rule.

PR #6767, opened 2021-11-25
findspam.py: update "obfuscated word" to include "in X"

tripleee:
... where X is "title", "body", or "username", as for other similar rules


tripleee:
Sucks that I failed to put this in in the original PR, but I guess we will want consistency. Having an abandoned reason show up in the statistics is a small price to pay for that, I hope.

PR #6772, opened 2021-11-28
Get all number strings for numbers tests

makyen:
The current numbers tests don't get all the possible combinations of numbers out of the text being tested. This results in not detecting numbers which are placed right next to each other. You can test this with:
```
!!/test-q site=math.stackexchange.com <p>Some1-888-884-0111 888-884-0111 +1-972-534-5446 972-534-5446 1-628-215-2166 628-215-2166 1-844-802-7535 844-802-7535 body</p>
```
which will, currently, produce the following output:
> Would not be caught as a question on site math.stackexchange.com.
[Note: This example uses math.stackexchange.com, because there's another detection which runs everywhere except on mat.SE which detects any phone number, which would just make this example more complex.]

It should have the following result:
> Bad phone number in body, potentially bad keyword in body
> ----------
> Bad phone number in body - 6282152166 found normalized; 8888840111 found normalized; 16282152166 found normalized; 9725345446 found normalized; 19725345446 found normalized; 18448027535 found normalized; 844-802-7535 found verbatim; 18888840111 found normalized
> Potentially bad keyword in body - 6282152166 found normalized

This PR:
* Changes how the list of potential numbers are generated such that all the combinations of numbers (i.e. all combinations containing sequences of numbers which are split at non-number characters). 
* Reduces the minimum number of digits to be tested from 9 to 7.
* Prevents users from adding patterns to the numbers lists which are impossible to detect, and provides some feedback as to what's needed.
* Informs the user that the pattern they are trying to add can not be detected verbatim when that's not possible. Users will need to append `-force` to add such patterns.
* Adds CI testing in test_findspam.py to verify that the code is detecting text which wasn't previously detected.
* Adds CI testing to verify the entries on the number watchlist and number blacklist can be detected.
* I've previously pushed changes to the number watchlist and number blacklist which normalized the entries, so there isn't, currently, a need for this PR to make changes to those. This was done separately, because it's the area which is most likely to result in merge conflicts and the entries were undetectable, even with the code prior to this PR, as they were.

Testing for this change is around: https://chat.stackexchange.com/transcript/message/59727159#59727159


makyen:
I'm going to go ahead and merge this based on the testing I performed, which included running it live for several hours and comparing against what SD was reporting/not reporting, seeing that running it live didn't take a noticeable amount of CPU (barely registered), and the facts that it A) fixes an issue preventing some detections, and B) prevents people from adding entries which can't be detected.

PR #6777, opened 2021-12-10
Move setup code out of "obfuscated word in {}"; adjust test cases

makyen:
This moves the setup code for the "obfuscated word in {}" rule out of the `obfuscated_word()` function, so it's run once per SD reboot instead of 2 to 4 times per tested post.


tripleee:
Thanks, good to go as far as I am concerned. I have a bigger refactoring to turn it into a class (in fact, I've been thinking several of our nontrivial rules should be turned into classes so we can initialize them only once and then use the resulting object) but I hit a few snags with unrelated changes (I wanted to use the homoglyphs library but it turned out to be less straightforward than I had hoped).

tripleee (approved):

PR #6790, opened 2022-01-25
findspam.py: add regexr (regexr.com) to WHITELISTED_WEBSITES_REGEX

tripleee:
Also, some mild refactoring to improvbe the legibility of the link_at_beginning and link_at_end functions


makyen:
!!/approve
(not that you actually need it :) )

PR #6985, opened 2022-05-13
Don't re-scan unchanged, recently-scanned posts; Add EditWatcher to detect grace period edits

makyen:
This adds every question which is scanned (or the question for any answers) to a SE WebSocket subscription and monitors the questions/answers for edits. If an edit is detected, then the question (and answers) are added to the BodyFetcher queue.

## Changes
* Getting the list of SE Site IDs, `site_id_dict`, [A dependency for `DeletionWatcher`, `EditWatcher`, and watching for posts on Meta sites (not in this PR)]
   *  Moved from in deletionwatcher.py to datahandling.py
   * Uses https://meta.stackexchange.com/topbar/site-switcher/all-pinnable-sites instead of https://meta.stackexchange.com/topbar/site-switcher/site-list, as the all-pinnable-sites route is in JSON and includes IDs for all Meta sites (enabling watching Meta sites, which isn't in this PR, but should be soon).
   * The list is cached and updated daily, rather than updated upon every reboot.
 * Thread safety improvements
    * Lots of changes to use locks for both reading and writing in many places where they are only used for writing.
       * This *isn't* a full review of the code, but I made changes as I saw issues in what I was looking at.
    * Use `call_soon_threadsafe()` instead of `call_soon` in `Tasks`
    * various in DeletionWatcher
    * various in BodyFetcher (BF), datahandling for BF
* Helper functions
   * Helper function `add_to_global_bodyfetcher_queue_in_new_thread()` to enable easier adding to the queue from multiple places and indicating in the thread name from where the item was added.
   * `log_current_thread()` to log with the name of the current thread
   * `append_to_current_thread_name()`
* Main SE WebSocket
   * When `BodyFetcher` is using queue depths of 1, skip pre-testing the 155-questions-active data
 * `EditWatcher`
   * Based on DeletionWatcher, this is an SE WebSocket connection to watch individual questions for edit activity.
   * It receives question IDs from the main 155-questions-active list, the scanning in `BodyFetcher`, and posts scanned manually via SD command
   * When an edit is detected, the question is scheduled for scanning by `BodyFetcher`
* `BodyFetcher`
   * Add recently_scanned_posts:
      * This is a list of posts which have been recently scanned. Currently, it keeps entries for 60 minutes.
      * Prior to sending posts to be checked for being spam, it's first checked to see if the post is in the recently scanned list and that its title, Markdown body, and username content matches what was previously scanned. If the content is a match, then it's not scanned. In addition to catching grace period edits, it also substantially reduces the number of posts scanned, due to not re-scanning posts many times (currently we rescan SO posts quite significantly, due to always fetching all recent posts, not just posts we know about).
      * Posts which were recently scanned, edited and rescanned are now re-reported, if the of detection reasons for the new version are not a subset of the prior set of reasons (exclusive of "blacklisted user"). Previously, posts were only reported once within a substantial period of time, so new versions which are worse/more clearly spam were not reported.
  * Move some longer operations (e.g. disk operations) into Tasks, so each enqueuing/scanning thread isn't as impacted.
     * Don't allow more than one of the same action to be in the Tasks queue at a time for some actions which are identical (e.g. saving data to disk).
  * Don't perform some operations which are only for flovis when flovis is disabled.
  * Rewrite the loop for picking a site from the main site posts queue:
     * Add a determination of current CPU load. If the load is too high (currently >98 out of 100), prevent the current enqueuing thread from launching an API fetch and scan.
     * Add a maximum number of scanning threads per site: currently 5 for SO, 3 each for all other sites
     * Prevent multiple scanning threads from scanning the same post at the same time. This can be quite important for posts which take longer to scan. Prior to limiting threads per site and launching threads when already at high CPU load, an example was seen where ~10 threads were scanning the same post at the same time.
     * Thread name shows which posts are to be processed and which posts have or are being processed, along with the times for some operations. This is primarily for debugging. However, there's data here that would be nice to surface in response to commands. It's currently a hack and really should be completely re-written.
* SD commands
  * `!!/queuestatus` now explicitly states the queue is empty, rather than giving no response.
* SD operation
  * `core_files` updated with missing files, so SD will reboot when non-blacklist files are changed. 

#### Testing on SD for this PR was done in [Makyen Test 02](https://chat.stackexchange.com/rooms/97026/makyentest02) starting [about here](https://chat.stackexchange.com/transcript/97026/2022/5/9), but the more interesting results were in later days, with the final operation [near the end of testing for this PR](https://chat.stackexchange.com/transcript/97026/2022/5/13).


makyen:
Shortly prior to me submitting this PR, it was announced by SE that they have changed the requirements for being able to create grace period edits. Now, users with < 10 reputation will have all revisions to their posts recorded (i.e. they won't be able to make "grace period edits" which are unrecorded by the system).

That adjustment changes the urgency for the need for the code contained in this PR. However, changes include here are still desirable to include in SD for the associated changes/fixes which surrounded the addition of `EditWatcher`. `EditWatcher` itself is still desirable, as it will detect grace period edits by users with > 10 reputation.

makyen:
Still very desirable, which is not currently in this PR, is to add CI tests to cover thee areas of functionality. Currently, there's nothing in the CI testing which covers the functionality contained in bodyfetcher.py.

tripleee:
An impressive improvement, by first glance. I don't know when I can review this properly, but with the testing you have done already, I would not be opposed to merging this before that.

editwatcher.py
```diff
@@ -0,0 +1,160 @@
+# coding=utf-8
+import json
+import os.path
+import requests
+import time
+import threading
+# noinspection PyPackageRequirements
```
tripleee, line 7: stylistic fix: add a space between standard library imports and third-party imports, and again between third-party imports and local imports

editwatcher.py
```diff
@@ -0,0 +1,160 @@
...
+import websocket
+# noinspection PyPackageRequirements
+from bs4 import BeautifulSoup
+from urllib.parse import urlparse
```
tripleee, line 11: import sort order: urllib.parse should be up among the standard library imports

editwatcher.py
```diff
@@ -0,0 +1,160 @@
...
+                        if site_id and max_time <= now:
+                            del self.posts[action]
+                            Tasks.do(self._unsubscribe, action)
+                    if max_time > now and data["a"] == "post-edit":
+                        add_to_global_bodyfetcher_queue_in_new_thread(hostname, question_id, False)
+
+    def subscribe(self, post_url=None, hostname=None, site_id=None, question_id=None,
+                  pickle=True, timeout=DEFAULT_TIMEOUT, max_time=None, from_time=None):
+        if GlobalVars.no_edit_watcher:
+            return
```
tripleee, line 80: Maybe add INFO logging here to make it discoverable when the code goes no-op?

tripleee:
The addition of several new threads and multiple new locks is unnerving. We are getting to the place where we should have an architecture diagram or something to provide an overview of these complexities. Not a blocker for this PR, and I'm hesitant to add a separate issue at this point, but noting this for possible further discussion.

tripleee:
Maybe spell out "RS" in the commit messages? I can guess from context but it's probably going to be hard to figure out if it comes up in `git blame` or etc.

recently_scanned_posts.py
```diff
@@ -0,0 +1,160 @@
+# coding=utf-8
+import time
+from tasks import Tasks
```
tripleee, line 3: More `import` stylistics: empty line here

helpers.py
```diff
@@ -229,12 +229,37 @@ def files_changed(diff, file_set):
 
 
 core_files = {
-    "apigetpost.py", "blacklists.py", "bodyfetcher.py", "chatcommands.py", "chatcommunicate.py",
-    "chatexchange_extension.py", "datahandling.py", "deletionwatcher.py", "excepthook.py", "flovis.py",
-    "gitmanager.py", "globalvars.py", "helpers.py", "metasmoke.py", "nocrash.py", "parsing.py",
-    "spamhandling.py", "socketscience.py", "tasks.py", "ws.py",
-
-    "classes/feedback.py", "_Git_Windows.py", "classes/__init__.py", "classes/_Post.py",
+    "apigetpost.py",
```
tripleee, line 232: Seems like this should perhaps be a separate PR but +1

helpers.py
```diff
@@ -385,8 +385,15 @@ def add_to_global_bodyfetcher_queue_in_new_thread(hostname, question_id, should_
     t.start()
 
 
-def convert_new_scan_to_spam_result_if_new_reasons(new_info, old_info, match_ignore=None):
+CONVERT_NEW_SCAN_TO_SPAM_DEFAULT_IGNORED_REASONS = set([
```
tripleee, line 388: Good catch!

tripleee:
I have reviewed all the commits and added some comments on things which might require some additional polish; but I'm fine with this being merged in its current shape, too.

Please understand, though, that I have a very cursory understanding of the crucial parts of this, and in general the huge ball of hair that is the queue handling around the Stack Exchange API in Smoke Detector (and also Flovis).

It would be nice to have more test cases around this, but the parts which are (to my mind) most likely to fail are probably around deadlock or race condition situations which are quite challenging to create adequate and realistic test coverage for. It should be done, I hope, but I don't think it makes sense to block this PR on requiring that.

In so many words, my approval is based more on trust than understanding.

tripleee (approved):
(What I just said should apparently have been left explicitly as a review. Sorry.)

editwatcher.py
makyen: I'm happy to use that style. It is, however, not something we're consistently using across SD. These lines are direcctly copied from the existing deletionwatcher.py file, upon which editwatcher.py was based.

editwatcher.py
makyen: Basically the same reply here. I'm happy to reorder these `import` statements and apply the style changes you mention, but the current ordering and style is something which is in the existing deletionwatcher.py.

The thing I didn't do, and definitely should have done, is go through each import and remove the ones which aren't actually being used, at least both here in editwatcher.py and remove those which are no longer needed in deletionwatcher.py.

I'll go ahead and make the changes you're asking for along with what I mentioned above.

editwatcher.py
makyen: This `if` should only ever trigger based on the SD owner explicitly providing the `no_edit_watcher` option on the command line when invoking nocrash.py. Logging that here, and in deletionwatcher.py, which does the same thing, would get really noisy for the cases where those are disabled. Those cases are fairly likely to be when someone is doing testing and actually looking at the console logging, so would have an impact on debugging. I'd be OK with an INFO log where we decode the command line and set those values in `GlobalVars`. That wouldn't be all *that* noisy, but might be annoying, given that `EditWatcher` and `DeletionWatcher` are only ever disabled when the user has explicitly specified it in the command line. While those command line options could be used for other purposes (e.g. they are used for the metasmoke instance), they were primarily added for testing.

recently_scanned_posts.py
makyen: OK

helpers.py
makyen: There should be a separate PR. :; Ironically, I have a branch in process which rewrites this to invert the logic. Instead of a list of files to cause a full reboot, it uses a list of files which allow for just reloading the blacklist, or for which SD will complain into chat that the instance needs operator attention (e.g. nocrash.py, requirements.txt, and user_requirements.txt). In other words, IMO, the default for a change should be that there's a full reboot, rather than the default being to just reload the black/watchlists.  While that code is basically done, it needs testing and was definitely out of the scope of this PR. :)

For the inciting change here of adding entries for the new "editwatcher.py" and "recently_scanned_posts.py" files, I'd argue those should be should be in this PR, as this PR is the one adding those files. Given that I was making that change, and had recently encountered a time SD didn't do a full reboot when it should have when another .py file was changed, going through and adding the other missing Python files seemed the right thing to do. It would be quite reasonable to have the fixing up the list to the currently existing files be a separate commit, potentially in a different PR or direct commit, and have only adding the new "editwatcher.py" and "recently_scanned_posts.py" files in this PR. Admittedly, that's how I'd normally do it. I did choose to cut a corner here.

helpers.py
makyen: Thanks.

makyen:
I agree the addition of new locks is concerning. How much of an impact they were having is one of the reasons I've had the debugging output that's in the currently final commit (it's something that I've been using in testing and kept updating to a single last commit, so I could easily remove it) to show how much time was being spent waiting for locks. In all testing, it's been negligible, always showing as, at most, 0.01 second per lock. The current debugging code puts the current state, including effectively being able to see what lock is being waited for into the thread name, which makes it visible when the `!!/threads` command is run. [NOTE: I'm rewriting that logging/reporting (it being really hacky is annoying :; ) and planning to have SD commands to provide reporting of more scanning performance data, including being able to see which posts take a long time to scan, retain seeing that information and the post currently being scanned in the `!!/threads` output.]

I definitely agree there should be better guards against deadlocking and there should be reporting into chat when resources are exhausted for extended periods of time (e.g. CPU, per-site scan threads, etc.). I've intended to have such. I really didn't feel this area of changes was *finished*, but put them into a PR now due to the, at the time, urgency of getting grace period watching available. Ironically, we were informed about SE eliminating grace period edits for low level users while I was actually in the process of writing the initial comment message on this PR. Given that, now, there's not *nearly* as much urgency for detecting grace period edits, I have no problem spending the time to add better handling and reporting of such things, even prior to merging these changes.

The locks which are most likely to be exhausted or deadlock are the ones for the maximum number of scan threads per site, which is logged, but not currently reported into chat, which it should be if it lasts for an "extended" period of time. That's fairly important, as it would indicate if some other deadlock has occurred which caused scan threads to stall (or error out without releasing). It's probably even more important, as I'm *planning* on changing the `test_post()` code to at least partially run the detections in parallel (at least minimally separate out the tests which are asynchronous in nature, i.e. those making network calls, and run those in parallel, rather than everything sequentially).

Similar to the max per-site threads, when SD doesn't launch a new scanning thread due to CPU exhaustion is logged, but it, also, should be reported into chat if it's happening for an "extended" period of time.

editwatcher.py
tripleee: Yeah, I'm fine with leaving that in, and tackling minor code style in a separate PR eventually. I've been meaning to do that, idly.

makyen:
@tripleee Thank you for spending the time to review this, I appreciate it quite a bit.

editwatcher.py
tripleee: Yup, no objection. I haven't tested this code but I have been in similar scenarios with other code where a simple log statement would have saved me some digging to figure out that I was disabling a code flow somewhere deep inside.

makyen:
I agree that a higher level description of the architecture would be helpful. There have been various times when I've had to expend not a small amount of time trying determine from a higher level POV what is supposed to be happening in various areas of SD's code. It would be nice to have it better documented.

makyen:
CI test cases would be *great*. There are, currently, none for the entire area of functionality affected by this PR.

makyen:
@tripleee

> Maybe spell out "RS" in the commit messages? I can guess from context but it's probably going to be hard to figure out if it comes up in `git blame` or etc.

I'm not really sure what you're asking for here.

I agree that it would be helpful to be able to see which, general, area of the code is being affected with each commit when just looking at a list of the commit messages and/or what the commit is doing (e.g. BUGFIX, DEBUG, FIXUP, etc.). At least with userscript repositories where there are multiple separate scripts in the same repository, I've been using a short prefix to commit messages to indicate which script is affected by a commit. I'd be quite happy to adopt something similar for SD to indicate what area of functionality is being affected by a commit. I routinely look at commits as, primarily, just a list of commit messages (with branch threading shown). I'd definitely find it helpful to use something similar here. However, I'm not sure that's what you're inquiring about.

editwatcher.py
makyen: A separate PR would be good for applying a general style template.

Given that I did need to go through and remove dependencies in some files where the need for them had been eliminated, I went ahead and made the style adjustments you've described in several of the files which I had already edited for this PR. A PR with a global `import` style change to all files would probably be a good idea.

makyen:
The recent commits add:
* `import` style changes which were requested in review
* Chat reporting for extended periods of CPU starvation and/or starvation of per-site scanning threads
* Re-implementation of the scan tracking in the thread-name which was considered debug code in prior versions.
   * The information is displayed in the results of the `!!/threads` command and in the console output at the "debug" loglevel.
   * This also changes the `!!/stats` command to provide more information. See [this chat message for an example](https://chat.stackexchange.com/transcript/message/61146837#61146837)
      > :61146836 Posts scanned: 718, Q(398), A(320); Scan time: 776.89; Posts per second: 0.92; Grace period edits: 32; Unchanged posts: 6643, Q(3366), A(3277); No post lock: 11; Errors: 0; Max scan time: 19.17; Max scan time post: [stackoverflow.com/72266934](//stackoverflow.com/q/72266934); Post processing lock: 0.11; Check unchanged: 0.26; Scan question: 677.96; Scan answer: 96.89
* Re-introduced having the scan threads re-check for additional queue entries upon completion of the posts they were invoked to test.  
  I had this in earlier versions, but it introduced problems, which were probably mostly CPU overloading issues. With the limitation on starting a new site/posts set, having it in there doesn't appear to be causing issues. It does resolve the issue that having queue depths of 1 means we need to be, on average, launching scans much more often.

PR #6993, opened 2022-05-17
findspam.py: body_text_repeated rule

tripleee:
This should hopefully identify a recently prevalent pattern.

The rule so far is very specifically tailored to the samples I have at hand; going forward, we might want to try to generalize it for broader use.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -808,6 +808,28 @@ def has_repeating_words(s, site):
     return False, ""
 
 
+# noinspection PyUnusedLocal,PyMissingTypeHints,PyTypeChecker
+@create_rule("body text repeated in {}", title=False, max_rep=10000, max_score=10000)
+def body_text_repeated(s, site):
+    """
+    Do some hacks to reduce the need for regex backtracking for this rule
+    """
+    initial_words = regex.match(r"\A(\w+)\W+(\w+)", s)
+    if not initial_words:
+        return False, ""
+    period = regex.match(
+        r"\A%s\W+%s\W+(\w+(?:\W+\w+)*?\W*)%s\W+%s\b" % (
+            tuple(regex.escape(x) for x in initial_words.groups()) * 2), s)
+    if not period:
+        return False, ""
+    repeats = regex.match(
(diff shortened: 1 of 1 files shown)


makyen:
First off, sorry for not reviewing quickly. I put off doing so a bit when I first saw this, then ended up quite tired and went to sleep.

This seems quite similar in intent to the existing rule "repeating words in {}". Are there cases where that rule and this rule will both match the same text?

It looks like you found and fixed the first issue I had, which was the redundant detection name. And fixed the second issue (that post bodies start with an HTML tag). Just FYI: a `<p>` is not the only HTML tag which can start a post body, but it is, by far, the most common.

#### In the version as of the point you removed it from SD:

You use `(.+?)` in the second regex. That seems over-broad. I'd be much more comfortable if there was a relatively/moderately short limit to the number of characters there. Perhaps a couple/few hundred characters? However, that's probably not all that critical.

The code:
```
repeats = regex.match(r"\A(\w+(?:\W+\w+){%i}\W*){10,}" % (len(period.groups(0)[0].split()) + 1), s)
```
doesn't appear to do what you're intending. It doesn't end up looking for a repeating pattern, it just looks for a bunch of words, which would match nearly anything and return the last portion of the text.

For example, on [this post](https://metasmoke.erwaysoftware.com/post/366947), the first regex does [this](https://regex101.com/r/tTnkxh/1), which then results in the second regex being `\Awhat\W+is\W+your\W+(.+?)what\W+is\W+your\b` and is doing [this](https://regex101.com/r/0oVnFf/1). That makes the final regex be `\A(\w+(?:\W+\w+){2}\W*){10,}`, which is doing [this on that text](https://regex101.com/r/RI5ePw/1). However, it's going to match anything. For example, using the text of this comment as the body text, it does [this](https://regex101.com/r/DXz0WZ/1).

tripleee:
Thanks for the comments. I backed out the hosed code yesterday, but will try again in the near future with a second PR. I was hoping to be able to catch the current spam wave with higher precision, but given that we ultimately remove those posts quickly anyway, I'll just force myself to ignore the urgency I feel.

Issue #6776, opened 2021-12-02
Delete post in the SOCVR chat room when the question git deleted


mitutoyo-ctl:
### What problem has occurred? What issues has it caused?
SmokeDetector is posting a message in the Stack Overflow Close Vote Room (SOCVR).
The message came ~20 seconds after the question has been asked.
Clicking on the message link as fast as I can, the question already has 6 or 7 downvotes and it's being closed by Community bot even before the SOCVR members (me) can take an action.
After that, the SmokeDetector message remains in SOCVR chat.

### What would you like to happen/not happen?
I'd like either 
a) SmokeDetector is not posting the message at all or
b) SmokeDetector is deleting his message after the question got deleted.

makyen:
SmokeDetector should already be deleting any of its reports in SOCVR where the reported post is deleted within 2 minutes (SE system limitation) of when the post is reported. So, I consider this a bug. At a guess, there's a timing issue.

tripleee:
My guess would also be that there is a timing-related problem. In particular, when there is a lot of chat traffic, we have to throttle what gets sent. Perhaps the deletions should bypass the regular queue? But we'd have to repro the situation with some debug logging in place to really see what's going on there.

ThatRyanPerson:
It looks to me like the problem is that deletions aren't working at all: https://chat.stackexchange.com/transcript/message/59782260#59782260

makyen:
Resolved with PR #6994.

PR #6994, opened 2022-05-17
Fix Smokey being able to self-delete messages (poof/delete-force, automated self-deletion, etc.)

Spevacus:
When reconnecting using a modified cookie-login method introduced in #6733 , as we're not calling one of the Chat Exchange "client" object's login methods, the Chat Exchange _worker thread doesn't get restarted, and as a result, message deletion is queued up but not performed. This PR explicitly starts that thread back up.

Had a chat with Makyen about this [here](https://chat.stackexchange.com/transcript/message/61152096#61152096), and for future reference, [he mentions](https://chat.stackexchange.com/transcript/message/61152294#61152294) we should probably implement an abbreviated login function into our fork of Chat Exchange that doesn't rely upon Main/Meta sites being up, but that's more of a "Would be nice"/"Implement if you have time" thing.

This resolves issue #6776 and a sub-issue [mentioned there by Ryan M](https://github.com/Charcoal-SE/SmokeDetector/issues/6776#issuecomment-985796442).

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -90,6 +90,7 @@ def init(username, password, try_cookies=True):
                             if me.id > 0:
                                 client.logged_in = True
                                 logged_in = True
+                                client._thread.start()
                                 log('debug', 'chat.{}: Logged in to chat only using cached cookies'.format(site))
                         except Exception:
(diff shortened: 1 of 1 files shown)


makyen:
!!/approve

Great! Thanks for the PR!

PR #7025, opened 2022-05-26
Allow us to force the regex package to not implicitly cache some regexes at compile time

makyen:
This adds a helper function `regex_compile_no_cache()` which can be used to compile a `regex` without the regular expression entering the `regex` package's implicit cache, which allows us to do so on a per-compilation basis. Unfortunately, the `regex` package only exposes `regex.cache_all()` and `regex.purge()`. The `regex.cache_all()` method turns the cache on and off globally, which is fine in a single threaded application, but has undesired side effects if used dynamically in an application where there are multiple threads, like SD. We have been using `regex.purge()`, which works, but results in a notable number of regular expressions needing to be recompiled into the implicit cache.

Not putting specific compiled regular expressions into the implicit cache can leave the slots available in the implicit cache open for other regular expressions. This is important if we were to have any detections which used a regex which was dynamically created based on the post content. If we did, then those regexes which are used for a single post would fill the cache and result in having to repeatedly recompile the static regular expressions used in the rest of the code.

This PR also uses the `regex_compile_no_cache()` function to compile the regular expressions which are used in regex-only detections, as well as a few regular expressions which are compiled in findspam.py once upon module initialization (i.e. outside of functions). This replaces our prior use of `regex.purge()` to prevent those from filling up the implicit cache.

I also went through all of the code for the existing detections in findspam.py looking for any cases where we were constructing dynamic regular expressions on a per-post basis and didn't see any. It's possible that there's code which is called on a per-post basis outside of findspam.py which should use `regex_compile_no_cache()`. While I intend to look around for it, I have not yet done so.

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -23,7 +23,7 @@ import dns.resolver
 import requests
 import chatcommunicate
 
-from helpers import log
+from helpers import log, regex_compile_no_cache
 import metasmoke_cache
 from globalvars import GlobalVars
 import blacklists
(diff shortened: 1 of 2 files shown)


makyen:
A more global issue: One possibility for watching for the issue of overflowing the `regex` package's implicit cache is that we could periodically check how full the cache is and report if it looks like we're overflowing it. That would allow us to auto-detect if we end up with a regex which is dynamically created based on post content, and also give us some data as to how full we have the cache. Filling it isn't bad. Overfilling it will result in increased CPU usage and reduced scan rates.

Issue #6955, opened 2022-05-06
Failing tests


tripleee:
### What problem has occurred? What issues has it caused?

Several of the recent tests are failures, possibly due to the recent refactoring #6948

For example, https://github.com/Charcoal-SE/SmokeDetector/runs/6320800336?check_suite_focus=true failed with the error traceback below.

### What would you like to happen/not happen?

Tests should succeed when nothing is wrong (-:

May be relevant: When I run the tests locally, I get this warning:

```
test/test_blacklists.py:215
  /home/tripleee/charcoal/SmokeDetector/test/test_blacklists.py:215: PytestUnknownMarkWarning: Unknown pytest.mark.xdist_group - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.xdist_group(name="long_runner_1")
```

Here is the traceback from the example:

```
def test_metasmoke():
        orig_tell_rooms_with = chatcommunicate.tell_rooms_with
        chatcommunicate.tell_rooms_with = dummy_tell_rooms_with
        msg = Fake({
            "owner": {
                "name": "ArtOfCode",
                "id": 121520,
                "is_moderator": False
            },
            "room": {
                "id": 11540,
                "_client": {
                    "host": "stackexchange.com"
                }
            },
            "_client": {
                "host": "stackexchange.com"
            }
        })
        msg_source = "metasmoke is {}. Current failure count: {} " + "({id})".format(id=GlobalVars.location)
    
        chatcommands.metasmoke(original_msg=msg, alias_used="ms-up")
        assert GlobalVars.MSStatus.is_up()
        chatcommands.metasmoke(original_msg=msg, alias_used="ms-down")
>       assert GlobalVars.MSStatus.is_down()
E       AssertionError: assert False
E        +  where False = <function GlobalVars.MSStatus.is_down at 0x7f41f81237f0>()
E        +    where <function GlobalVars.MSStatus.is_down at 0x7f41f81237f0> = <class 'globalvars.GlobalVars.MSStatus'>.is_down
E        +      where <class 'globalvars.GlobalVars.MSStatus'> = GlobalVars.MSStatus

test/test_chatcommands.py:656: AssertionError
----------------------------- Captured stderr call -----------------------------
[...]

PR #7033, opened 2022-05-27
Use threading.RLock() locks for global values in chatcommunicate.py

makyen:
In CI testing, we've been getting a notable number of errors in `test/test_chatcommunicate.py` and some other affected tests. These have indicated real thread safety issues in chatcommunicate.py (and some in the tests themselves). This PR is intended to resolve those issues.

In order to resolve the thread safety issues in chatcommunicate.py and resolve some issues in CI testing this PR:
* Adds and uses `threading.RLock()` locks for each of the global values in chatcommunicate.py, or groups of values which are used in similar areas, within chatcommunicate.py.
* Uses those locks in other modules where those modules directly access and/or manipulate the values protected by a lock.
* Adds some decorators to some tests in test_chatcommands.py to store and restore some of the values which are being protected by the locks. Those decorators are used where the tests intentionally clobber the protected values.
* Adjusts the code in `test_validate_yaml()` to perform some additional tests and display the results for all users, rather than failing after a single problem user and reporting only that one, when there might be additional issues.

These changes should largely resolve the failing tests noted in issue #6955. It's possible that there are remaining issues which are not addressed here, but this should be the bulk of the failing tests I've seen reported.

#### Testing of the changes in this PR was done in the [Makyen Test 02 chat room](https://chat.stackexchange.com/rooms/97026/makyentest02) around [here](https://chat.stackexchange.com/transcript/message/61227909#61227909) for about 4 hours of running a test SD instance doing normal scanning and some chat commands. The latter two commits were not included in that testing, but those commits affect only CI testing.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -15,6 +15,7 @@ import threading
 import time
 import yaml
 import shlex
+import copy
 
 import datahandling
 import metasmoke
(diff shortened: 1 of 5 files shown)


makyen:
We've had 3 more CI failures in the last 24 hours which should be resolved by the changes in this PR.

Adding/adjusting locks is always something which comes with some risk of introducing a deadlock. I would be more comfortable merging this on a weekend, when SE sites are less busy and uninterrupted SD operation is less critical. Thus, I'm going to go ahead and merge this, so it's first introduced over a weekend.

PR #7036, opened 2022-05-29
Add command: !!/unnotify-all-admin <user_id>

makyen:
Osiris is intermittently logging errors when it posts some reports in chat, because [Xnero's old chat account](https://chat.stackexchange.com/users/435118) was deleted while there were still notifications set for that account to get pings for some reports.

This PR adds the command `!!/unnotify-all-admin <user_id>` which can be used by metasmoke admins to remove the notifications for the specified user ID on the chat server on which the command is executed. So, when merged, I expect to execute the command `!!/unnotify-all-admin 435118` in Charcoal HQ to remove the impossible to send pings.

Also related: https://stackoverflow.com/c/charcoal/questions/824

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1804,12 +1804,62 @@ def migrate_notifications():
 @command(whole_msg=True)
 def unnotify_all(msg):
     """
-    Unsubscribes a user to all events
+    Unsubscribes a user from all events
     :param msg:
     :return: A string
     """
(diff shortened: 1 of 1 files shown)


gparyani:
Does this work on only the one instance in use at the time the request is executed, or on all instances?

makyen:
Currently, there is no synchronization of notification between SD instances. So, this works only on whatever instance(s) receive the command.

PR #7074, opened 2022-06-10
Thread safety 003: GlobalVars, git file actions, etc.

makyen:
This is another step towards thread safety for SD.

### This PR does the following:
* Add threading locks:
   * for values stored in the `GlobalVars` class in globalvars.py
      * Some values which are currently set once in ws.py do not have locks. As long as they are only set once when ws.py starts up, the need for a lock is low.
   * for file operations for any operation which affects files in the `git` repository.
      * This lock is ignored under some conditions (e.g. under some conditions when rebooting)
   * for reloading the blacklists/watchlists
   * for values in the `MetasmokeCache`
   * for `datahandling.last_feedbacked`
* Reorganizes the values in `GlobalVars` to make it more clear what values a lock applies to and how some of the values relate to each other.
* Increase use of existing locks in places where the values are accessed, but where the lock was not being obtained
* Adds a `thread_safe` value to be used in `Rule` detections based on functions to indicate that the function is thread safe. If the detection does not indicate that it's thread safe, then it's permitted to be run in a single thread at a time.
   * Currently, all function based `Rule` detections are not marked as `thread_safe`. Each function should be reviewed.
* Use Charcoal's fork of the `regex` package, which has additional changes for thread safety, primarily affecting the implicit `regex` cache.
* Use a `copy.deepcopy()` of the regex used for each regular expression based `Rule`.
   * I need to do more investigation into the code in the `regex` package to see if this is really needed here and within the implicit `regex` cache, but having it is the safe option.
* Adjust some uses of `.acquire()` and `.release()` to the `with` semantic where the necessary additional handling of exceptions, etc. didn't already exist. I did not do a global search for such situations, so there may still be some more.
* Make sure we're obtaining a lock for each set of values loaded from pickles in `datahandling.load_files()`
* Some bug fixes for bugs which I found while working on threading:
   * Don't double @create_rule on `mostly_non_latin()`
   * CPU starvation chat warning: inconsistent use of a variable name
   * `remove_all_from_notification_list()` didn't pass in the `msg._client.host`, which could result in removing the wrong notifications if there was a chat user ID conflict across chat servers.
* DRY: replace a use of `"errorLogs.db"` with `ErrorLogs.DB_FILE`, so the filename is only specified in one place.
* Adjust the titles in the parameterized test_findspam.py tests in order to make them more easily identifiable when looking at CI testing output.

This PR is not a complete and comprehensive review of thread safety for SmokeDetector. It is a step in the right direction and should, hopefully, prevent some of the intermittent failures we are experiencing in CI testing.

### Things still needed (not a comprehensive list):
* Review ChatExchange and our use of it with respect to thread safety.
* Review each of the function based `Rule` detections for thread safety.
* Use a better method of forcing single threaded use of function based `Rule` detections than just stalling waiting for the lock.
   * It would be better to skip them and return to them once other detections have run.
   * The main area where this may affect us is for detections which have long, network-based runtimes. While this will impact us, we will need to evaluate how much of an impact it has, but it can, probably, be alleviated by reviewing such rules for thread safety.
   * Doing this may get pushed into a planned/expected change to making the running of the individual `Rule` detections multi-threaded. One of the overall intents is to get to the point where the detections can be run in parallel, rather than single-threaded, at least to some degree.
* `regex` package:
   * review the core code to see if making deep copies is actually needed
   * Coordinate the changes in the `regex` package with the author of the package (e.g. there's a documented, changeable global value we don't use which would be a notable hassle to make thread safe, so need to determine if and how doing that is desired; submit a PR for existing changes)
* Review the `phonenumbers` package for thread safety.
   * Currently, this is handled by making the one detection which uses it as single threaded only. It may end up that we continue to use that method to handle this, rather than spend time making the package thread safe.


makyen:
Further testing indicates this is now introducing unacceptable side-effects in scanning. I'll re-submit a PR once those are resolved.

PR #7212, opened 2022-07-29
Add watch of Google redirection service

CoconutMacaroon:
Context is [here in the transcript in Charcoal HQ](https://chat.stackexchange.com/transcript/message/61684882). If I haven't already, I can also add a link to my GitHub from my SE profile to verify it is me (I'm `cocomac` on SE). Here's the tl;dr of the rationale:
* This domain AFAIK hasn't ever been legitimately used on SE
* It has ten TPs, but it is still a redirector not a spam site so I'm not blacklisting it outright as in theory it could be used not in spam
* It doesn't appear to be documented by Google (maybe somewhere?) so I don't think it will really get used legitimately

And a quick explanation of the regex as it is huge:
```none
(an OPTIONAL protocol) (some subdomain up to six characters*)(any Google domain)(a protocol**)(a domain name)
```
In the Google domain, the protocol is optional, but for the redirect thing to work, there has to be a protocol there

A few notes:
* This was too big for a chat message on SE, hence the manual PR
* If someone knows what the number before my username in the file is, feel free to fill it in properly


watched_keywords.txt
```diff
@@ -41515,3 +41515,4 @@
...
...le\.lu|google\.lv|google\.com\.ly|google\.co\.ma|google\.md|google\.me|google\.mg|google\.mk|google\.ml|google\.com\.mm|google\.mn|google\.ms|google\.com\.mt|google\.mu|google\.mv|google\.mw|google\.com\.mx|google\.com\.my|google\.co\.mz|google\.com\.na|google\.com\.ng|google\.com\.ni|google\.ne|google\.nl|google\.no|google\.com\.np|google\.nr|google\.nu|google\.co\.nz|google\.com\.om|google\.com\.pa|google\.com\.pe|google\.com\.pg|google\.com\.ph|google\.com\.pk|google\.pl|google\.pn|google\.com\.pr|google\.ps|google\.pt|google\.com\.py|google\.com\.qa|google\.ro|google\.ru|google\.rw|google\.com\.sa|google\.com\.sb|google\.sc|google\.se|google\.com\.sg|google\.sh|google\.si|google\.sk|google\.com\.sl|google\.sn|google\.so|google\.sm|google\.sr|google\.st|google\.com\.sv|google\.td|google\.tg|google\.co\.th|google\.com\.tj|google\.tl|google\.tm|google\.tn|google\.to|google\.com\.tr|google\.tt|google\.com\.tw|google\.co\.tz|google\.com\.ua|google\.co\.ug|google\.co\.uk|google\.com\.uy|google\.co\.uz|google\.com\.vc|google\.co\.ve|google\.vg|google\.co\.vi|google\.com\.vn|google\.vu|google\.ws|google\.rs|google\.co\.za|google\.co\.zm|google\.co\.zw|google\.cat)\/url\?q=https?:\/\/.?
```
tripleee, line 41518: Repeating `google` for every subdomain is bulky and inefficient, `foo\.bar|foo\.baz` can be refactored to `foo\.(?:bar|baz)` (or, of course, in this simple example case, to `foo\.ba[rz]`).

tripleee (changes requested):
Please refactor as suggested in review comment.

makyen:
Is there a reason to limit the possible matching google domains?

CoconutMacaroon:
@makyen I don't get what you mean... I don't want to match any random domain that has that pattern, so that is why I match the Google ones only. They are from [here](https://www.google.com/supported_domains) if you are wondering

watched_keywords.txt
tripleee: Much better, but a couple of minor tweaks still: the optional `https://` can be removed (there is no need to backslash the slashes, but just take out the thing entirely, as it doesn't contribute anything to the match) and the `.{0,6}` should probably be something like `(?:\w+\.)*+` ... or actually removed as well, actually.

tripleee: The ????? should be replaced with a Unix timestamp, the time you created this would usually be used, but the time right now (1659070435) is a suitable approximation.

tripleee (changes requested):
Still requesting a few tweaks.

tripleee:
If I get what @makyen is suggesting, I suppose it could be simplified to just `google(?:\.\w++)++` (which would however then also erroneously match `google.subdomain.of.example.com`)

watched_keywords.txt
CoconutMacaroon: Timestamp done, I've removed the `https://`, but I changed the `.{0,6}` to `.{0,16}` because if I remove it and someone adds the protocol, bookmarking on the regex will make it not work, so I think that is needed

makyen:
I was going to wait for a MS search using the original or second version of the regex to complete, so I could get an idea about what results that regex is returning and compare it against the regex below. However, the MS search I started using the original regex is *still* running (> 20 minutes), so I'm guessing a bit at what you're trying to find.

the regex here seems roughly equivalent to:
```
google(?:\.[a-z\d]+){1,2}/url\?q=https?://
```
which has results of [23 Total / 16 TP (69.57% TP) / 5 FP / 2 NAA](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title_is_regex=1&title=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3A%5C.%5Ba-z%5Cd%5D%2B)%7B1%2C2%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&body_is_regex=1&body=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3A%5C.%5Ba-z%5Cd%5D%2B)%7B1%2C2%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&username_is_regex=1&username=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3A%5C.%5Ba-z%5Cd%5D%2B)%7B1%2C2%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&or_search=1)

I'd note that there are, potentially, a notable number of uses of this on SE. Unfortunately, I'm not seeing a good way to search just for what is desired on SE, so to find how many, I'd probably start with a search for "/url\?q=https://" [in text](https://stackexchange.com/search?pagesize=100&q=%22%2Furl%5C%3Fq%3Dhttps%3A%2F%2F%22) (65) and [in URLs](https://stackexchange.com/search?q=url%3A%22%2Furl%5C%3Fq%3Dhttps%3A%2F%2F%22) (182). That's more than will actually match the regex, so I'd probably sample several of the hits to see if they are actual matches. Some of them certainly are, based just on the text on the search results page.

makyen:
The regex I'm suggesting above specifically is not capturing the actual intended destination URL, because doing so would prevent it from being caught by other watchlist entries. Having it be able to be caught by other watchlist entries will allow the people looking at the `why` data to see if the destination domain itself is currently watched on its own.

CoconutMacaroon:
Yeah, this regex was slow. IIRC, it was 10/0/1 for an earlier revision (should still be the same-ish results), that was just far more verbose. My guess is there won't be any meaningful amount of people legitametly using this as it doesn't appear to be documented anywhere that I saw. I'll try that one in a moment and look at it, one moment

watched_keywords.txt
tripleee: There should still be a separator, you want to match www.google but not shmgoogle

CoconutMacaroon:
@makyen I think I fixed it so that the end domain isn't included in the match, but I'm rusty on bookmarking so I'd double-check that. In theory, this should only watch the proper Google domains + subdomains, but shouldn't catch the part at the end

makyen:
If you're wanting to be more restrictive for the domain, then you could use something like:
```
google(?<!-google)(?:\.com?)?\.[a-z]{2,3}/url\?q=https?://
```

It gets the same [23 Total / 16 TP (69.57% TP) / 5 FP / 2 NAA](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&title_is_regex=1&title=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3C!-google)(%3F%3A%5C.com%3F)%3F%5C.%5Ba-z%5D%7B2%2C3%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&body_is_regex=1&body=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3C!-google)(%3F%3A%5C.com%3F)%3F%5C.%5Ba-z%5D%7B2%2C3%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&username_is_regex=1&username=(%3Fs)google(%3F%3C%3D(%3F%3A%5E%7C%5Cb)google)(%3F%3C!-google)(%3F%3A%5C.com%3F)%3F%5C.%5Ba-z%5D%7B2%2C3%7D%2Furl%5C%3Fq%3Dhttps%3F%3A%2F%2F(%3F%3A%5Cb%7C%24)&or_search=1) results on MS, but substantially limits the possible matching domains, which would prevent many of the ways that the domain could be spoofed, while still matching all the valid domains at the link you provided.

I don't really see a need to prevent catching spoofed google-like domains for this, as spoofed domains that use this format are quite likely to be TP anyway.

CoconutMacaroon:
@makyen I totally agree, done!

makyen:
I looked through the first 10 of the [181 hits for "/url\?q=https://" in URLs on SE](https://stackexchange.com/search?q=url%3A%22%2Furl%5C%3Fq%3Dhttps%3A%2F%2F%22). All 10 were valid hits for this regex. So, we'll probably be wanting some limitations on matching in order to reduce FP. Based on those 10 posts, I'd recommend excluding some domains which are very unlikely to be FP.  In order to do that, I'd suggest the following regex:

```
google(?:\.[a-z\d]++){1,2}/url\?q=https?://(?!(?:www\.)?(?:google.com|(?:(?:[a-z\d]++\.){0,2}(?:stack(?:exchange|overflow)|superuser|serverfault|askubuntu|stackapps)\.com|mathoverflow\.net))[/\s"'<])
```

It excludes the main SE domains, and their subdomains, along with the main google domain. I expect that there will be more for which it would be helpful to exclude, but it will be easier to determine what's needed with more data, which is most easily acquired by adding it as a watch and then revisiting after some time of it being on the watchlist.

makyen:
I realized a couple of minutes ago that I went back to using the less restricted google domain version of the regex in my most recent suggestion. I was going to post a note switching it to the more restricted version, but then I noticed that [this post](https://metasmoke.erwaysoftware.com/post/312518) uses "maps.google.it.ao/url?q=https://". That domain, google.it.ao, isn't on the list of valid Google domains, and the regex which is more restrictive for the google domain wouldn't match it. However, it seems like it's something we should match. Admittedly, there's way more than enough other instances of this pattern in that particular post, such that yet one more match isn't needed to be caught, in that post. However, it could be beneficial for other posts.

Overall, I'd prefer to start with the regex which is less restrictive for the Google domain (i.e. the regex I most recently posted). We can always tighten it up later after finding out we don't need it more relaxed.

makyen:
How would you like to finish this off? I see a couple of general options:

1. Resolve the merge conflicts and merge this PR. Resolving merge conflicts can be a bit frustrating, particularly the first times you do it. It's made a bit less convenient by the fact that the watched_keywords.txt file is too large to use the GitHub online tools (although, personally, I don't like how GitHub resolves merge conflicts, as it can leave a pre-merge commit which is quite disingenuous as to what changes are really being made). There are a variety of ways to resolve the conflicts. Which you'd use depends on what you're comfortable with.
2. You originally created this PR because the watch entry was substantially larger than what would fit into a chat message. Given that the regex which was the end result of everyone working on this is substantially shorter and can fit into a chat message, the easiest thing to do is to just close this PR and use:
    ```
    !!/watch google(?:\.[a-z\d]++){1,2}/url\?q=https?://(?!(?:www\.)?(?:google.com|(?:(?:[a-z\d]++\.){0,2}(?:stack(?:exchange|overflow)|superuser|serverfault|askubuntu|stackapps)\.com|mathoverflow\.net))[/\s"'<])
    ```
   in Charcoal HQ to create a new, automatically-generated PR.

I'm fine with either way, or another way, if you want to do something else. If it was me, I'd probably choose to use `!!/watch`, because it's now possible and easy.

makyen:
Superseded by #7221.

PR #7226, opened 2022-08-04
Add a second verification pass for initial NS verification failures

makyen:
The intent of this change is to prevent CI testing from failing due to intermittent DNS resolution issues. While we already do a second pass, it doesn't seem to resolve the issues we're seeing on particular domains. This collects the entries which failed the first round of testing and re-tests all of the failures (many/most will have already been tested twice, so this is a third and fourth attempt, but delayed).

In addition, with the prior code we don't have easy access to the entry which is is failing. With this change, the failing entries are collected into a list in both a first and second pass. If we continue to have issues with a specific domain that isn't actually a failure and something we want to remove from then list, then we can just exclude it from being reported as an error in CI testing.

### DIFF
--- a/blacklists.py
+++ b/blacklists.py
@@ -6,9 +6,10 @@ import regex
 import yaml
 import dns.resolver
 import sys
+import time
 
 from globalvars import GlobalVars
-from helpers import log, log_current_exception, color
+from helpers import log, log_current_exception, color, pluralize
 
 
 def load_blacklists():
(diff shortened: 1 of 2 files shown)


makyen:
I'm going to go ahead and merge this. Usually I prefer to leave PRs open for at least a day to see if anyone has comments/issues/suggestions/questions/etc. With this one, it's an intermittent issue in CI testing which is causing notable additional work by people checking out why the CI failures occurred, and the failures are relatively common (at least a couple/few per day, it seems, at least for one domain in particular).

It's not clear to me that this will really resolve the issue. We may end up having to explicitly exclude from reporting the domain which is most commonly causing the CI failures, but this should at least move us in the direction of doing that. I'll keep an eye out for additional DNS failures and see if we need to add an explicit exclusion from reporting.

PR #7230, opened 2022-08-15
Remove Erica from Erica rule (sic)

tripleee:
Proposed by Glorfindel in chat: https://chat.stackexchange.com/transcript/message/61791436#61791436

I didn't change the rule's description to keep the history intact on the metasmoke side. Perhaps that is not something we want, though?

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -2730,7 +2730,7 @@ create_rule("several emoji in title",
 # Parenting troll
 # This rule is currently broken and detects nothing. It appears it was supposed to be for usernames.
 create_rule("bad keyword in {}",
-            r"(?i)\b(erica|jeff|er1ca|spam|moderator)\b",
+            r"(?i)\b(jeff|spam|moderator)\b",
             all=False, sites=["parenting.stackexchange.com"],
             title=False, body_summary=True, max_rep=52,
             rule_id="bad keywords: erica, jeff, er1ca, etc., bodies, parenting")


tripleee:
The `rule_id` actually doesn't seem to reflect anywhere important so it could be changed too if we don't simply remove the whole rule.

There's a comment which incorrectly states that the rule does nothing at the moment, so that should be removed or refactored, too.

makyen:
`rule_id` is not currently recorded anywhere. Changing it is not a problem at all. The primary requirement is that each `rule_id` must be unique among all `Rule` entries. That uniqueness is enforced. SD will raise an exception if there's a duplicate `rule_id`, which will cause CI testing to fail. My current intent for `rule_id` is primarily as a debugging/development aid, both for short term (e.g. being able to identify which `Rule` detected something and how long it took) and, potentially, longer term use (e.g. statistics as to how long specific `Rule` entries take). Current uses for `rule_id` are:
* config file's ability to only enable matching `Rule` entries (used while developing/debugging a `Rule`); and 
* reporting of Rules with excessive execution times into chat (functional, but not currently enabled for main SD instances).

Given that ["Jeff" came back and used "Erica" in a post 10 months ago](https://metasmoke.erwaysoftware.com/post/332061), it's probably a good idea for us to keep it in there, probably add the new username which Erica is now using and, potentially, other existing Parenting moderators. "Jeff" really seemed to have a grudge against "Erica" and has demonstrated that they might come back and do the same thing a year or more later. Just `(?:erica|er1ca)` is [30 Total / 23 TP (76.67% TP) / 5 FP / 2 NAA](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=%28%3Fs%29%28%3F%3A%5E%7C%5Cb%29%28%3F%3Aerica%7Cer1ca%29%28%3F%3A%5Cb%7C%24%29&site=90), which is about a 7% better %TP rate than the full `Rule`.

I suggest we change this to a "potentially bad keyword in {}" instead of a "bad keyword in {}". The record for this detection `Rule` doesn't justify the "bad keyword in {}". It's fine for us to have it as "bad keyword in {}" while the troll is active, but we really should revert it back to what its record justifies wrt. %TP ([currently 46 Total / 32 TP (69.57% TP) / 10 FP / 4 NAA](https://metasmoke.erwaysoftware.com/search?utf8=%E2%9C%93&body_is_regex=1&body=%28%3Fs%29%28%3F%3A%5E%7C%5Cb%29%28erica%7Cjeff%7Cer1ca%7Cspam%7Cmoderator%29%28%3F%3A%5Cb%7C%24%29&site=90)), which is "potentially bad keyword in {}".

The other alternative is to just disable it or remove it. However, I'd prefer to keep it as a "potentially" rule, as it's really not generating any significant number of FP/NAA over time. It's been a `Rule` for [nearly 7 years](https://github.com/Charcoal-SE/SmokeDetector/commit/e7c0caf77d4e8e7b0851a5efa0d9649d87c8f607), with a total of 14 FP+NAA over that time.

As to the
```
# This rule is currently broken and detects nothing. It appears it was supposed to be for usernames.
```
comment, I'm unsure why I added it. It was probably a brief note to myself which I should have deleted after taking a better look at the rule, and definitely prior to committing. It's inaccurate and shouldn't be in the code.

makyen:
I'm happy to keep this open, if desired.

I've [pushed a commit](https://github.com/Charcoal-SE/SmokeDetector/commit/2078d8213e2d886bb1d490d61b37319606d867f0) to change this `Rule` from "bad keyword in {}" to "potentially bad keyword in {}" and remove the inaccurate comment.

tripleee:
Actually, I guess the linked commit takes care of everything I really wanted to change here. I concur that we probably want to keep Erica in the watch with these changes.

PR #7231, opened 2022-08-20
Don't forward non-reply chat messages to MS as comments

makyen:
This removes the feature of SD forwarding chat messages which are not direct replies to reports to MS as comments.


teward:
I think this is critical to include,  Long has the problem of 'irrelevant conversations' been added in, but only recently has it spawned across MULTIPLE different channels to unrelated MS posts.  Plus, attribution violation is bad.

I'm all for including and merging this.

PR #7237, opened 2022-08-30
Improve message for watch during ongoing CI testing

CoconutMacaroon:
Tl;dr when someone does `!!/watch` during CI testing it will ping them instead of just saying that it reloaded the watchlist. This PR explains what is happening (it will be added once the current CI testing is done), so it is clear what is happening

Context is [in the transcript](https://chat.stackexchange.com/transcript/message/61892018#61892018), phrasing suggested by @codygray.

TODO:
The spot in the code is here:

https://github.com/Charcoal-SE/SmokeDetector/blob/master/gitmanager.py#L260-L263

Do we also need to adjust that message for blacklisting? I suspect yes, but I should check that as well

### DIFF
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -260,7 +260,7 @@ class GitManager:
         if op == 'blacklist':
             return (True, "Blacklisted `{0}`".format(item))
         elif op == 'watch':
-            return (True, "`{0}` will be added to the watchlist once the ongoing CI testing is done".format(item))
+            return (True, "Added `{0}` to watchlist".format(item))
 
     @classmethod
     def remove_from_blacklist(cls, item, username, blacklist_type="", code_privileged=False, metasmoke_down=False):
--- a/gitmanager.py
+++ b/gitmanager.py
@@ -260,7 +260,7 @@ class GitManager:
         if op == 'blacklist':
             return (True, "Blacklisted `{0}`".format(item))
         elif op == 'watch':
-            return (True, "Added `{0}` to watchlist".format(item))
+            return (True, "`{0}` will be added to the watchlist once the ongoing CI testing is done".format(item))
 
     @classmethod
     def remove_from_blacklist(cls, item, username, blacklist_type="", code_privileged=False, metasmoke_down=False):
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -554,7 +554,18 @@ def do_blacklist(blacklist_type, msg, force=False):
             return None
         except Exception:
             pass
-    return result
+    additional_state = (" However, the blacklists were not reloaded at this time. The most likely issue"
+                        " is another change was made on SD's master branch on GitHub and everything"
+                        " is waiting for CI to pass and MS to update the deploy branch.")
+    if GlobalVars.MSStatus.is_down():
+        additional_state += (" But, metasmoke is currently down, so that won't happen. Someone with write"
+                             " permission to SD's GitHub reository will need to manually update the deploy"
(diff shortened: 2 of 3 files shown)


makyen:
I'd rather not adjust this in gitmanager.py. The code in that function knows nothing about the issue which is causing the blacklists not to be reloaded (because that's attempted in the calling function). At the point the changed text is returned, `add_to_blacklist()` has completed its task and the blacklists/watchlists have been updated on GitHub. While we currently only call that function from one place, making the current change pushes assumptions about how the function will be used from a different module into the gitmanager.py module, which doesn't seem ideal.

IMO, it would be better to make a change in [chatcommands.py on line 557](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L557). At that point we know that
1. SD successfully made a change to the master branch on GitHub, but 
2. SD did *not* reload the blacklists, even though it normally would. 
We *don't* actually know why the blacklists weren't reloaded. There could be multiple possible reasons. The most common reason is that an earlier change was made to the master branch on GitHub and we're waiting for CI to finish and MS to update the GitHub deploy branch. However, there could be more complex issues.

I'd suggest changing [chatcommands.py on line 557](https://github.com/Charcoal-SE/SmokeDetector/blob/master/chatcommands.py#L557) to something like:
```
    additional_state = (" However, the blacklists were not reloaded at this time. The most likely issue"
                        " is another change was made on SD's master branch on GitHub and everything"
                        " is waiting for CI to pass and MS to update the deploy branch.")
    if GlobalVars.MSStatus.is_down():
        additional_state += (" But, metasmoke is currently down, so that won't happen. Someone with write"
                             " permission to SD's GitHub reository will need to manually update the deploy"
                             " branch. Usually, this means you should ping an MS admin.")
    else:
        additional_state += (" That could take a few to several minutes. If SD doesn't automatically reload"
                             " the blacklists or automatically reboot after that time, then someone should"
                             " investigate why that hasn't happened.")
    return result + additional_state
```
That text is, however, probably a bit verbose.

If the above change is made, then it's necessary to also change at least one CI test case in test_chatcommands.py. If the above change is used, then I'd suggest changing [line 337 in test_chatcommands.py](https://github.com/Charcoal-SE/SmokeDetector/blob/master/test/test_chatcommands.py#L337) to:
```
        assert wrap_watch("male enhancement", True).startswith("Hahaha ")
```

gitmanager.py
```diff
@@ -260,7 +260,7 @@ def add_to_blacklist(cls, blacklist='', item_to_blacklist='', username='', chat_
         if op == 'blacklist':
             return (True, "Blacklisted `{0}`".format(item))
         elif op == 'watch':
-            return (True, "Added `{0}` to watchlist".format(item))
+            return (True, "`{0}` will be added to the watchlist once the ongoing CI testing is done".format(item))
```
makyen, line 263: This shouldn't change. The code in gitmanager.py only knows that the item was, in fact, added to the watchlist. It has no information about if the rest of the system thinks that it's acceptable to reload the blacklists and/or reboot.

makyen (changes requested):

gitmanager.py
CoconutMacaroon: Done, that was what I originally suggested, I've reverted that [in this commit](https://github.com/CoconutMacaroon/SmokeDetector/commit/b62860a6a8781f7e74b7f01e6c1d4c879461bb01)

makyen (approved):

PR #7239, opened 2022-09-03
Add per-entry ignore of NS validation failures in CI testing

makyen:
This adds the ability to ignore NS validation failures during CI testing on a per-entry basis in the *nses.yml files. If an entry in those files has the value `ignore_validation_errors: true`, then validation errors will not cause CI failures.

This is primarily in order to be able to prevent CI failures due to the persistent intermittent issues with validating `dns-parking.com.`.  This PR adds `ignore_validation_errors: true` to the `dns-parking.com.` entry in watched_nses.yml, which will result in not seeing CI testing failures when that domain fails DNS validation.

There probably should be some automated testing which still validates, and fails on, entries which have `ignore_validation_errors: true`, as we do want to know about such domains if they *actually* stop having a DNS entry. However, this PR just patches over the persistent frustration of looking at CI failures and finding that it's yet another time that `dns-parking.com.` had an intermittent problem. No attempt is made here to address the issue of wanting some type reporting for such domains when they actually go away.

### DIFF
--- a/blacklists.py
+++ b/blacklists.py
@@ -356,44 +356,59 @@ class YAMLParserNS(YAMLParserCIDR):
         # 20 max_workers appeared to be reasonable. When 30 or 50 workers were tried,
         # it appeared to result in longer times and intermittent failures.
         with ThreadPoolExecutor(max_workers=20) as executor:
-            return list(executor.map(self._validate, list_to_validate, timeout=300))
+            results = list(executor.map(self._validate, list_to_validate, timeout=300))
+        entries_with_exception = [entry for entry in results if entry is not True]
+        exception_entries_not_ignored = [entry for entry in entries_with_exception
+                                         if entry.get('ignore_validation_errors', None) is not True]
+        ingored_failures = [entry for entry in entries_with_exception
+                            if entry.get('ignore_validation_errors', None) is True]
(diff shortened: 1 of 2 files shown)


tripleee (changes requested):
Updating the schema should also update the `Schema_version` at the top of the YAML file.

I never got around to properly formalizing this, but the purpose of this information is to track changes in the schema, so let's apply it here.

tripleee:
In theory, I dislike the idea of bypassing quality checks when the quality sucks, but the DNS problems of our CI is beyond our control, so I guess let's do what we can to focus on the things we _can_ affect. +1

tripleee:
I guess we still want this, even though the problem has been somewhat mitigated by other means. Sorry for being slow to review.

makyen:
While I do intend to finish this at some point (i.e. integrate the requested changes for which I have some work along those lines), I wasn't intending to work on this at the moment. The activity today was because I managed to mistakenly delete the branch on GitHub, which automatically closed the PR. Restoring the branch on GitHub and reopening the PR re-ran CI, which, of course, had failures, due to being 13k commits behind. So, the work today was just to `rebase` the already pushed commits onto the current master branch.

tripleee:
For the record, the requested change simply requires the `Schema_version:` key at the top of the file to get a new value.

PR #7294, opened 2022-10-03
Updated GH Actions

Mou-Ikkai:
Updated Checkout, Cache, and Setup Python to the latest version!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,17 +18,17 @@ jobs:
           - '3.8'
           - '3.10'
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v3.0.2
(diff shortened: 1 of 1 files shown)


makyen (changes requested):
Why the version lock to a specific minor and patch version level on each of these? Doing so is counter to the current recommendations for each of the packages for which you've changed the version.

What are the benefits and drawbacks of the changes between the major versions which we were using and the versions specified here? Specifically `actions/setup-python` being moved from major version 3 to major version 4 and `actions/cache` changing from major version 2 to major version 3. I'm not actually objecting to changing those major versions. In general, updating is *usually* good, but there should be some evaluation done as to the benefits gained and consideration for any possible drawbacks.

PR #7300, opened 2022-10-10
Update README.md

hackerman321:


tripleee:
It is unclear how this brings any benefits or why you submitted it. Can you please update the PR with a rationale for why this should be changed?

PR #7323, opened 2022-10-22
Create sry.md

Cipher-unhsiV:


tripleee:
We do support Hacktoberfest, but we definitely do not welcome junk submissions like this.

makyen:
Given that you've created 11 PRs which are effectively identical to this PR into 10 different projects, I'm marking this as spam and reporting it to GitHub as abusive. While your desire to participate in Hacktoberfest is great, this is really **not** the way to do it.

Issue #7345, opened 2022-11-17
"airline's" really isn't obfuscated "airlines"


tripleee:
### What problem has occurred? What issues has it caused?

False positive on English s-possessive; noted here:
https://chat.meta.stackexchange.com/transcript/message/9348931#9348931

### What would you like to happen/not happen?

Probably need to add a special case for this.

PR #7346, opened 2022-11-17
finspam.py: obfuscated_words: avoid FP on simple possessive

tripleee:
(Closes: #7345)


tripleee:
(Sorry for the noise, should be good to merge now.)

Issue #7333, opened 2022-11-05
PullApprove v2 shutting down


tripleee:
### What problem has occurred? What issues has it caused?

I received an email from PullApprove:

> Hey!
> 
> This is Dave from PullApprove. I'm emailing you because you've logged into PullApprove v2 in the last year. While I'm glad people are still finding a use for it, it's been *years* since it has gotten any meaningful updates as most people have moved to PullApprove v3. 
> 
> And now we've started working on v4! (https://www.pullapprove.com/) 
> 
> Long story short, it's time to move on. *On January 1st, 2023, we will be shutting down v2.* Any active subscriptions will be cancelled and all data will be deleted 30 days after. 
> 
> I'd encourage you (and your company) to take a look at what we're building for v4 and see if it's a good fit for you. In addition to more powerful code review assignment and approval policies, we're moving into reporting (especially for auditing purposes), formal review bypass procedures, Slack/calendar integration, and even CLAs for organizations that use them. Potentially even some similar features for managing/triaging issues. 
> 
> I'll personally help you migrate over -- just reply to this email and we'll get it going!
> 
> Thanks for being a PullApprove user all these years!
> 
>Dave Gaeddert (Owner)

### What would you like to happen/not happen?

First of all, decide whether we still think this is useful. _Requiring_ a review is sometimes a nuisance, and I've been thinking maybe we could get by with a lighter process which doesn't require external tooling.

makyen:
We've, effectively, been operating without really using PullApprove for the last couple of years. While I agree with the intent of requiring a review of changes by at least someone else other than the person making the changes, we haven't been doing so in an *organized* manner for quite some time. In that time, the people contributing non-automated PRs, primarily @tripleee and myself, have been leaving the PR open for some amount of time in order for it to be looked at by others. However, we have not really been using the PullApprove service to gate PRs (i.e. to enforce that PRs are actually reviewed).

I'm OK with either dropping PullApprove, while maintaining that we do want to provide for the review of "significant" changes/PRs, or moving to the newer version of PullApprove and keeping it as an indicator that we do intent for PRs to see some amount of review by contributors other than the author of the PR.

tripleee:
Thanks for following up. If there are no objections, I'll prepare to tear it out in a week or two.

PR #7383, opened 2022-12-01
.pullapprove.yml: remove this file

tripleee:
(Closes: #7333)

I believe the linked bug report already established a vague consensus that this component adds more friction than we really need. We can always reconsider if we manage to attract more contributors, but for the time being, my impression is that we will be fine with the regular Github Pull Request flow mechanism for getting reviews.


tripleee:
I tried to look for places in the code which would somehow depend on this, but I could not find any. There is a comment in the `!!/approve` command's implementation but it merely pulls in an image which was apparently generated or hosted by PullApprove.

teward (approved):
No objections here, seems like we've been working without it for the better part of an age so.

makyen (approved):

PR #7453, opened 2023-01-03
Add chat user starball to users.yml and Tavern in rooms.yml

tripleee:
Chat around here: https://chat.meta.stackexchange.com/transcript/message/9388273#9388273


starball5:
ahoy!

PR #7650, opened 2023-03-15
add special-character arrows

superplane39:
add special-character arrows to the "link following arrow" reason


tripleee:
There are probably more Unicode arrows we could add in addition to these two, but sure.

PR #7716, opened 2023-04-15
Add a config option to override the prefix in SD chat messages; use config file for prefix configuration

CoconutMacaroon:
This PR adds a config option that provides a setting to override the entire prefix for messages posted by SD in chat. This would allow one to add any link as a prefix (not just to the GitHub repo), but a user could also set the prefix to something else as well.

It also changes the repo/bot name prefix options to use the `config` file instead of env variables.


makyen (changes requested):
Configuration information like this should primarily be in the "config" file, rather than as an environment variable. As far as I'm aware, the use of environment variables is a legacy artifact that is not currently in use.
While it might be useful to also have it available as an available environment variable, the config file seems like the right place for this type of configuration information.

CoconutMacaroon:
I changed the one to a config option, I’ll change the other ones there to config options as well later today

CoconutMacaroon:
I've converted the other two options to the `config` file. I also tested this change [here](https://chat.stackexchange.com/rooms/148260/bot-testing), and it seems to work as expected.

makyen (approved):

makyen:
I'm sorry for the delay in merging this. I *really* need to get better at keeping track of these and actually dealing with them.

PR #7726, opened 2023-04-18
blogspot regex updates for blacklisted_websites.txt along with corresponding removals from watched_keywords.txt

jeffschaller:
generalized most of the blogspot regexes in blacklisted_websites.txt; this resulted in some duplication with older entries in watched_keywords.txt, so this PR removes those older entries.

I don't know why the preview here doesn't show `everything4comp\.blogspot` remaining in blacklisted_websites; maybe because it was previously duplicated with my regex change?


tripleee:
Thanks a lot for the cleanup; much appreciated!

PR #7728, opened 2023-04-19
generalized most of the blogspot entries to remove the TLD portion

jeffschaller:


jeffschaller:
The initial commit's changes were an even change (adds versus removes), but based on https://github.com/Charcoal-SE/SmokeDetector/actions/runs/4747755704/jobs/8433075667 I then removed duplicates and removed entries from watched_keywords that had been escalated to blacklisted.

Issue #7649, opened 2023-03-15
Don't report Staging Ground posts to other rooms than CHQ


tripleee:
### Is your feature request related to a problem? Please describe.

https://m.erwaysoftware.com/posts/uid/stackoverflow/75721102 was reported in SOCVR. One of the participants in the room was surprised that they were unable to visit the post. https://chat.stackoverflow.com/transcript/message/56097157#56097157

### Describe the solution you'd like

My proposal would be to not report Staging Ground posts to other rooms, but I suppose an alternative could be to clearly mark them up to identify them as Staging Ground posts. (This would be attractive for other reasons as well.)

### Additional context

Many users do not have access to the Staging Ground, and so reporting posts that they cannot access or assess seems like a waste of their time.

codygray:
I [partially] disagree with this. Staging Ground posts are *specific* to Stack Overflow; thus, it makes sense to report them in SO-specific chat rooms, including SOCVR (unless the SOCVR room owners specifically object, which I have not seen evidence of). If anything, it's actually more likely that lurkers in SO-specific chat rooms will have access to the SG than it is lurkers in CHQ will have access to it.

Of course, I'd have no objection to clearly indicating that the posts are from the Staging Ground in the message itself.

makyen:
Overall, working on this made quite a bit more sense when Staging Ground was in private beta, which was when this was proposed. After this was proposed, but prior to it being implemented, the company shut down the Staging Ground private beta. When the Staging Ground returned, about a year later, it was directly released publicly. There were, initially, significant issues with access, because the requirements for access were, initially >= 500 reputation *and* having posted >= 2 well received questions. That was changed after a couple of weeks to only requiring >= 500 reputation, which is similar to the first review queues on SO.

Now that the Staging Ground is public and the requirement for viewing Staging Ground posts is only >= 500 reputation, I feel that this is, at best, a low priority.

* I do object to a blanket "don't show".  
* I don't object to each room having the option to not show them, similar to how a room can be configured to show or not show experimental reasons.
* I don't object to indicating that such posts are in the Staging Ground, but someone not being able to view the post on the site is *vastly* less of a problem than it was when Staging Ground was in a private beta and only very few people were able to view the posts.

codygray:
Given what @makyen already said, combined with the fact that these posts have been being reported everywhere since the public re-launch of the Staging Ground without any complaints that I've seen, I think we can close this.

[...]

PR #7827, opened 2023-06-03
Add a Staging Ground message indicator to reports on Staging Ground posts

Papershine:
As mentioned in #7649, it would be helpful to indicate that if a post might be hidden due to being in the Staging Ground. This pull requests implements one of the suggestions of displaying that information in the message report. Sample:

> [ SmokeDetector | MS ] No whitespace in title (71): SomeQuestion‭ (Staging Ground) by some-user on stackoverflow.com

Note: This change might potentially break certain userscripts on those reports?

### DIFF
--- a/spamhandling.py
+++ b/spamhandling.py
@@ -211,7 +211,7 @@ def build_message(post, reasons):
         username = post.user_name.strip()
         escaped_username = escape_format(parsing.escape_markdown(username))
         user = "[{}\u202D]({})".format(escaped_username, poster_url)
-        
+
     # If the post URL is a staging ground link, indicate that; ensure that the post is on Stack Overflow
     if shortened_site == "stackoverflow.com" and "staging-ground" in post.post_url:
         staging = "(Staging Ground) "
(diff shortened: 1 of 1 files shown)


tripleee:
An additional complication might be how it affects the logic for forcing the final message to be within the maximum length allowed by the chat API. I'm not in a place where I can investigate that, but I'm noting it for when we can do a proper review.

Thanks for the contribution!

makyen (changes requested):
This "request changes" review is really "needs verification, with possible needs changes". With Staging Ground currently shut down and empty, I'm unable to verify that `post.post_url` will have a URL which contains the text "staging-ground" (note: "/staging-ground/" would be a better test). The `post.post_url` is derived from the `link` value returned from the SE API for the question. I suspect that what will be supplied by the SE API is the canonical `/questions/<post ID>/...` URL, rather than the staging ground URL.

SD was changed to parse `/staging-ground/<post ID>` URLs, but that seems to have been as a result of my realizing that SD needs to do that in order to accommodate user's providing such URLs in chat commands, rather than as the `link` value returned by the SE API.

If you verified that the `link` value provided by the SE API has "staging-ground" in it prior to submitting this PR, then that's sufficient for this concern.

Papershine:
I recall verifying with the SE API that they also return a `link` with `staging-ground` in it somewhere, but my memory is hazy about the exact format. Will reconfirm when (or if) Staging Ground is online again.

codygray:
I wanted to re-open this, but I cannot do so because "The repository that submitted this pull request has been deleted." Instead, I posted [a new issue](https://github.com/Charcoal-SE/SmokeDetector/issues/13136).

PR #8304, opened 2023-10-04
Update GitHub Actions CI testing to Python 3.12 and flake8 to 6.1.0

makyen:
* Update GitHub Actions CI testing to use Python 3.12 instead of Python 3.11 for the most recent version tested, as Python 3.12 was recently released.
* Update the requirements.txt entry for flake8, which is used only in CI testing, from `=~3.8.0` to `=~6.1.0`, which is the current version. This is needed, because 3.8.0 does not appear to be compatible with Python 3.12 (i.e., there were [functional errors](https://github.com/makyen/Charcoal-SE-SmokeDetector/actions/runs/6409188108/job/17399847914) and updating was easy with minimal changes to our code).
  Updating to 6.1.0 required some minor changes for:
  * Do not compare types, use 'isinstance()' (E721)
  * Missing whitespace after ',', ';', or ':' (E231)

  I'm unsure why we locked the version to `=~3.8.0` (see [Compatible release](https://peps.python.org/pep-0440/#compatible-release)). I *assume* it was due to not wanting to deal with periodic additions of new tests and/or changed tests, so I've left the requirement as `=~6.1.0` , rather than `>=6.1.0`.


editwatcher.py
CoconutMacaroon, line 1: The original version contained !=. The updated version does not do that, and checks if it is an instance, rather than it not being an instance. Elsewhere, you've added a not to the condition to address this.

Is this status-bydesign or

CoconutMacaroon, line 1: The original version contained !=. The updated version does not do that, and checks if it is an instance, rather than it not being an instance. Elsewhere, you've added a not to the condition to address this.

Was this status-bydesign and I’m missing something or does a not need to be added here…?

I normally probably wouldn’t have checked, but I noticed a potential issue in CHQ, and wanted to mark the line as it seemed like a possible cause

CoconutMacaroon (commented):
-

PR #8380, opened 2023-10-21
README: Added Our Contributors

Dev7g:


tripleee:
While we welcome Hacktoberfest contributions, I find the value of this to be dubious. Doesn't Github already make it easy and convenient to see who contributed to this project? Adding an external dependency for this very marginal feature seems more brittle than useful.

makyen:
I, also, find this dubious for this project. I agree with @tripleee's points about adding an additional external dependency being more brittle than useful and that the contributor's list is already easily available in the stock GitHub interface. In addition, I'm uncomfortable with what feels like over-aggrandizement of the contributors. I would have felt better about adding something like this if the wording for the section heading wasn't "Amazing" and included the "⭐" emoji, the combination of both of which seem quite over the top, but I probably would have objected to either of those two portions of the text on their own.

In other words, while I believe it's important to acknowledge our appreciation for the work that our contributors have provided, this feels over the top. In addition, it doesn't acknowledge the large volume of contributions made by people which are entered into the project's repository as commits authored by SmokeDetector.

PR #8502, opened 2023-11-01
Moved zero-TP entries from blacklisted_websites.txt into watched_keywords.txt

jeffschaller:
Please consider this PR with caution, since an unfrozen caveman lawyer wrote it!


makyen:
@tripleee I'd appreciate you taking a look at these, particularly the ones later in the blacklist. A notable number of them are ones which you added. At least some of those I suspect to be from the Server Fault event video spam wave as the destination of a link shortener. If you could do a basic check on those, at least from memory, as to if they should remain on the website blacklist, that would be helpful. Thanks.

tripleee:
I manually merged these to master, with the changes I had proposed. Unfortunately, there will be no explicit merge PR.

PR #10121, opened 2024-02-19
Update rooms.yml to add privileges for Cow to SOCVR

rschrieken:
see https://chat.stackoverflow.com/transcript/message/57110674#57110674

### DIFF
--- a/rooms.yml
+++ b/rooms.yml
@@ -640,6 +640,7 @@ stackoverflow.com:
       - 72095    # Ryan M
       - 11421497 # cigien
       - 7156005  # Vega
+      - 11355926 # Cow
       # NOTICE: please don't add new users until you have Room Owner approval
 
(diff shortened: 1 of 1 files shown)


rschrieken:
I assumed users.yml didn't need changes as the record for that user was already present, complete and correct AFAICT

rooms.yml
```diff
@@ -640,6 +640,7 @@ stackoverflow.com:
       - 72095    # Ryan M
       - 11421497 # cigien
       - 7156005  # Vega
+      - 11355926 # Cow
```
tripleee, line 643: The correct network user ID for Cow is 15725245 (see `users.yml`). This file needs the network user, not the local ID. (I too wish all this was less byzantine.)

tripleee (changes requested):

PR #11283, opened 2024-05-23
Restore the blacklisters-approved/rejected icons in PR comments

Spevacus:
We used to store these icons behind a camo.githubusercontent.com link, but attempting to access that link throws a 403 now.

Example PR comment where the image is no longer visible: https://github.com/Charcoal-SE/SmokeDetector/pull/11282#issuecomment-2127452764

The source for these images is from https://shields.io/ 's [static badge](https://shields.io/badges/static-badge) generator. It has an API that this PR uses to generate and return a specified label with a given style.

We just need a simple green/red coloring for the "blacklisters|approved" and  "blacklisters|rejected" actions we perform, and the URL for that is very simple.

Here's how these two URLs will be rendered in a comment:

Approved with SmokeyApprove

Where the markdown is `Approved with SmokeyApprove`

Rejected with SmokeyReject

Where the markdown is `Rejected with SmokeyReject`

Shields.io has been around for a loooong time, since 2013 (?) and is unlikely to be deprecated in the near term as far as I can tell. An alternative approach, which we could do, is to store the image in SE's i.sstatic domain instead and refer to that. I am not against changing this to do so instead of relying upon shields.io's API, if requested.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -676,9 +676,8 @@ def approve(msg, pr_id):
             msg._client.host, msg.owner.id)
         comment = "[Approved]({}) by [{}]({}) in {}\n\n![Approved with SmokeyApprove]({})".format(
             message_url, msg.owner.name, chat_user_profile_link, msg.room.name,
-            # The image of (blacklisters|approved) from PullApprove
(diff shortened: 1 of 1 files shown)


tripleee:
Thanks for the contribution, and sorry for leaving it unreviewed for so long.

Issue #8642, opened 2023-11-14
Feature request: the ability to revoke your own PR


fastnlight0:
### Is your feature request related to a problem? Please describe.
Not really, as you can request for a blacklister to reject your PR, but it would make the blacklister's life easier if someone could do it themselves.

### Describe the solution you'd like
Either a `!!/revoke` command, or the ability to `!!/reject` yourself

### Describe alternatives you've considered
As mentioned above, you can request for a blacklister to do it, but many people would prefer having this possible

### Additional context
None.

gparyani:
FYI: this was already requested as #1639 and declined.

gparyani:
@fastnlight0 D'oh, got it confused. Stoopwafel-ified.

PR #11285, opened 2024-05-23
Allow users to reject their own watch/blacklist PRs

Spevacus:
Implements feature request #8642

This PR allows users to reject/close (including `-force`) their own auto-created watch and blacklist PRs.

This was tested in [this chatroom](https://chat.stackexchange.com/transcript/message/65699938#65699938) (where I removed SD ignoring its own messages when considering them for commands and some other changes, the full diff of which is available [here](https://github.com/Charcoal-SE/SmokeDetector/pull/11285/commits/d3993d17b832c66085ef56a5cd063cf09dd07ff1) where I mistakenly committed them to this branch and had to remove them).

One small change I made was making the `payload` argument for `get_pull_request` an optional arg with a blank string as its default. This change has no effect elsewhere in the project as `get_pull_request` isn't used anywhere aside from in this PR's changes, and you don't need to supply a payload when performing a GET request for a PR anyway.

I also took the liberty to remove the "so the user can learn from their mistakes" content from the required-rejection CmdException. I have a couple of reasons for this removal:
- This wording makes no sense if the user is rejecting their own PR
- We frequently reject duplicate PRs. In such a case, there is no mistake being made.

## Known limitations:
This PR operates off of the chat messager's chat id, and so does submitting a watch PR. This can be different across chat servers, such as if a user submits a watch PR in SO chat and attempts to self-close it in Charcoal HQ. This will yield a `"You need blacklist manager privileges to reject pull requests that aren't created by you."` error.

Given that 95% of all watch/blacklist PRs occur in Charcoal HQ, I consider this an acceptable limitation.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -722,7 +722,7 @@ def reject(msg, args, alias_used="reject"):
     pr_author_id = regex.search(r"(?<=\/users\/)\d+", pr_body).group(0)
     self_reject = int(pr_author_id) == int(msg.owner.id)
     if not code_permissions and not self_reject:
-        raise CmdException("You need blacklist manager privileges to reject pull requests " \
+        raise CmdException("You need blacklist manager privileges to reject pull requests "
                            "that aren't created by you.")
     if len(reason) < 20 and not force:
         raise CmdException("Please provide an adequate reason for rejection that is at least"
@@ -734,10 +734,10 @@ def reject(msg, args, alias_used="reject"):
     message_url = "https://chat.{}/transcript/{}?m={}".format(msg._client.host, msg.room.id, msg.id)
     chat_user_profile_link = "https://chat.{}/users/{}".format(msg._client.host, msg.owner.id)
     rejected_by_text = "[Rejected]({}) by [{}]({}) in {}.".format(message_url, msg.owner.name,
-                        chat_user_profile_link, msg.room.name)
+                                                                    chat_user_profile_link, msg.room.name)
     if self_reject:
         rejected_by_text = "[Self-rejected]({}) by [{}]({}) in {}.".format(message_url, msg.owner.name,
-                            chat_user_profile_link, msg.room.name)
+                                                                            chat_user_profile_link, msg.room.name)
     reject_reason_text = " No rejection reason was provided.\n\n"
     if reason:
         reject_reason_text = " Reason: '{}'".format(reason)
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -15,7 +15,7 @@ from metasmoke import Metasmoke
 from blacklists import load_blacklists, Blacklist
 from parsing import *
 from spamhandling import check_if_spam, handle_spam
-from gitmanager import GitManager
+from gitmanager import GitManager, GitHubManager
 import threading
 import random
 import requests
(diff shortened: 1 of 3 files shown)


Spevacus:
The current error is due to the GitHubManager not being spun-up during testing (or at least I think) - I'll resolve this at a later time.

tripleee:
Thanks for this contribution; I'll take a closer look when I am less swamped.

chatcommands.py
```diff
@@ -717,11 +717,21 @@ def reject(msg, args, alias_used="reject"):
         reason = ''
     force = alias_used.split("-")[-1] == "force"
     code_permissions = is_code_privileged(msg._client.host, msg.owner.id)
-    if not code_permissions:
-        raise CmdException("You need blacklist manager privileges to reject pull requests")
+    self_reject = False
+    try:
+        pr_json = GitHubManager.get_pull_request(pr_id).json()
+        if 'body' in pr_json:
+            pr_authored_by_rejector = regex.search(r"(?<=\/users\/)" + str(msg.owner.id),
```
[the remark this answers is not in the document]
tripleee: I guess the link actually contains the chat user ID for the creator of the PR, so that part should be fine.

tripleee (changes requested):
On the whole, the code looks neat, but the commit history is really messy; can you rebase this to a more logical 4-5 commits max (realistically 1-3 I guess)?

gitmanager.py
```diff
@@ -73,7 +73,7 @@ def comment_on_thread(cls, thread_id, body):
         return response.json()
 
     @classmethod
-    def get_pull_request(cls, pr_id, payload):
+    def get_pull_request(cls, pr_id, payload=""):
```
tripleee, line 76: Is this change necessary and useful?

tripleee: I guess this actually fixes a bug which had gone unnoticed?

Spevacus: Yeah, a payload isn't really required when sending a GET request broadly speaking. If you're supplying a payload, generally you'd be wanting to do so in a POST request. In this case, we just need to perform a GET to get a PR, so a blank payload is sufficient. This argument, really, could be completely removed.

chatcommands.py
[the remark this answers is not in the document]
Spevacus: re: Commit history: Yeah, the commit history is horrendous, I agree. I'll clean that up.

`get_pull_request` returns a request response in string format. I call `.json()` on that response here to just convert it to json, then check for a `'body'` in it. If it doesn't exist in the json'd response, the request failed or had another issue, and I just let error handling after this account for that (as messing with our unit testing would be a scope change I'm not comfortable venturing into).

If the chat user ID is different from the user ID in the PR body (generated by the SmokeDetector git user) then it'll just not match here. Further, this won't affect PRs that weren't opened by the SD git user by virtue of the a

For the regex thing, yeah, I think I can drop the backslashes.

makyen:
Personally, I don't have a problem with there being lots of commits, as that's sometimes useful in order to group changes into conceptual blocks, although in my own work, I do *tend* to go through and tidy commits into conceptual changes when possible/reasonable.

I do, however, have a problem with commit messages which are something like just "Update chatcommands.py". Such a commit message is counterproductive, because it gives almost no information as to what the changes are in the commit. Commit messages are part of what guides both reviewers and people using the code in the future when looking through what changes were made and what to take a more detailed look at. Please see #3 in [CONTRIBUTING_PULLREQS.md](https://github.com/Charcoal-SE/SmokeDetector/blob/master/CONTRIBUTING_PULLREQS.md).

I, generally, allow such commit messages to slip through from time to time, but there are a lot of there here and you're already saying you'll make changes to the overall set of commits.

Spevacus:
Yeah, some of them were 100% unintentional (merged in a branch that I didn't intend). I'll be certain to clean them up when I have a bit more time.

Spevacus:
I've closed this in favor of #11549 since I wasn't able to do crazy rebase magic to make the commit history better here.

Issue #1639, opened 2018-02-22
Add the ability for users to retract manual reports in any room


gparyani:
Right now, if I file a report that is by mistake, others tell me that my report is improper, or the wrong post gets reported due to a bug (a current known bug is that the *question* sometimes get reported if an *answer* is reported as a full-form link), there's nothing much I can do, other than drop into Charcoal HQ from whatever other room I was using to file the report (N/A if the report was from CHQ itself) and reply to my own report saying that it's a false positive. Unfortunately, the report may spread to other rooms during that time, and it wastes others' time as they also have to flag the report as a false positive. 

As such, there should be the ability for users to retract their own manual reports, provided that no one has given feedback on them.

quartata:
You can just use `delete-force`

gparyani:
@quartata What if you didn't realize until more than two minutes later, after the time Smokey can delete its own messages? You can't use it then.

angussidney:
The only way to 'retract' a post (other than marking it as FP which is sufficient enough IMO) is to delete the MS record. Currently only Undo and Art have the ability to do this, and even then we reserve it for exceptional circumstances.

gparyani:
@angussidney So just to be clear here, you're considering this request declined? In that case, can you please clearly explain why each bit of my initial reason for implementing it is wrong? I know it sounds like a bit of a harsh request, but I'd legitimately like to know.

ArtOfCode-:
Yeah, gonna call this declined. Actually _retracting_ a report (as in making it as if you never reported in the first place) is a lot of work to implement correctly - mostly on the MS side, where you have to remove not only the post record but anything that touches it; but also on the Smokey side, where we'd like to delete the message and any feedbacks to it, and drop it out of any lists that it got put in as a result of feedbacks or report.

Doing that amount of work is... not _entirely_ pointless, but given that we can apply a social solution that's much easier than the technical solution, I'd rather do that. So, steps to retracting a report:

 - Mark the report as an FP, preferably via a chat reply to Smokey rather than through MS so that you can add some explanation.
 - If _necessary_ ping Undo or me to remove the MS record. We tend to do that only for test/debugging reports, or for content so objectionable that we don't want it in the database (we have a database of literal spam, so that's a high bar to hit - think child porn, we've deleted a couple of those before).
 - Use `delete-force` if you're in the 2-minute window; if you're after that but you still want the message deleted, ping a Charcoal-local mod.

angussidney:
[...]

PR #11549, opened 2024-06-18
Allow users to reject their own watch/blacklist PRs (v2: Now with cleaner commit history!)

Spevacus:
This is #11285 but with a less messy commit history (and a small regex adjustment).

I'd have done rebase-foo on THAT PR instead, but my Git wizardry is... Novitiate. That's a word, right?

---

Implements feature request #8642

This PR allows users to reject/close (including `-force`) their own auto-created watch and blacklist PRs.

This was tested in [this chatroom](https://chat.stackexchange.com/transcript/message/65699938#65699938) (where I removed SD ignoring its own messages when considering them for commands and some other changes, the full diff of which is available [here](https://github.com/Charcoal-SE/SmokeDetector/pull/11285/commits/d3993d17b832c66085ef56a5cd063cf09dd07ff1) where I mistakenly committed them to this branch and had to remove them).

One small change I made was making the `payload` argument for `get_pull_request` an optional arg with a blank string as its default. This change has no effect elsewhere in the project as `get_pull_request` isn't used anywhere aside from in this PR's changes, and you don't need to supply a payload when performing a GET request for a PR anyway.

I also took the liberty to remove the "so the user can learn from their mistakes" content from the required-rejection CmdException. I have a couple of reasons for this removal:
- This wording makes no sense if the user is rejecting their own PR
- We frequently reject duplicate PRs. In such a case, there is no mistake being made.

## Known limitations:
This PR operates off of the chat messager's chat id, and so does submitting a watch PR. This can be different across chat servers, such as if a user submits a watch PR in SO chat and attempts to self-close it in Charcoal HQ. This will yield a `"You need blacklist manager privileges to reject pull requests that aren't created by you."` error.

Given that 95% of all watch/blacklist PRs occur in Charcoal HQ, I consider this an acceptable limitation.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -720,7 +720,7 @@ def reject(msg, args, alias_used="reject"):
     try:
         pr_json = GitHubManager.get_pull_request(pr_id).json()
         if 'body' in pr_json:
-            pr_authored_by_rejector = regex.search(r"(?<=\/users\/)" + str(msg.owner.id),
+            pr_authored_by_rejector = regex.search(r"(?<=/users/)" + str(msg.owner.id),
                                                    pr_json['body'])
             self_reject = pr_authored_by_rejector is not None
     except Exception as e:
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -15,7 +15,7 @@ from metasmoke import Metasmoke
 from blacklists import load_blacklists, Blacklist
 from parsing import *
 from spamhandling import check_if_spam, handle_spam
-from gitmanager import GitManager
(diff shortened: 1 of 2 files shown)


gparyani:
FYI: this was already requested at #1639 before and was declined.

tripleee:
Tested and approved.  https://chat.stackexchange.com/rooms/154945/room-for-tripleee-and-ad-oisee

tripleee:
Sorry for the really long time this took. I'll try to plow through the backlog of pending PRs over the next few weeks.

PR #11828, opened 2024-07-01
added dominobet, dewapoker and poker88 IP addresses

Cowm00:
https://metasmoke.erwaysoftware.com/post/485596 and https://metasmoke.erwaysoftware.com/post/485597


tripleee (changes requested):
Thanks for the PR! However, the intention is to keep the file sorted alphabetically. Could you move these entries to the right place in the file?

Also, while the comments are vaguely helpful, I would still suggest to remove them, just to keep this consistent with the rest of the file. Over time, comments like these lose their value; we rarely need to know with this specificity why something was added, and if you do, it's not unreasonable to find this information from Halflife's transcript. (There are periods when Halflife was down or unwell, so it doesn't always work, but that's what I've been doing.)

tripleee:
Small nitpick still; it's unintuitive, but 131 sorts before 14 alphabetically. (Yes, this sucks for human consumption; but the reason to keep it sorted is to keep it convenient for the forthcoming SD command.)

Also, if you could, please rebase so that the PR is a single commit.

Cowm00:
> Small nitpick still; it's unintuitive, but 131 sorts before 14 alphabetically. (Yes, this sucks for human consumption; but the reason to keep it sorted is to keep it convenient for the forthcoming SD command.)
> 
> Also, if you could, please rebase so that the PR is a single commit.

Sorry tripleee, but I'm not sure I understand. 131? I don't see 131 in the IP addresses I added. Also, how do you rebase on the Github GUI/website? I'm a total newb.

makyen:
@tripleee If we want only one commit, which is reasonable in this case, GitHub has a "Squash and merge" option in the "Merge pull request" dropdown menu. The main drawback, as I recall, is that it doesn't produce a merge commit, but that's not something that's strictly needed.

tripleee:
Yeah, I think I'd prefer a merge commit but in this case, I'll just squash it.

Issue #11540, opened 2024-06-16
DeletionWatcher: only one post can be watched per question


makyen:
The `posts` data structure only allows for a single post to be watched per `action`. Each `action` is a question. This means that only the first question or answer on the question that is reported will be watched for deletion.

All reported questions and answers should be watched and their deletion reported.

makyen:
This was resolved.

Issue #11539, opened 2024-06-16
DeletionWatcher: answers are never re-watched upon reboot


makyen:
Rewatching an answer upon reboot currently depends on `GlobalVars.post_site_id_to_question[post_site_id]` to have the question ID for the answer. That is only populated in spamhandling.py and is not stored persistently (i.e., it's empty upon reboot). It would be better to store the question ID in the datawatcher pickle, so it doesn't need to be discovered upon reboot (and the question ID isn't available from the SE API unless there's at least one comment (it's available in the `link` provided for the comment).

makyen:
@codygray  No, that PR explicitly doesn't resolve this issue, #11539, or #11540.

PR #11839, opened 2024-07-02
EditWatcher and DeletionWatcher bug fixes

makyen:
This fixes a few/several issues in EditWatcher and DeletionWatcher:
* EditWatcher and DeletionWatcher weren't actually rewatching the questions they were watching when those WebSocket was "recovered". This was a bug I introduced when going a bit too far with DRY in that code. Basically, they were both attempting to resubscribe to their `self.socket` WebSocket prior to the new WebSocket reference being stored in `self.socket`
* More aggressively unsubscribe to questions in EditWatcher. Questions were previously unsubscribed only if there was something received about that question after the time expired. Now, all expired questions are unsubscribed upon each receipt of a message. This keeps the total number of questions subscribed to from expanding throughout the time between reboots.
* EditWatcher wasn't actually saving its pickle, so posts weren't rewatched upon reboot.
* Move saving both pickles into the main `exit_mode()` code, which is executed even if we're exiting due to an exception, so we don't need to repeatedly save the pickles.

There are still more issues to handle in DeletionWatcher, particularly #11539 and #11540.

Testing of this was done in [Makyen+Test+02 about here](https://chat.stackexchange.com/transcript/97026?m=65883051#65883051). Primary verification was done by adding additional logging to the WebSocket actions and manually looking through the console logs.

### DIFF
--- a/deletionwatcher.py
+++ b/deletionwatcher.py
@@ -92,8 +92,8 @@ class DeletionWatcher:
             except websocket.WebSocketException as e:
                 ws = self.socket
                 self.socket = None
-                self.socket = recover_websocket(self.__class__.__name__, ws, self._subscribe_to_all_saved_posts, e,
-                                                self.connect_time, self.hb_time)
+                self.socket = recover_websocket(self.__class__.__name__, ws, e, self.connect_time, self.hb_time)
+                self._subscribe_to_all_saved_posts()
                 self.connect_time = time.time()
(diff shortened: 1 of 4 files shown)


makyen:
Given that this fixes some the recovery of the WebSockets, which is something that we seem to need to do *way* to often, I'm going to go ahead and merge this.

PR #12206, opened 2024-07-22
Add the type of watch or blacklist (e.g., keyword, website, number) to the title of the PR

### LINKED ISSUE #12205: In non-privileged watch and blacklist PRs, make the automatic title indicate the type of watch or blacklist
Currently, the title of automatic PRs filed as a result of commands from non-privileged users indicate whether the user is requesting a watch or a blacklist, but don't indicate the specific type of request the user is making (e.g., which blacklist the user is requesting to add to). The only way to know is to open the PR and look.

I propose that the titles be reformatted to state something like:

- "Amy Rose: Watch number 1234567890"
- "Dr. Eggman: Blacklist keyword (?-i:Chaos Emeralds)"
- "Silver the Hedgehog: Blacklist website gray\\.fur"

This seems important to me since PRs can be approved in chat without opening them up on GitHub. The automatic message posted when a PR is opened only indicates the title, so there is no way to catch, for example, a user mistakenly adding a website to the keyword blacklist without clicking through to GitHub, which users aren't likely to do if the command was executed in a room other than Charcoal HQ.

gparyani:
Resolves #12205. See explanation there.


Spevacus:
Tested this here: https://chat.stackexchange.com/rooms/154419/conversation/testing-pr12206

Looks good to me.

Issue #4398, opened 2020-08-20
Allspam should also accept a reason argument


gparyani:
The `!!/report` command accepts an optional reason argument, to allow the reporter to add in a reason for why they're reporting a given post or set of posts, so it's clear to others.

However, the same reason argument isn't accepted for the `!!/allspam` command. Using that command is no different from going through a user's profile and `!!/report`ing all their posts, so it doesn't make sense to not also allow supplying a custom reason when using that option.

Can we make the `!!/allspam` command also take a custom reason, just like the `!!/scan` and `!!/report` commands?

tripleee:
Would be nice if this could be extended to `!!/addblu` too, and `!!/rmblu` I suppose for orthogonality.

makyen:
@tripleee What would you want us to do with a comment on `!!/addblu` and `!!/rmblu`? For `!!/rmblu`, as far as I know, we don't record the reason a user was removed from the user blacklist, at all. Removing them from the user blacklist removes the entry for the user with nothing remaining. For `!!/addblu`, the only thing currently recorded is the URL of the chat message. Are you wanting to expand that and record more information?

tripleee:
My main use case is that it's convenient to provide a rationale in the chat message itself, in case somebody would search for it. In fact, I tried to do that earlier today, and was mildly miffed that this command worked differently from e.g. `!!/report`. But I agree that we should probably have a more disciplined treatment of this information if indeed we allow users to submit it.

makyen:
OK. Please create a separate issue for those commands. Where and how the changes would be made will be notably different. Someone taking on that issue or this issue shouldn't need to do the other one in order to resolve the issue.

PR #12818, opened 2024-08-23
Let !!/allspam accept a custom close reason

Andrew5057:
If a custom reason is supplied to `!!/allspam`, appends "Reason: {Custom reason}" as a new line to the auto-generated "User manually reported by..." note. Calls without a custom reason are unaffected.

The reason must be separated from the URL by a single space and enclosed in double quotes.

Heavily inspired by the reason handling in `!!/report`.

Addresses [#4398](https://github.com/Charcoal-SE/SmokeDetector/issues/4398).


Andrew5057:
Testing is occurring in a [gallery chat room](https://chat.stackoverflow.com/transcript/message/57597652) on a local SD instance.

Posts are not reported to metasmoke; instead, I am reporting the values of the `URL` and `why_info` variables in chat, plus providing a summary of the results.

Formatting is a little inconsistent, but each entry consists of an `!!/allspam` call, a brief summary labeled "Results" the values of the `url` and `why_info` variables, and an optional "Behavior" section for calls that require more detail.

talflon:
For `!!/report` the custom reason also gets added to Metasmoke as a comment. It would be nice to do the same for `!!/allspam`.

PR #13017, opened 2024-09-01
Add ability to have users_custom.yml

makyen:
This adds the ability to override the users.yml file with a users_custom.yml file for testing. The feature needed for running test instances that was missing is the ability for the runner of the test instance to give a user SD privileges without changing users.yml from the text that's stored in the GitHub repository. This PR resolves that problem by adding the ability to have a users_custom.yml file that is conceptually similar to rooms_custom.yml.


tripleee:
Confirmed working and useful; thanks!

PR #13126, opened 2024-09-06
Watches: Migrate phone number out of regexes

micsthepick:
Some time ago, Mayken wrote a heavily obfuscated phone number matching regex. This migrates it to watched_numbers.txt.

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -66255,3 +66255,4 @@
 1725949134	tripleee	accurate[\W_]*+steels?(?!\.com(?<=accuratesteels\.com))(?<!(?-i:^accuratesteels$))
 1725949604	Cow	ebillpk\.com
 1725951045	tripleee	cre8iveskill(?!\.com)(?<!(?-i:^Cre8iveSkill$))
+1725951129	Dragonrage	bachnghe\.com\.vn


codygray:
The CI checks are failing because you did not follow the correct format. The items in each line of `watched_numbers.txt` are *tab*-separated. Spaces will not work. It must be:

   1725623085→micsthepick→919145958860

where `→` is an actual tab character (`\t`).

micsthepick:
!!/blame Github.dev

codygray:
The file also has to end with a newline character. That got omitted somehow. This is why the CI is failing this time (`Failed: watched_numbers.txt:4713:No newline`).

tripleee:
For context, I encouraged @micsthepick to submit this as a PR so that we could have a brief discussion and rope in @makyen as a reviewer. I think we would want to do this, but I wanted to check before merging.

tripleee:
Perhaps to remain faithful to the original, 1530522060 should remain the date stamp on the new entry?

micsthepick:
@tripleee I don't think that makes sense, unless we also attribute mayken there also, but that seems less genuine?

tripleee:
Well yes, if this is a migration then the original author and timestamp should be preserved; sorry for being vague.

micsthepick:
oops, somehow in all that rebasing I managed to copy paste the wrong whitespace, one moment

PR #13135, opened 2024-09-07
improve a number matcher in watched keywords

micsthepick:
don't match an ipv6 format, make sure we only start matching on the first digit.

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -66187,3 +66187,4 @@
 1725867376	Cow	tailormadejourney\.com
 1725868688	VLAZ	tailormade[\W_]*+journey(?!\.com(?<=tailormadejourney\.com))
 1725868974	VLAZ	airlinesofficedesk\.com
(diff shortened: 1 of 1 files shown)


micsthepick:
some more info - I expect this to cause at least one less FP, and it should not remove any TP. If it does then I'll review

micsthepick:
Metrics:
Before: TP: 1067 FP: 8 NAA: 0
After: TP: 1067 FP: 7 NAA: 0

makyen:
I'm confused. This says that it's "improve a number matcher", which implies that it's *changing* something, but the change to the watchlist is to only add an entry. Was something already removed somewhere else?

micsthepick:
Seems to have gotten lost in the rebasing, will check this @makyen

micsthepick:
actually, this may have been something that I noticed for two reasons, it looks like one of the migrations to watched_numbers makes this PR redundant

micsthepick:
relevant info: #13187

PR #13180, opened 2024-09-09
Watch NS seodns.one.

micsthepick:
RFC.

### DIFF
--- a/watched_nses.yml
+++ b/watched_nses.yml
@@ -5633,3 +5633,4 @@ items:
 - ns: hostarmada.net.
 - ns: edusum.com.
 - ns: bajika.co.
+- ns: seodns.one.


tripleee:
Like for any other PR, it would be nice if you included a motivation.

From Halflife's records, I see the following:

```
virtualfundsresurrection.info ['dns10.seodns.one.', 'dns9.seodns.one.'] (2024-09-09 22:44:40,452)
trustgeekshackexpert.com ['dns10.seodns.one.', 'dns9.seodns.one.'] (2024-06-08 00:43:32,137)
```

Both of these are spam (4/4 and 2/2 respectively) so with that, I'd call this a slam dunk. Feel free to remove the "draft" before I merge if you like.

PR #13245, opened 2024-09-12
Functionality to test longer posts by url

micsthepick:
Add a chat command to automatically test a long post in raw text at a given URL for rules by providing just a url, to avoid limitations of chat messages such as length.
*This allows testing github gists, or anything with a raw text file download URL.*

Considerations: unless the suffixes -q and -a are meant to not make any difference, we may want to check that they work in the intended manner.


makyen:
This does not appear to do what you state it is intended to do. It fetches the contents of a URL and tests that entire content. That's not testing a post. That's testing the contents of a full page, which, if it's an SE question page, is the question and answers and all of the surrounding HTML. It's unclear to me how that is beneficial. What is the use case?

What your text indicates that you're looking for is something similar to the `!!/scan` command, which tests an SE post. Having something that does something effectively like a `!!/scan`, but which outputs the `why` data instead of the report could be interesting. Perhaps something like a `!!/scan-why-only`.

I've also long thought that a `!!/bisect-post <SE or MS post URL>` command would be useful for those times when someone wants to find out what watch/blacklist entry was triggered by a post. I've recently been quite tempted to write that one, as that, combined with the `bisect` command's output of the detailed timing for long running regexes, would be a quite helpful tool for tracking down if there are specific entries which are taking a long time running.

micsthepick:
@makyen we don't want to test an existing SE post, but a raw page.

micsthepick:
see discussion at https://chat.stackexchange.com/transcript/message/66280595

Andrew5057:
Why does this PR modify `config.yml`? I take it that's accidental?

micsthepick:
@Andrew5057 I have no idea how that got in there, It should either be on a different branch or on a different clone in my local repositories.

Andrew5057:
@micsthepick At least you didn’t have your password in there :)

micsthepick:
Every time I force push to this repository, it closes the pull request?

Andrew5057:
I assume you did not mean to change 94 files? This might be a dumb question, but you're not doing `git add --all` are you?

micsthepick:
this time I used github codespaces as well as github.dev, the only change I did was to delete a file, and then continue in a codespace to be able to force push. I *think* somehow I've triggered a bug in github somewhere.

Andrew5057:
Yeah, I'd just use the CLI or maybe the desktop.

micsthepick:
Except I'm away from home,and the WiFi probably wouldn't take me downloading SD to my iMac fresh

micsthepick:
1. I didn't close this
2. I can't reopen?
but the branch is up to date now I think...

micsthepick:
asdf

PR #13346, opened 2024-09-16
fix catastrophic backtracking

micsthepick:
1544674518 in watched_keywords.txt

Possible alternatives:
1. just limit how much context we need:
https://chat.stackexchange.com/transcript/message/66302917
2. Use a named capture group too
this would reduce the time for the matching cases, by a tiny margin.

### DIFF
--- a/watched_keywords.txt
+++ b/watched_keywords.txt
@@ -6469,7 +6469,7 @@
 1544635783	Makyen	ca\.papersowl\.com
 1544635791	Makyen	wobbiewobbit\.com
 1544654286	Makyen	atten\.com
-1544674518	Makyen	\w*(?:约炮|約炮)\w*(?#slang for sex)
+1544674518	Makyen	(?=\w*(?:约炮|約炮))\w*(?:约炮|約炮)\w*(?#slang for sex)
 1544680688	Tetsuya Yamamoto	hifigadget\.com
 1544681529	Tetsuya Yamamoto	satishbrothersandsons\.com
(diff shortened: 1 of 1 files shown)


makyen:
How and what are you testing against and under what conditions that has resulted in you seeing catastrophic backtracking? I'm not seeing it. For example, [here is a regex101](https://regex101.com/r/yIsItq/1) with the original regex tested against both all four detections on MS and what I believe to be your test text (the Formatting Sandbox answers which you scanned: [1](https://metasmoke.erwaysoftware.com/post/497957), [2](https://metasmoke.erwaysoftware.com/post/497958)).

Are you using the `bisect` command and looking at times for regexes where this one is combined with another entry, rather than this entry on its own? If so, you may have encountered a bug in the `regex` package, which I've seen on some other occasions where an entry like this one combined with an entry that's even just straight text results in taking massive amounts of time, which isn't seen when the entries are tested alone.

I don't see how this change would actually solve a real catastrophic backtracking issue with this specific regex, but it could get around the above mentioned bug. In addition, this regex doesn't really have the characteristics that are normally necessary to trigger catastrophic backtracking (multiple quantifiers that can interact).

micsthepick:
@makyen that's my regex, not the original one!

makyen:
Sorry, I copied the wrong regex101 link. [Here is the one I intended to link to](https://regex101.com/r/YEypZe/1).

micsthepick:
closing as I can no longer replicate the issue

PR #13576, opened 2024-10-01
Optimize backoff handling and improve error handling in api_get_post

Ludusm11:
- Improved the lock usage and backoff handling by ensuring sleep only occurs for positive durations.
- Replaced try/except block with .get() for safer and more efficient dictionary access to 'last_edit_date'.
- Minor improvements to code readability and maintainability without altering core functionality.

### DIFF
--- a/apigetpost.py
+++ b/apigetpost.py
@@ -50,14 +50,16 @@ class PostData:
         setattr(self, key, value)
 
     def __getitem__(self, item):
-        getattr(self, item)
+        return getattr(self, item)
 
 
 def api_get_post(post_url):
     with GlobalVars.api_request_lock:
         # Respect backoff, if we were given one
         if GlobalVars.api_backoff_time > time.time():
-            time.sleep(GlobalVars.api_backoff_time - time.time() + 2)
+            # Use max(0, ...) to ensure no negative sleep times
+            time.sleep(max(0, GlobalVars.api_backoff_time - time.time() + 2))
+
         d = parsing.fetch_post_id_and_site_from_url(post_url)
         if d is None:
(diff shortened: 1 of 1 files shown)


tripleee:
These changes look good to me. Have you done any testing of this?

tripleee:
I'll be happy to merge this if you fix the spacing errors. Is this an attempt to get Hacktoberfest approval or where are you coming from?

apigetpost.py
```diff
@@ -86,17 +92,18 @@ def api_get_post(post_url):
...
     post_data.score = item['score']
     post_data.up_vote_count = item['up_vote_count']
     post_data.down_vote_count = item['down_vote_count']
     post_data.creation_date = item['creation_date']
-    try:
-        post_data.last_edit_date = item['last_edit_date']
-    except KeyError:
-        post_data.last_edit_date = post_data.creation_date  # Key not present = not edited
+    
+    post_data.last_edit_date = item.get('last_edit_date', post_data.creation_date)  # Use .get() for safer access
```
makyen, line 104: This line is a direct equivalent to the `try`/`except` block which it is replacing. There's significant discussion in the Python world as to which is "better", for various values of "better".  Personally, I prefer the `.get()`, but not enough to have previously pushed the change, but I wouldn't object to it being changed.

However, I would retain the functional comment that's in the old code, rather than have the comment that's been added here, which is just a comment on how to write code. In other words, the old comment is a reasonable code comment. The added comment is something that would be more appropriate in a commit comment instead of a code comment.

makyen (changes requested):

PR #13762, opened 2024-10-26
Prevent recompiling the BL/WL regexes which are unchanged by a blacklists reloaded event.

makyen:
This prevents the blacklist and watchlist regexes from being recompiled upon the blacklists being reloaded when the source text for the regex hasn't changed. Currently, all of the blacklist and watchlist regexes are recompiled each time the blacklists are reloaded (i.e., for each and every `!!/watch*` and `!!/blacklist-*` command). That costs quite a bit of computing time, currently looks like about 12 seconds each time, which is expended, even when a particular list didn't change. This cost is, potentially, significantly higher, because the compilation of each regex might be done multiple times, once per thread/CPU, due to the thread concurrently needing the compiled regex and starting the compilation.

There's a lot more that can be done in this area to reduce the amount of redundant compute cycles spent. This is just an easy first pass, which should provide substantial benefit in the current circumstance where we are adding a lot of entries to the numbers lists.

Testing starts: https://chat.stackexchange.com/transcript/message/66513237#66513237

### DIFF
--- a/findspam.py
+++ b/findspam.py
@@ -587,33 +587,40 @@ class FindSpam:
         global bad_keywords_nwb
 
         blacklists.load_blacklists()
-        cls.rule_bad_keywords.regex = get_bookended_keyword_regex_text_from_entries(GlobalVars.bad_keywords)
-        try:
-            del cls.rule_bad_keywords.compiled_regex
-        except AttributeError:
-            pass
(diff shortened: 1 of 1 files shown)


makyen:
I've gone ahead and merged this, as it has, potentially, significant impact on being able to keep up with an event like the spam wave we've recently seen.

PR #13765, opened 2024-10-27
Chunk the blacklists and watchlist to reduce regex recompiles upon reloading the blacklists and watchlist

makyen:
This PR does:
1. Splits each of the watchlist, bad keywords, blacklisted websites, and blacklisted usernames `Rule`s into two `Rule`s with the bulk of each list of entries in one of the two and up to 100 entries in the second. This results in dramatically reducing the computation time expended upon a watchlist or blacklist change for recompiling the regexes used. Overall, it should reduce the computation time spent on that task by an estimated 95%.
2. Implements `!!/scan-time` and `!!/scan-force-time`, which add an elapsed time to the output of the `!!/scan` command showing how long it took to execute the scan(s).
3. Adjusts some of the post titles for test cases in tests/test_findspam.py in order to make the test cases easier to identify in CI testing output.

Testing for this begins [here](https://chat.stackexchange.com/transcript/message/66517738#66517738) with the times for the final version tested starting [here](https://chat.stackexchange.com/transcript/message/66518195#66518195) and a baseline without the chunk/split of the BL/WL `Rule`s after that, starting [here](https://chat.stackexchange.com/transcript/message/66518215#66518215). The notable difference being the time after the `!!/watch opjd(?<!d)` command, which is 14.439 seconds in the baseline, but 4.641 seconds with these changes. The 4.641 seconds is only slightly higher than the 4.491 to 4.619 seconds required for the scan when there hasn't been a change that needs at least one of the regular expressions recompiled.


makyen:
The force-push was only to rebase to the current head, which will make the resulting commit tree look a bit cleaner, as the parent was previously a commit at the end of the prior PR I had submitted and merged, but without the merge commit.

PR #13813, opened 2024-11-20
WIP: Replace nonresolving domain names in URLs in test cases to speed up tests

tripleee:
Inject .example. where pertinent to avoid long DNS timeouts

This shaves off several minutes from tests on my laptop, but I don't know how well it transfers.

There are also a couple of tests which still exhibit DNS timeouts even after I moved them to example.com.

FWIW none of the domains here had blacklist entries which changed in any way because of these changes (either they were not blacklisted in the first place, or the change did not affect which blacklist entries would trigger).


tripleee:
No discernible difference in Github Actions timings; CircleCI seems slower than otherwise, if anything. )-:

tripleee:
https://chat.stackexchange.com/transcript/message/66658848#66658848

Issue #10228, opened 2024-03-04
Generated MS search lacks is_regex=1


tripleee:
### What problem has occurred? What issues has it caused?

When you create a PR from chat, SmokeDetector follows up with search results after a while. However, the generated search fails to specify that the search is for a regular expression. Thus e.g. https://github.com/Charcoal-SE/SmokeDetector/pull/10227 has a search which could never work.

(It didn't work anyway for other reasons, but without the "regex" check box, it couldn't have.)

### What would you like to happen/not happen?

The generated search should have the regex box ticked.

Andrew5057:
`[field]_is_regex` already gets added to website, keyword, and username watches/blacklists in [`gitmanager.py`](https://github.com/Charcoal-SE/SmokeDetector/blob/7bbb081c0e0c3ffedd2fee5c2ba443160335ac5f/gitmanager.py#L113). Number watches/blacklists are the only exception. How common are comments for number blacklists? Would a hard-coded exception for `(?# NO NorAm)` be enough?

Andrew5057:
Maybe we could just strip comments from number blacklists entirely. Something like `item_to_blacklist = regex.sub("\(\?#.*\)", "", item_to_blacklist)`?

makyen:
That regex is incorrect and will, potentially, strip things other than just comments. Please use the already existing regex to strip such comments. Actually searching for something that approximates what the phone number code does is substantially more involved than just stripping the comment and searching for the number. An *approximation* that's reasonably easy to code would be to use the same strategy as is used with the number search bookmarklets. Those don't attempt to handle potential obfuscations within the text. However, they do try to get at least some of those by also looking in the `why` data for the number, but the number showing up in the `why` data requires that it's already watched or blacklisted.

PR #13814, opened 2024-11-21
Use regex for auto-generated MS searches.

Andrew5057:
Add "body\_is\_regex=1" to the MS searches that SmokeDetector automatically creates for new PRs. Should resolve [#10228](https://github.com/Charcoal-SE/SmokeDetector/issues/10228).

Not tested yet - I was planning to, but I didn't want to remove the self-prompting filter in my SD fork, and my sock puppet doesn't have enough rep to use chat yet.


Andrew5057:
Actually, there's already some code to handle this in `[gitmanager.py](https://github.com/Charcoal-SE/SmokeDetector/blob/7bbb081c0e0c3ffedd2fee5c2ba443160335ac5f/gitmanager.py#L115)`. Not sure if this PR is still needed, but it definitely needs work.

PR #13831, opened 2024-11-24
Update watched_nses.yml - re-disable cloudxspeed.com

jeffschaller:
Resolves recent build failures based on cloudxspeed.com being NXDOMAIN again.

MX Toolbox result: https://mxtoolbox.com/SuperTool.aspx?action=mx%3acloudxspeed.com&run=toolpage

e.g.  https://github.com/Charcoal-SE/SmokeDetector/actions/runs/11996884075/job/33441859183 and https://github.com/Charcoal-SE/SmokeDetector/actions/runs/11996884075/job/33441859292

### DIFF
--- a/watched_nses.yml
+++ b/watched_nses.yml
@@ -5184,7 +5184,8 @@ items:
   - now1.dns.com.
   - now2.dns.com.
 - ns: thcservers.com.
-- comment: NXDOMAIN 2023-08-30; enabled again 2024-10-18
+- comment: NXDOMAIN 2023-08-30; enabled again 2024-10-18; disabled again 2024-11-24 (NXDOMAIN)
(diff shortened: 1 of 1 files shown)


makyen:
Interestingly, it's a different domain than what's been failing on CircleCI. I've pushed a commit for that domain, healthymum.org, too. Once that was pushed, CircleCI does report cloudxspeed.com as failing too. We may need to rework the tests a bit so that all of the errors are reported. The reason the failure of healthymum.org on CircleCi hides the failure of cloudxspeed.com is likely that they are in different files (blacklist vs watch). I'll need to look into it some more.

Using a `!!/dig` on SD shows a failure for both domains, so removing both is the way to go.

Issue #7425, opened 2022-12-27
Should we use type hints?


CoconutMacaroon:
### Is your feature request related to a problem? Please describe.
Python has support for [type hints](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5). They have advantages (IDEs and other tools can help users avoid making type-related errors by checking, for one). I'm not aware of any major disadvantages currently (SD supports Python 3.6 and they were added in 3.5, so compatibility isn't a huge concern I think). Overall, I personally tend to like them as having my IDE verify I don't mess up types is handy.
### Describe the solution you'd like
Use type hints on new code added to SD (and maybe some existing code, although I don't think that is really a priority).
### Describe alternatives you've considered
N/A
### Additional context
I asked if they should be used and [Makyen said it would be reasonable to open an issue on GH to discuss it](https://github.com/Charcoal-SE/SmokeDetector/commit/2db2bccab9b24d49a609e1c3518a87b6c417a4af#r94257578)

teward:
Personal opinion: yes.

However implementation code-base wise will be complex and may introduce some headaches for syntax, etc.  While generally I agree with us using type hinting, this is really a wider scale change than you think it is.

PR #13994, opened 2025-01-16
Support for blacklisting a user network-wide

### LINKED ISSUE #6845: Allow blacklisting user across the network
I'm cocomac on Stack Exchange. [Here is my Ask Ubuntu account](https://askubuntu.com/users/1438484/cocomac), which has GitHub linked on it. 

I'd like the ability to be able to blacklist a SE user, and have all of their SE per-site accounts get blacklisted as well. There aren't any cases that I know of where a user spams on one site, and is a great user on a different one. Most of the time, if they are spamming/trolling/being rude or abusive, they really should get blacklisted everywhere. I ran into this in recently, where I had to blacklist each of their per-site accounts manually after they were causing trouble on several sites.
Having one command to blacklist all of a user's SE accounts (and any that they create in the future, ideally) would be very helpful.
Blacklist the user on each site by hand - slow, tedious
### Additional context
There was a conversation in Charcoal HQ:

https://chat.stackexchange.com/transcript/message/60711269#60711269

CoconutMacaroon:
Resolves #6845.
This PR:

- [x] Adds the account ID to the `Post` class. It may be better to not do that in this PR, and instead use [the change\(s\) Makyen describes here](https://chat.stackexchange.com/transcript/11540?m=66878132) instead.
- [x] Updates `addblu` and `rmblu` text to note a network-wide blacklist.
- [ ] Change `addblu` and `rmblu` to note if a user is already network-wide blacklisted if changing a site blacklist
- [ ] Figure out the API filter - I wasn't seeing the `account_id` with the existing filter, although I'm not sure why. I've changed it for testing, but I (or someone else) need to look into that properly
- [x] Change the `isblu` text based on if they're (a) on the site blacklist & (b) if they're on the network-wide blacklist
- [ ] Decide if `isblu` should list site blacklists when checking a SE account
- [ ] Determine if `get_acount_id(self) -> int` should have `@property`
- [x] Add a note in the `check_if_spam` function - not sure if I did that correctly
- [ ] Keep/remove type hints - for the new code, I've added type hints as I find they improve maintainability/readability, but I could remove them. See also #7425.
- [ ] Properly write `is_user_net_blacklisted(user: tuple[str, str]) -> bool`, which determines if a given user is network-wide blacklisted. I wasn't sure how to approach this without a `Post` object to rely on (e.g., if called from `!!/isblu`), so I wrote some placeholder code, but the `else` section needs to be worked on.
- [ ] `.format(...)` vs [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings) - I opted to use f-strings in `!!/isblu` considering I was changing the strings anyways. In other places (e.g., `!!/addblu`), I left it as `.format(...)` as I was changing that function less. Still, if I shouldn't use them here, I can replace the f-strings I used with `.format(...)`

### DIFF
--- a/spamhandling.py
+++ b/spamhandling.py
@@ -42,11 +42,18 @@ def sum_weight(reasons: list):
             pass  # s += 0
     return s
 
+def is_post_owner_net_blacklisted(post: Post) -> bool:
+    """Check if the owner of a post is network-wide blacklisted"""
+    for acct_id, site in GlobalVars.blacklisted_users.keys():
+        if int(acct_id) == post.get_account_id() and site == "stackexchange.com":
+            return True
+    return False
 
 # noinspection PyMissingTypeHints
 def check_if_spam(post, dont_ignore_for=None):
     test, why = findspam.FindSpam.test_post(post)
-    if datahandling.is_blacklisted_user(parsing.get_user_from_url(post.user_url)):
+    net_blacklisted: bool = is_post_owner_net_blacklisted(post)
+    if datahandling.is_blacklisted_user(parsing.get_user_from_url(post.user_url)) or net_blacklisted:
         test.append("blacklisted user")
         blacklisted_user_data = datahandling.get_blacklisted_user_data(parsing.get_user_from_url(post.user_url))
         if len(blacklisted_user_data) > 1:
@@ -62,6 +69,8 @@ def check_if_spam(post, dont_ignore_for=None):
                 ms_url = datahandling.resolve_ms_link(rel_url) or to_metasmoke_link(rel_url)
                 why += "\nBlacklisted user - blacklisted for {} ({}) by {}".format(
                     blacklisted_post_url, ms_url, blacklisted_by)
+            if net_blacklisted:
+                why += "\nNetwork-wide blacklisted user"
             else:
                 why += "\n" + u"Blacklisted user - blacklisted by {}".format(blacklisted_by)
     if test:
--- a/apigetpost.py
+++ b/apigetpost.py
@@ -16,6 +16,7 @@ class PostData:
         self.owner_url = None
         self.owner_name = None
         self.owner_rep = None
+        self.owner_acct_id = None
         self.title = None
         self.body = None
         self.body_markdown = None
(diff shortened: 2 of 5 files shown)


spamhandling.py
```diff
@@ -62,6 +71,8 @@ def check_if_spam(post, dont_ignore_for=None):
                 ms_url = datahandling.resolve_ms_link(rel_url) or to_metasmoke_link(rel_url)
                 why += "\nBlacklisted user - blacklisted for {} ({}) by {}".format(
                     blacklisted_post_url, ms_url, blacklisted_by)
+            if net_blacklisted:
+                why += "\nNetwork-wide blacklisted user"
```
danthe1st, line 75: Is there any reason `blacklisted_by` is not included here?

danthe1st:
I guess the new commands should be [added to the wiki](https://github.com/Charcoal-SE/SmokeDetector/wiki/Commands/_edit) once this PR is merged.

CoconutMacaroon:
> I guess the new commands should be [added to the wiki](https://github.com/Charcoal-SE/SmokeDetector/wiki/Commands/_edit) once this PR is merged. - @danthe1st 

How does this look?

`!!/addnetblu <profile_URL>` or `!!/addnetblu <user_ID> <site_name>` — Adds a user to the network-wide user-blacklist (this means that any post by this user, on any site, will be reported when scanned). Accepts either a site profile or a network profile.

`!!/rmnetblu <profile_URL>` or `!!/rmnetblu <user_ID> <site_name>` — Removes a user from the network-wide user-blacklist. Accepts either a site profile or a network profile.

danthe1st:
This looks fine for me but I am not at all involved with maintaining anything in this repository at all.

talflon:
How does "fp" interact with your changes? Would it unblacklist the local user but not the network user? (I imagine that's the best option?)

CoconutMacaroon:
> How does "fp" interact with your changes? Would it unblacklist the local user but not the network user? (I imagine that's the best option?)

(From CHQ) - IIRC an FP would take a user off the per-sire blacklist but not the network-wide one. Agreed that doing that is likely the best choice.

danthe1st:
> (From CHQ) - IIRC an FP would take a user off the per-sire blacklist but not the network-wide one. Agreed that doing that is likely the best choice.
IMO this behavior is possibly a bit unintuitive and it might be a good idea to send a warning to the user giving the feedback (or maybe including it in the message sent to CHQ on the first feedback).

PR #14317, opened 2025-04-08
Add three Gujarati digits to the equivalents dict in number_homoglyphs

jeffschaller:
Spammers continue to look for ways to avoid detection of their fake support numbers; this PR expands the capabilities 
 of the number deobfuscation code by adding yet more visual matches to the translation table:
* add '૩' (Gujarati Digit Three) as a homoglyph for '3', with `0xae9`
* add '૫' (Gujarati Digit Five) as a homoglyph for '4', with `0xaeb`
* add '૭' (Gujarati Digit Seven) as a homoglyph for '9', with `0xaed`


jeffschaller:
These commits are a mess; will come back after I clean them up.

PR #14364, opened 2025-04-23
Update and improve util.py CLI script

talflon:
## 1. What does your pull request change or introduce to the project?

It updates the `util.py` CLI so that it works on Python 3.11+ and supports more commands than just `bisect`. The following new commands call into the same internal code as their namesake chat commands:

- `bisect-number`
- `deobfuscate-number`
- `normalize-number`
- `test-question`
- `test-answer`
- `test-user`
- `test-title`
- `test`
- `test-json`

The following new commands create temporary copies of the relevant blacklists, then manually add or remove items from these temporary copies, so that subsequent calls to `bisect-*` or `test-*` commands will test these modifications. They don't run the full logic of their namesake chat commands:

- `blacklist-keyword`
- `blacklist-website`
- `blacklist-username`
- `blacklist-number`
- `unblacklist`

## 2. What is the justification for the inclusion of your Pull Request (or, what problem does it solve?)

It lets someone test the blacklists outside of the chat, by cloning the SmokeDetector repository and installing the dependencies. If they want to test a lot of different things, it avoids spamming a chatroom. Furthermore, they could use this to test new regexes using the actual SmokeDetector code, before creating a commit on the public repository.

`util.py` used to be able to do this for `bisect`, but that was broken due to the removal of `inspect.getargspec()` in Python 3.11. I tried to fix that and extend the functionality in a similar way to what used to exist.

## 3. Write meaningful commit messages ...

I used brief commit messages because, by more or less duplicating the chat commands in `util.py`, it seemed the changes would be relatively self-explanatory.

## 4. Include comments in your code ...

A few comments were added to `util.py`, but it mostly relies on the `@utility` decorator's strings and the function and variable names.

The only changes outside of `util.py` were in `normalize_number()` and `deobfuscate_number()` in `chatcommands`: instead of assuming `msg` is not `None`, these methods now treat `msg=None` similar to how `bisect()` and `bisect_number()` already do, by falling back to the passed-in `pattern` argument.

## 5. Use meaningful variable names ...

I've tried to do that.

## 6. What testing have you done?

I have only tested it by running `util.py` myself and playing around with it. `util.py` wasn't working anyhow, so I'm assuming I can't have broken it too badly (or that many people have been using it).

I believe the only way it could introduce errors in other parts of SmokeDetector is if there was code relying on `deobfuscate_number()` or `normalize_number()` returning an error on `msg=None`. I don't see any error handling code there, so I'm hoping there's no requirement for it to error.

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -2825,8 +2825,9 @@ def normalize_number(msg, pattern):
     :param pattern:
     :return: A string
     """
-    minimally_validate_content_source(msg)
-    pattern = get_pattern_from_content_source(msg)
+    if msg is not None:
+        minimally_validate_content_source(msg)
+        pattern = get_pattern_from_content_source(msg)
     full_entry_list, processed_as_set, normalized = phone_numbers.process_numlist([pattern])
     return "Strict normalization: `" + phone_numbers.normalize(pattern) + \
            "`; Normalized numbers used by the number detections: `" + str(normalized) + "`"
@@ -2841,8 +2842,9 @@ def deobfuscate_number(msg, pattern):
     :param pattern:
     :return: A string
     """
-    minimally_validate_content_source(msg)
-    pattern = get_pattern_from_content_source(msg)
+    if msg is not None:
+        minimally_validate_content_source(msg)
+        pattern = get_pattern_from_content_source(msg)
     deobfuscated = phone_numbers.deobfuscate(pattern)
     normalized_deobfuscated = phone_numbers.normalize(deobfuscated)
     return "Deobfuscated: `" + deobfuscated + "`; deobfuscated and normalized: `" + normalized_deobfuscated + "`"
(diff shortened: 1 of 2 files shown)


talflon:
Another example of what this enables was today, when I could run
```
sed 's/^/bisect-number /g' two_hundred_numbers.txt | ./util.py 
```
and see quickly that none of them were watched or blacklisted.

talflon:
Potential improvements might be:

1. Instead of supporting multiple values per line and requiring shell quoting of the values, just take the entire line as a single argument.
   This would break the ability to use `xargs` with this, but would improve most any other workflow
2. Don't quit on a blank line.
   I recognize this existing behavior as somewhat of a "standard" from somewhere else, but I've never seen it be useful. I've only seen it annoying. Right now we implement it manually. We could stop doing that.

PR #14501, opened 2025-05-08
Update watched_nses.yml - disable hostnac.com (NXDOMAIN)

jeffschaller:
hostnac.com has been causing build errors for about the past 10 hours, starting with: https://github.com/Charcoal-SE/SmokeDetector/actions/runs/14901618248 and as recently as: https://github.com/Charcoal-SE/SmokeDetector/actions/runs/14912838315/job/41891197554; this comments out the entry.


jeffschaller:
Dan pointed out an extra dash; I'll fix it later, if no one else beats me to it.

Issue #14355, opened 2025-04-22
How should we go about blacklisting Unicode-obfuscated keywords?


talflon:
## Current method

Currently we're blacklisting a number of Unicode-obfuscated keywords which have shown up multiple times, like
- ℚ𝕒𝕥𝕒𝕣
- 𝑬𝒙𝒑𝒆𝒅𝒊𝒂
- ᏟႮՏͲϴᎷᎬᎡ
- 𝔸𝕚rlines

I like this, because I think it's very unlikely someone would naturally use Unicode formatting on these words. But by only blacklisting the ones we notice, we're missing the opportunity to catch new variants, which we can already expect to see later: if we've seen "𝓕𝓪𝓬𝓮𝓫𝓸𝓸𝓴", it'll be no great surprise to later see "𝐅𝐚𝐜𝐞𝐛𝐨𝐨𝐤". And I've seen that some of the spammers, instead of simply formatting the whole word, will format individual letters. That's a large number of possible variants, so it'd be easier to write regexes or code to catch them automatically than to wait and see which letter they'll pick next.

Also, it looks to me like we might not be catching things l̤̈ï̤k̤̈ë̤ ẗ̤ḧ̤ï̤s̤̈.

I have a couple different ideas for how to tackle this and was hoping for feedback on which might be the best way forward. I'm happy to try my hand at any of these.

## Idea A: expand the "Obfuscated word" reasons to include a lot more Unicode 

This was originally my Plan B, but looking at the SmokeDetector code, it looks like it would be fairly simple to do, so I'm calling it my Plan A now. `obfuscated_word()` could be rewritten to catch a lot of these Unicode tricks, instead of just 1337 speak and p.u.n.c.t.u.a.t.i.o.n.

A number of the above-mentioned blacklisted keywords could be removed, if desired, once `obfuscated_word()` is patched to catch them.

A small downside is that `obfuscation_keywords` needs manual PRs to update, while the blacklisters are familiar and comfortable with updating the blacklist.

### Idea A2: also check for blacklist and watchlist items

Many of the Unicode characters are unambiguous what letter they would stand for, even more so than in 1337 speak. It might be possible to do something like:

1. "deobfuscate" titles and bodies based on unambiguous Unicode characters, and stripping diacritics
2. run the blacklist and watchlist items over the deobfuscated versions
3. anywhere there's a match, check that region of the original to see if it was changed by the deobfuscation
4. if so, count it as an "obfuscated word"

[...]

PR #14768, opened 2025-06-05
Rewrite obfuscated_word to handle Unicode and deliberate spacing

talflon:
## 1. What does your pull request change or introduce to the project?

- Expands the `Obfuscated word in {}` reason to include:
	- Unicode-styled text, like 𝘼𝙞𝙧𝙬𝙖𝙮𝙨, 🇦​​🇮​​🇷​​🇱​​🇮​​🇳​​🇪​​🇸, or 𝗡𝘂𝗺𝗯𝗲𝗿, including sᴍᴀʟʟ ᴄᴀᴘs
	- s p a c i n g out of words
	- adding díãćrítìçš (or Zalgo text)
	- Unicode lookalikes, like ✞ҕɩ∫

	while keeping it based on a list of "keyphrases" to look for: it will not guess that any arbitrary word is obfuscated, just those predefined.
	
	This addresses #14355 (and this PR was based on my "Idea A" from that issue)
- Expands the list of keyphrases to look for obfuscation of, adding some more company names, scam keywords, and offensive words
- Splits the keyphrase list into two: one triggers `Obfuscated word` as an ordinary reason, and the other `Potentially bad keyword`, an experimental reason.
- Adds a couple seconds to startup time
- refactor `FindSpam.span_infos()` out of `FindSpam.match_infos()` so it can be used with non-regex spans
- `letter_homoglyphs` can be run as a script with a few diagnostic tools.

See Section 7 for an overview.

## 2. What is the justification for the inclusion of your Pull Request (or, what problem does it solve?)

Many of the things which this PR catches were being posted by spammers or trolls, most of them obvious to the human eye, but not caught by SmokeDetector unless we added them individually to the blacklist. Even when added to the blacklist, new variants of the same word might not be caught. It's just hard to manage this by pure regex.

## 3. Write meaningful commit messages

I tried to do this as I developed the code for this PR.

## 4. Include comments in your code to help guide reviewers through what you intend for the sub-portions of your code to be doing.

I tried to do this, too, though many of the comments are brief. See Section 7 for an overview.

## 5. Use meaningful variable names.

I tried to do this everywhere.

## 6. What testing have you done?  

I added some automated tests. In particular, they include a couple tests for situations which could cause running time blowup with the wrong combination of algorithm and input, and also tests inspired by FPs I saw during manual testing.

I also have run a SmokeDetector instance in the test room, unconnected to Metasmoke, and it seems to be working.

## 7. Overview of the new code

The text has code blocks, tags, and potential URLs stripped before being searched. It is not, however, "deobfuscated" first and then searched. This is because its list of possible homoglyphs is ambiguous. Despite that ambiguity, in my (limited) testing this algorithm does not appear to take much time or memory.

The code to search for keyphrases is in `ObfuscationFinder`, and some helper functions, in `letter_homoglyphs.py`. It compiles the keyphrases into a trie, then runs over the input text to find possible matches. Some of these are filtered out by regular expressions: most importantly, this is where matches are filtered out if the keyphrase is in plain text and not actually obfuscated at all.

### Keyphrases

The lists of keyphrases are `obfuscation_keyphrases` and `obfuscation_watch_keyphrases` in `findspam.py`. Each keyphrase has an optional "exclude" regex, which is bookended to not match in the middle of a word. This is usually used to exclude anything that includes a certain literal word.

The keyphrases usually won't match a partial word, but they will automatically match a plural S added to the end.

Some short keyphrases have the exclude regex `"^[[:ASCII:]]+$"`. This is to avoid false positives: `546E` *could* in theory be 1337speak for "sage", but it's much more likely to be a device model number, or a hexadecimal number.

Keyphrases right now only really handle the following characters:
- ASCII letters and digits
- period/full stop (".")
- hyphen, underscore, and spaces ("-", "\_", " ", `\t`, etc.) which are treated as word breaks

Other characters in a keyphrase are taken literally, so that it matches only if that exact character is in the input.

This means a keyphrase shouldn't include diacritics: it doesn't handle them automatically. Instead, I've written those without diacritics and used an exclude with `\w+` (because diacritics in theory could be combining characters) where the letter with diacritic would be:

```
("atencion", r"atenci\w+n"),
```

It's not counted as obfuscation if a word breaks, nor if it does not break, nor *how* it breaks, at a keyphrase word break: any space, hyphen, underscore, or even ChangeOfCase in the original keyphrase. This means that an obfuscation keyphrase "PayPal" will not trigger for "Pay Pal", "pay-pal", or "pay_pal", but will for "P ay P al".

### Homoglyphs

The lists of homoglyphs are in `letter_homoglyphs.py`. Some are added by name: `LATIN_LOOKALIKE_LETTERS`, `GREEK_LOOKALIKE_NAMES`, `CYRILLIC_LOOKALIKE_NAMES`, where each alphabet has a number of letter names:
- lowercase-only: name written in lowercase
- uppercase-only: name written in uppercase
- both/either: name written in a call to `add_case_insensitive_lookalikes()`

`LETTER_LOOKALIKE_DIGITS` is self-explanatory. A number of Unicode codepoints are also added manually with calls to `add_equivalents()`.

`POSSIBLE_SEPARATOR_CODEPOINTS` is a set of codepoints which could stand for characters, or could just be "decoration", like the vertical bar `|`: `d|e|c|o|r|a|t|e|d` vs `|etter`.

If not manually defined already, more homoglyphs are automatically added by the Unicode confusable list, added in the `unicode/` directory, and by compatibility decomposition (`NFKD`).

If not defined by any of the above, more games are played with the Unicode name to try to find a few more.

While this system is a bit over-engineered, the end result is over 5,000 homoglyph mappings, most of which appear reasonable to me, from short lists of letter names and codepoints in the Python code. It also means, though, that the list of mappings can change with different versions of the Unicode database (see [c5e5ab8](https://github.com/Charcoal-SE/SmokeDetector/pull/14768/commits/c5e5ab8645eef99b5c2fb10acc0528d4a127bfd1))

## 8. Possible future directions

- The couple seconds of setting up homoglyph lists could be changed:
   1. Could be turned into simple lists, or a text file to load, which would probably load almost instantly. This feels harder to read and maintain to me. Others might feel the opposite.
   2. Could be made lazy, that it spends the couple of seconds only when it first scans something.
   3. Lazy loading could be spread out by doing it per-character when that character actually appears in a scanned post. This might avoid any noticeable slowdown on scans or startup
- Skip fewer "potential URLs": the old `obfuscated_word()` caught some dot-obfuscation `li.ke th.is` which is now skipped as being possible URLs. I think a stricter URL test might be better for this rule.
- Handle keyphrases with diacritics better and automatically
- Look in code blocks
- Skip some MathJax matches: I tweaked the keyphrases to avoid MathJax FPs I saw during testing, but maybe we could use a heuristic to skip more MathJax. A problem with that is that it makes it a loophole. I believe at least one person has already posted TPs inside MathJax.
- Handle multiple-character lookalikes: one-Unicode to multiple letters, multiple characters to one letter (like `/\/\`)
- Use an unambiguous subset of these letter homoglyphs to "deobfuscate" text before blacklists are run on it?
   - or load a "simple" (not complicated regex) subset of the blacklists into keyphrase tries? I'd be curious how performance compares with regexes: initially I assumed regexes had to be much faster when compiled, but dictionary lookups *are* quite fast and I saw that regexes have some overhead, so now I'm not sure which would come out on top. Could code like in this PR speed up all scans?
   - Could the more liberal bookending rules here be useful elsewhere? (I saw a possible example of that the other day for a phone number.)
- Catch use of HTML tags to evade regexes
- Move the keyphrase lists to YML files
- More nuanced handling of possibly-innocent typos
	- Could some matches be automatically changed to experimental, based on the type of mistake made, to reduce chance of autoflagged FPs?


talflon:
I've since learned that my awkwardly-named `fullchars()` method probably should have been called `graphemes()`.

And there is in fact a Unicode standard for "grapheme clusters" ([UAX #29](https://unicode.org/reports/tr29)), and it's implemented in python-regex as `\X`. So in theory most of `fullchars()` could be replaced by

```python
regex.finditer(r'\X', text)
```

However:

1. I had wanted to rely on the first codepoint in the grapheme being the only one we usually need to look at. (In the current code, it's the only one that gets looked at.) But that's not guaranteed under UAX 29: it recognizes "prepending [...] marks" as part of a cluster.
2. Looping over `finditer` like this appears to run slower than the existing python loop.

There are some other differences between `fullchars()` and Unicode grapheme clusters, but I haven't found one yet that I think truly matters. This PR treats combining marks as *junk* to either be ignored or counted as obfuscation, not as part of the text, so whether the junk is combined with the main character or not is not as huge a deal as you might think: it really only affects whether it counts as obfuscation on the last letter of a word.

talflon:
Update on how I see this PR, some months later:
- currently less critical than when I first proposed it for catching spammers—they seem to be using this tactic less often now
- an evergreen troll tactic, that anyone could think up themselves, so I think something like this ought to be merged for that reason
- some improvements probably ought to be made, if this is merged
- I am happy to put a little effort into improvements, if someone could confirm if this will be merge-able at all

PR #14778, opened 2025-06-11
commented out nsluckup.com in watched_nses.yml (NXDOMAIN)

jeffschaller:


talflon (approved):
https://lookup.icann.org/ says "redemptionPeriod" but until it is either deleted or renewed, the old entry produces a lot of test failures

PR #14781, opened 2025-06-13
add test-ms command

### LINKED ISSUE #14532: Feature request: test-ms command

Just nice to have. After adding a blacklist or watchlist entry, I think it would be nice to be able to confirm if a past post would be caught or not, and for what, with similar output to the `test-*` commands. It also would help with old posts found with an MS search, to check what (or if) they'd be caught for today.

```
!!/test-ms https://metasmoke.erwaysoftware.com/post/...
```
would do the equivalent to:
- run `test-t site=<site> <post title>`
- run `test-q site=<site> <post body>` if it was a question
- run `test-a site=<site> <post body>` if it was an answer
- run `test-u site=<site> <poster username>`

Maybe the output could include a direct MS link.

I've thought about creating command line scripts I could run to do something similar, to create a JSON and pass to `test-json`. I think having it be public on Charcoal HQ, without a long JSON string, would be ideal, so other members could easily see the success or failure. Plus, many posts are too big to fit within chat message limits.

Currently when I wonder things like this, I try to guess which text from the post might be relevant, and copy and paste into a `test` or `bisect` command. A `test-ms` command would be quicker and more accurate.

talflon:
Adds a `test-ms` chat command to SmokeDetector, which runs the detection algorithms, like the other `test` commands would, on an old post loaded from Metasmoke. Closes #14532.

This allows testing if current spam criteria and blacklists would match an old post.

With URLs like

```
test-ms https://metasmoke.erwaysoftware.com/post/1
test-ms https://m.erwaysoftware.com/post/1
test-ms /post/1
```

it loads the post with MS id 1. (Yes, I am being very loose with the concept of a URL: it can be a URL or path.) With URLs like

```
test-ms https://metasmoke.erwaysoftware.com/posts/uid/stackoverflow/1
test-ms https://m.erwaysoftware.com/posts/uid/stackoverflow/1
test-ms /posts/uid/stackoverflow/1
```

it attempts to load posts from URL with native (SE/SO/etc.) id 1. If there's exactly one, it runs the test on that. If there's more than one, it returns a response like

```
Multiple posts found: /post/2, /post/3
```

which can be copied and pasted into new `test-ms` commands like:

```
test-ms /post/2
```

If the post fails to load from MS, it returns a response

```
Could not load URL '/some/invalid/url'
```

### DIFF
--- a/chatcommands.py
+++ b/chatcommands.py
@@ -1553,9 +1553,13 @@ def test(content, alias_used="test"):
         kind = "an answer" if fake_response['IsAnswer'] else "a question"
         fakepost = Post(api_response=fake_response)
     elif alias_used == "test-ms":
-        fakepost = Metasmoke.get_post_from_ms(ms_url=content)
-        if fakepost is None:
+        fakeposts = Metasmoke.get_posts_from_ms(ms_url=content)
+        if not fakeposts:
             return "Could not load URL {!r}".format(content)
+        elif len(fakeposts) > 1:
+            return "Multiple posts found: {}".format(", ".join("/post/{}".format(i) for post, i in fakeposts))
+        else:
+            fakepost = fakeposts[0][0]
     else:
         kind = "a post, title or username"
         fakepost = Post(api_response={'title': content, 'body': content,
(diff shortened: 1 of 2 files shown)


Cowm00:
This would add great value to CHQ, thanks for taking the time.

PR #14858, opened 2025-08-19
findspam.py: implement username_similar_anchor() and minor refactoring

### LINKED ISSUE #9989: New report reason: Username similar to link text
It would help catch more spam
SD should report posts where the text of a link is similar to a user's name, as that is almost always a sign of spam.
Post example (Username: Cool Spam Service, imagine the link went to a spam site instead of SO):
> Your problem can be solved easily [Cool Spam Service](//stackoverflow.com) with recursion...

We would want to catch these and give them a higher score. This is different from URL similar to username, as some companies' website doesn't always include their username ([MS example](https://metasmoke.erwaysoftware.com/post/460901))
N/A
### Additional context
N/A

micsthepick:
implementation based on username_similar_website,

Also because it uses a REGEX that shows up in two other places, refactor that to be defined once.

resolves #9989


micsthepick:
* and now local lint passed on WSL2

PR #14863, opened 2025-08-25
Prevent trying to fetch reply parent data for replies not to SD

makyen:
This expands on PR #14794, preventing SD from attempting to get data for the reply parent when the message being replied to is not from the user under which SD is running. In addition, a reply is sent to chat upon the HTTP errors we're logging and ignoring. This should result in both the user being informed that what they wanted to happen didn't occur, but also result in a further exception if we're blocked from posting, which would mean rebooting is probably a good idea.

### DIFF
--- a/chatcommunicate.py
+++ b/chatcommunicate.py
@@ -348,7 +348,14 @@ def on_msg(msg, client):
         if message.content.endswith("</div>"):
             message.content = message.content[:-6]
 
-    if message.parent:
+    if (msg.data.get('parent_id') and msg.data.get('show_parent')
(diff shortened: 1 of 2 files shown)


makyen:
Note the HTTP errors appear to be upon ChatExchange trying to fetch the transcript page containing the message, which it then scrapes for all of the messages on the page. It would probably be better to use the events endpoint and get the events. This would, however, provide different data as it would get the N messages prior to (and including) the desired message, rather than the messages that happen to be on the same transcript page, which would likely be messages both before and after the desired message.

makyen:
CI testing is failing on an entry, teammas.in, in watched_nses.yml, which is not failing in the GitHub Actions CI testing
.

PR #15174, opened 2026-02-06
Migrate from requirements.txt to uv

iBug:
In the past few years [uv](https://docs.astral.sh/uv/) has evolved rapidly as the most popular dependency management solution among the Python community. I've personally migrated projects both in school and in workplace from traditional `python3 -m venv` + pip solutions to uv and I must say it worths the effort. I think smokey would be another good chance to try out uv.

<s>Also [Ruff](https://docs.astral.sh/ruff/) came out as a very fast linter from the same team so I migrated from flake8 as well - it even comes with auto-fixing capabilities which flake8 doesn't. I can split this part (flake8 → Ruff) out to a separate PR if you want.</s>


iBug:
@makyen @teward Given your responses I've turned this PR into draft so it won't be merged (we stay with `pip` and `venv`).

Though I'd still like to see Ruff replacing flake8 - it's newer, faster, in active development, and I won't be turning on autofixes automatically (having the capability doesn't mean we'll be relying on it). I'll also tweak the rules to match current behaviors.

@teward I saw you gave a `-1` on the Pydantic proposal - mind elaborating why? I'm planning to use it on some JSON (in particular with HTTP API) and YAML things so that we don't write dict key accesses everywhere, and it's more robust.

iBug:
@teward This PR has been chopped down to migration of just the package manager and virtual environment. Mind taking a look now?

PR #15176, opened 2026-02-11
Migrate from flake8 to Ruff

iBug:


iBug:
@copilot+claude-opus-5:medium Determine whether this PR replicates the previous `flake8` behavior with Ruff.

iBug:
I'm giving a green-light to Ruff. I like this change and nothing has yet to break.

PR #15228, opened 2026-05-08
Require -force to blacklist/watch expressions with more types of potential mistakes

### LINKED ISSUE #14856: require using !!/watch-force to watch expressions with a lookahead at the start
When adding watched keywords, the usual advice from Makyen is to not start with a lookahead for performance reasons, https://chat.stackexchange.com/transcript/message/68148509 and the kerfuffle that follows is a prime example of how this can turn problematic fast.

I propose making !!/watch force require you to use -force if you REALLY want to start a regex with `(?!` or `(?=` or variants, but am open to suggestions.

might want to make it a bit more robust?

### Additional context
cc @makyen

talflon:
This PR expands the list of things checked before creating a blacklist or watch (similar to the existing check for an unescaped `.`), spitting out a message about why it might be a mistake and requiring `-force` to commit it. Fixes #14856 

The new checks are:

1. starting an expression with a lookaround

    > The pattern starts with a negative lookahead, which will match on most characters. For performance reasons, this should almost always be avoided.

   > The pattern starts with a lookbehind. For performance reasons, lookarounds should be as far into the regex as reasonably possible.

   - I chose to ignore this if the initial lookaround is a positive lookahead `(?=`, as I think this ought to be less of a performance problem than the other 3 kinds of lookarounds.

2. global flags being set, like `(?a)`

   > Please don't globally set regex flags in watch/blacklist entries. Use the `(?-i:...)` style to limit your change in flags to only the entry you are adding.

3. starting an expression with overbroad repetitions of nonword chars like `[\W_]*+` or `.{0,20}`

   > Starting a pattern with a repetition over all non-word characters, like `[\W_]*+`, can be computationally expensive, due to the Unicode `\b` in the bookending, and is often unnecessary.

   - I did this by attempting to compile and run a small test on the first token/group of the regex.

4. control verbs, like [`(*PRUNE)` or `(*FAIL)`](https://github.com/mrabarnett/mrab-regex#added-prune-skip-and-fail-hg-issue-153)

   > Control verb found in pattern. These might have global effects.

The existing and new checks have been refactored out of `chatcommands.do_blacklist()` into a new function `check_blacklist_mistakes()`, which now has a bunch of unit tests. To reduce naming confusion, the existing function `check_blacklist()` has been renamed `check_blacklist_duplicate()`, as that's all it checks for. (Naming confusion is not *entirely* eliminated, however, as `check_blacklist_mistakes()` still checks for duplicate *number* blacklists.)

My unit tests all pass, ~~but for some reason I've been unable to run a SmokeDetector instance to test this. The login is being rejected, even though I've double-checked the username and password (and they're unchanged since the last time when they had worked, and I can log in via web browser).~~

I did [some manual testing today](https://chat.stackexchange.com/transcript/message/68962012#68962012), thanks to @CoconutMacaroon's help, and the code appeared to be working.


chatcommands.py
```diff
@@ -278,98 +278,14 @@ def check_blacklist(string_to_test, is_username, is_watchlist, is_phone):
...
-    :param raw_pattern:
-    :param blacklist_type:
-    :param msg:
-    :param force:
-    :return: A string
+    Tests the string for some common blacklisting mistakes.
+    For some mistakes, raises `CmdException` to refuse to commit the blacklist or watch.
+    Returns a list of lesser issues, so that they can be reported to the user, or the commit allowed with -force.
     """
+    is_watchlist = bool("watch" in blacklist_type)
```
CoconutMacaroon: Is there a need for `bool(...)` here, or would this be fine?
```diff
-     is_watchlist = bool("watch" in blacklist_type)
+     is_watchlist = "watch" in blacklist_type
```

talflon: That would do the same thing, so I agree it would be better.

PR #15241, opened 2026-05-29
refactor: replace deprecated datetime.utcnow() with datetime.now(timezone.utc)

MD-Mushfiqur123:
Replace deprecated `datetime.utcnow()` with `datetime.now(timezone.utc)` in ws.py.

`datetime.utcnow()` is deprecated in Python 3.12 and removed in 3.14. This change uses the recommended timezone-aware alternative.

### DIFF
--- a/ws.py
+++ b/ws.py
@@ -21,7 +21,7 @@ from threading import Thread
 import traceback
 from bodyfetcher import BodyFetcher
 import chatcommunicate
-from datetime import datetime
+from datetime import datetime, timezone
 from spamhandling import check_if_spam_json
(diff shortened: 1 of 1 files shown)


makyen:
While the deprecation and removal of `datetime.utcnow()` needs to be handled, this PR will not resolve the issue. There have already been a couple of PRs attempting to address this issues, one of which, PR#8552, was much more comprehensive. That more comprehensive PR had to be reverted, as it resulted in hard crashes as a result of operations on incompatible types (i.e., operations involving both timezone-aware and timezone-unaware types).

